PR #75, opened 2015-01-07
Remove CIB version in case no --no-version.

seabres:
In case of switch --no-version of crm_diff not available, skip the version information in cib diff.
Skipping the epoch and num_updates information in source and target has the same result as completely skipping the version section in generated diff.


krig:
Thanks!

PR #77, opened 2015-01-19
Collection of patches for crmsh

krig:


krig:
Alright, with that sorted I will merge this. Thanks both of you!

krig:
Hmm. I noticed now that the transition patch broke the history test case.

It seems no transitions are found after applying this patch:

```
-Transitions: 43 44 45 46 48 272 49 50
+Transitions: 
+ERROR: 23: no transitions found in the source
```

Is this a reason to revert the patch?

dmuhamedagic:
On Thu, Jan 22, 2015 at 01:52:27AM -0800, Kristoffer Grönlund wrote:
> Hmm. I noticed now that the transition patch broke the history test case.
> 
> It seems no transitions are found after applying this patch:
> 
> ```
> -Transitions: 43 44 45 46 48 272 49 50
> +Transitions: 
> +ERROR: 23: no transitions found in the source
> ```

Oops, that's not good.

> Is this a reason to revert the patch?

Hmm, perhaps, I don't know, will take a look.

dmuhamedagic:
On Thu, Jan 22, 2015 at 01:52:27AM -0800, Kristoffer Grönlund wrote:
> Hmm. I noticed now that the transition patch broke the history test case.
> 
> It seems no transitions are found after applying this patch:
> 
> ```
> -Transitions: 43 44 45 46 48 272 49 50
> +Transitions: 
> +ERROR: 23: no transitions found in the source
> ```
> 
> Is this a reason to revert the patch?

The message format in this patch got introduced with this change:

commit d13ce588e9fe4277d6570a54f0728223fe505627
Author: Andrew Beekhof <andrew@beekhof.net>
Date:   Mon Sep 10 11:35:03 2012 +1000

    Low: PE: Minor cleanup

...
-            crm_err("Transition %d:"
-                    " ERRORs found during PE processing."
-                    " PEngine Input stored in: %s", transition_id, filename);
+            crm_err("Calculated Transition %d: %s",
+                    transition_id, filename);
 
         } else if (was_processing_warning) {
-            crm_warn("Transition %d:"
-                     " WARNINGs found during PE processing."
-                     " PEngine Input stored in: %s", transition_id, filename);
+            crm_warn("Calculated Transition %d: %s",
+                     transition_id, filename);
 
         } else {
-            crm_notice("Transition %d: PEngine Input stored in: %s", transition
_id, filename);
+            crm_notice("Calculated Transition %d: %s",
+                       transition_id, filename);

Maybe something like this would do:

    "pengine.* process_pe_message: .*Transition ([0-9]+): .*([^ ]*/pe-[^-]+-(%%)[.]bz2)",

krig:
Updating with this regex fixes the history test case, not only that, it manages to discover an additional transition thanks to this change :) So this seems good now.

PR #80, opened 2015-02-11
Medium: cibconf: preserve schema if set

dmuhamedagic:


krig:
I'm not sure I understand the subtleties here, why is validate-with different? Shouldn't all attributes be preserved in the same way?

dmuhamedagic:
On Thu, Feb 12, 2015 at 01:19:01AM -0800, Kristoffer Gr�nlund wrote:
> I'm not sure I understand the subtleties here, why is validate-with different? Shouldn't all attributes be preserved in the same way?

The cib element is a mixture of house-keeping and user
attributes:

<cib crm_feature_set="3.0.9" validate-with="pacemaker-2.0"
epoch="144" num_updates="6575" admin_epoch="0"
cib-last-written="Sun Jan 25 17:17:03 2015" have-quorum="1"
dc-uuid="178325681">

validate-with is the only one which the user can modify (with the
schema command). The others are managed mostly by pacemaker. We
update internally epoch and admin_epoch in order to prevent
commit errors.

Actually, that list in constants needs a different name:
cib_user_attrs :)

And the subject line is wrong too.

krig:
Alright, I see. Yes, I think maybe something like `cib_modifiable_attributes` is more explanatory, and the function could be `cib_preserve_user_modifiable_attributes`, maybe..

krig:
Looks good, merging :)

dmuhamedagic:
On Thu, Feb 12, 2015 at 02:25:56AM -0800, Kristoffer Gr�nlund wrote:
> Merged #80.

OK. Hope that it'll work too ;-)

Hmm, we don't have schema command in the regression tests. But
it's not clear to me how to exercise a schema change without
breaking the configuration.

krig:
Well, something like this could be tested with a unit test actually. I can see about adding one.

PR #86, opened 2015-03-24
Project Update

rikkotec:
### Updates
***

- Updated Makefile to remove prefix from python flags when installation
being performed on Debian
- Added README.md which is preferred over README on Github.com. This is
done so that on Github.com users can have a prettier display, yet
developers attempting to read the source file will not have all the
markdown spam in the way, since README still exists.
- Updated README itself to correct a spelling error (dependendent
changed to dependent).
- Updated .gitignore file so that tool specific files (eclipse's
.project, .pydev, .settings) are ignored and not added to the
repository.
- Tested everything as working

Change-Id: I06e9340e5c028aea44cafb0febad0690e87947e0
Signed-off-by: Richard B Winters <rik@mmogp.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -124,10 +124,16 @@ all-local:
 		--build-base $(shell readlink -f $(builddir))/build \
 		--verbose)
 
+if PREFIX_IS_USR
+python_prefix = --root=$(DESTDIR)
+else
+python_prefix = --prefix=$(DESTDIR)$(prefix)
+endif
+
 install-exec-local:
 	-mkdir -p $(DESTDIR)$(pkgpythondir)
 	$(PYTHON) $(srcdir)/setup.py install \
-		--prefix=$(DESTDIR)$(prefix) \
+		$(python_prefix) \
 		--record $(DESTDIR)$(pkgpythondir)/install_files.txt \
 		--verbose
 	$(INSTALL) -d -m 770 $(DESTDIR)/$(CRM_CACHE_DIR)
--- a/configure.ac
+++ b/configure.ac
@@ -114,6 +114,8 @@ eval libdir="`eval echo ${libdir}`"
 eval infodir="`eval echo ${infodir}`"
 eval mandir="`eval echo ${mandir}`"
 
+AM_CONDITIONAL([PREFIX_IS_USR], [test "${prefix}" = "/usr"])
+
 dnl Home-grown variables
 eval docdir="`eval echo ${docdir}`"
 if test x"${docdir}" = x""; then


krig:
Great! Unfortunately I think your fix breaks the rpm build. But... I think we can check especially for Debian instead of trying to make the condition more general.

I'm not entirely convinced about having both a README and a README.md, I might just throw away the "regular" readme and keep the one with the suffix. Markdown is not that intrusive so it should be acceptable even when looking at it directly.

Thanks for the PR either way! In general it looks good. I'll tweak and merge it as soon as I have time.

rikkotec:
This should be more as you desire now :)

rikkotec:
@krig 

Am going to test once more, and especially test the doc results -> but the latest update I think is what is necessary to add the proper value to a specific UNAME_IS_DEBIAN variable, via AM_CONDITIONAL -> which should allow cross-platform compiling without error.

rikkotec:
@krig 

Tested again, `README.md` was copied to `/usr/share/doc/crmsh`, and it works:

```shell
mmod@mmasx:/usr/share/doc/crmsh$ sudo crm
crm(live)# Ctrl-C, leaving
bye
```

`man crm` and `man crmsh` do not work, though said commands don't work in the previous build either - so guessing you just don't offer a manual for crm?

Either way, hope that helps and is what you wanted for a resolution :)

Makefile.am
```diff
@@ -22,7 +22,7 @@ MAINTAINERCLEANFILES    = Makefile.in aclocal.m4 configure
 
 sbin_SCRIPTS		= crm
 
-doc_DATA = AUTHORS COPYING README ChangeLog $(generated_docs)
+doc_DATA = AUTHORS COPYING README.md ChangeLog $(generated_docs)
```
rikkotec: This was changed to accommodate the file name change.

Makefile.am
```diff
@@ -124,8 +124,9 @@ all-local:
 		--build-base $(shell readlink -f $(builddir))/build \
 		--verbose)
 
-if PREFIX_IS_USR
-python_prefix = --root=$(DESTDIR)
+# Fix for GNU/Linux
+if UNAME_IS_DEBIAN
```
rikkotec: UPDATED `if`/`fi` to check for Debian specific flag `UNAME_IS_DEBIAN` - which should not interfere with other distributions as the previous implementation using `PREFIX_IS_USR` did.

configure.ac
```diff
@@ -45,7 +45,7 @@ AC_ARG_WITH(
     [ CRM_DAEMON_USER="hacluster" ],
   )
 
-AM_INIT_AUTOMAKE([no-define])
+AM_INIT_AUTOMAKE([foreign])
```
rikkotec: This modification allows `automake` and `autoconf` to ignore the fact that there is no `README` file with no extensions. The `README` file can be implemented via distribution package specific means.

Typically, I would not suggest to change this if it was set to gnu - though having been set to no-define makes this perfectly acceptable IMHO.

configure.ac
```diff
@@ -114,7 +113,7 @@ eval libdir="`eval echo ${libdir}`"
 eval infodir="`eval echo ${infodir}`"
 eval mandir="`eval echo ${mandir}`"
 
-AM_CONDITIONAL([PREFIX_IS_USR], [test "${prefix}" = "/usr"])
+AM_CONDITIONAL([UNAME_IS_DEBIAN], [test x`uname -v | grep -oh Debian` = x"Debian"])
```
rikkotec: This is a check which reads the 

> version of the LSB specification against which the distribution is compliant.

It works on RHEL and Debian - and should work on any Linux distribution.  It enables the `if`/`fi` in `Makefile.am`

.gitignore
```diff
@@ -6,3 +6,9 @@ Makefile.in
 autom4te.cache
 patches/*
 
+# Tool specific files
+.README.md.html
+.*.*~
+.project
+.settings
+.pydevproject
```
rikkotec: Lines 9-10 add simple ignores for Eclipse and PyDev Eclipse Plugin based files, as well as GEdit files.

PR #87, opened 2015-03-25
Fix Build for Debian & Add README.md

rikkotec:
#### Fix build on Debian
***

- Updated AM_CONDITIONAL defining `PREFIX_IS_USR`, to define
`UNAME_IS_DEBIAN` instead, by checking the version of the LSB
specification against which the host distribution is compliant, in order
to check whether we are building for Debian or not.
- Updated `if`/`fi` conditional check in `Makefile.am`, so that it does
a check on - `UNAME_IS_DEBIAN` instead of `PREFIX_IS_USR`.
- Added simple ignores in .gitignore for Eclipse and PyDev Eclipse
Plugin based files, as well as GEdit files.

#### Add README.md
***

- Added `README.md`, populated with markdown-formatted text.
- Removed `README`.
- Updated the `doc_DATA` variable so that it includes `README.md` rather
than `README`.
- Updated `configure.ac`, changing `AM_INIT_AUTOMAKE([no-define])` to
`AM_INIT_AUTOMAKE([foreign])`, in order to suppress error messages from
autotools with regards to a missing README file.


.gitignore
```diff
@@ -6,3 +6,9 @@ Makefile.in
 autom4te.cache
 patches/*
 
+# Tool specific files
+.README.md.html
+.*.*~
+.project
+.settings
+.pydevproject
```
rikkotec: These updates add ignores for Eclipse and PyDev specific files.

configure.ac
```diff
@@ -45,7 +45,7 @@ AC_ARG_WITH(
     [ CRM_DAEMON_USER="hacluster" ],
   )
 
-AM_INIT_AUTOMAKE([no-define])
+AM_INIT_AUTOMAKE([foreign])
```
rikkotec: `foreign` allows the the error surrounding a lack of a plain `README` file with no file extension to be suppressed.

configure.ac
```diff
@@ -114,7 +113,7 @@ eval libdir="`eval echo ${libdir}`"
 eval infodir="`eval echo ${infodir}`"
 eval mandir="`eval echo ${mandir}`"
 
-AM_CONDITIONAL([PREFIX_IS_USR], [test "${prefix}" = "/usr"])
+AM_CONDITIONAL([UNAME_IS_DEBIAN], [test x`uname -v | grep -oh Debian` = x"Debian"])
```
rikkotec: `uname -v | grep -oh Debian` should work with any Linux distribution.

krig:
Forgot to close this. Thanks for contributing :)

PR #91, opened 2015-04-26
high: report: Rewrite transition detection to fix missing transitions

krig:
@dmuhamedagic : I rewrote the transition detection so that it would use either of the transition start messages, since I noticed that some transitions only printed the old message (I think this is why we've been missing transitions).

One thing that's a concern is that a lot of transitions will be detected twice, and right now we keep the second, not the first Transition object created. I think we'd want to keep the first, most likely?

### DIFF
--- a/modules/constants.py
+++ b/modules/constants.py
@@ -338,17 +338,4 @@ extra_cluster_properties = ("dc-version",
                             "cluster-name")
 pcmk_version = ""  # set later
 
-# r.group(1) transition number (a different thing from file number)
-# r.group(2) contains full path
-# r.group(3) file number
-transition_patt = [
-    # transition start
-    "pengine.* process_pe_message: .*Transition ([0-9]+): .*([^ ]*/pe-[^-]+-(%%)[.]bz2)",
-    # r.group(1) transition number (a different thing from file number)
-    # r.group(2) contains full path
-    # r.group(3) transition status
-    # transition stop
-    "crmd.* run_graph: .*Transition ([0-9]+).*Source=(.*/pe-[^-]+-(%%)[.]bz2).: (Stopped|Complete|Terminated)",
-]
-
 # vim:ts=4:sw=4:et:
--- a/modules/report.py
+++ b/modules/report.py
@@ -16,7 +16,6 @@
 #
 
 import os
-import sys
 import time
 import datetime
 import re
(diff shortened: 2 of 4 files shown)


dmuhamedagic:
On Sun, Apr 26, 2015 at 04:09:26PM -0700, Kristoffer Grönlund wrote:
> @dmuhamedagic : I rewrote the transition detection so that it
> would use either of the transition start messages, since I
> noticed that some transitions only printed the old message (I
> think this is why we've been missing transitions).

Excellent.

> One thing that's a concern is that a lot of transitions will be
> detected twice,

One and the same twice? Or is it due to the transition number
wrap?

> and right now we keep the second, not the first
> Transition object created. I think we'd want to keep the first,
> most likely?

Hard to say. Depends on the answer above :)

krig:
Since it detects two different log lines as possibly starting a transition, if both forms appear in the log, two start messages will be detected for the same transition. The first transition will probably be earlier in the logs, so that's the one that should be kept.

dmuhamedagic:
In case there was a transition number wrap, it should pick the latter. At least that was why I chose to go with the latter line.

krig:
Hmm I see.. the possibility of wrapping complicates things. If you have two transitions with the same number in a report, you would probably want to see both though, not just one of them. So maybe the transitions should be merged somehow..

dmuhamedagic:
I doubt that they can be merged in any meaningful way. The reason
that only one survives in a report is that I've never figured out
a decent referencing scheme (1-old? 1-new? g1-1, g2-1, g3-1, ...
both look and sound horrible :)  At any rate, I don't think that
it's a big issue. People shouldn't be creating reports which span
such a great time periods. Or keeping the number of archived
peinputs too low.

dmuhamedagic:
Where do we stand with this one? Me lost track ;-)

krig:
I suspect that this isn't quite right yet, but I need to look at more reports.. though I think it is already an improvement over what is in master now. Maybe I should merge it and hopefully we can work out any issues afterwards. Sounds good?

dmuhamedagic:
On Mon, May 18, 2015 at 10:37:08AM -0700, Kristoffer Grönlund wrote:
> I suspect that this isn't quite right yet, but I need to look at more reports.. though I think it is already an improvement over what is in master now. Maybe I should merge it and hopefully we can work out any issues afterwards. Sounds good?

Yes, let's merge and then we'll test more.

PR #96, opened 2015-05-18
When generating CIB syntax, elide initial in some cases

krig:
This is something I worked on this weekend, but now that I have it working, I don't know if I really think it is a good idea. This would make eliding the "params" or "attributes" the default even when displaying the configuration, except when there are multiple params or attributes tags.

Dejan, how do you feel about this? I'm apprehensive, because although I think it improves the readability of examples like

```
primitive vip IPaddr2 ip=192.168.0.200
```

...there are other cases that aren't as clear cut, and it makes the generated syntax context-sensitive, not just the accepted input (where I don't think there are any downsides to allowing it).

So right now I am inclined to throw this away :P

### DIFF
--- a/modules/cibconfig.py
+++ b/modules/cibconfig.py
@@ -978,16 +978,18 @@ class CibObject(object):
             desc = self.node.get("description")
             if desc:
                 l.append(nvpair_format("description", desc))
+            first = True
             for c in self.node.iterchildren():
                 if is_comment(c):
                     comments.append(c.text)
                     continue
-                s = self._repr_cli_child(c, format)
+                s = self._repr_cli_child(c, format, first)
                 if s:
                     l.append(s)
+                    first = False
             return self._cli_format_and_comment(l, comments, break_lines=(format > 0))
 
-    def _attr_set_str(self, node):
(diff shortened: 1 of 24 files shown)


dmuhamedagic:
I'm always for being more concise. Now, keywords such as "params"
or "meta" served to signal what follows. BTW, meta attributes can
be listed _before_ params in an element and crmsh should also
preserve that order.

Otherwise, let me think about it some more.

P.S. I've always hated the "params" keyword :->

krig:
BTW, I have also been thinking about handling the "meta target-role=started|stopped" case in a special way, and replace it with for example the special keywords "::start" | "::stop". Not sure about this either :) (the :: is one suggestion, other suggestions are =>start, =>stop

So
```
primitive p0 Dummy state=1 =>start
```
instead of
```
primitive p0 Dummy params state=1 meta target-role=Started
```

Another idea I have is to rename "primitive" to "resource". `primitive` seems like a terrible XML-ism that has crept into the CIB syntax ;) But I don't know if this is something worth changing. The `resource` concept also includes groups and clones, but I think calling primitives resources makes sense.

PR #99, opened 2015-07-17
Medium: cibconfig: skip sanity check for properties other than cib-bo…

dmuhamedagic:
…otstrap-options


krig:
Seems reasonable to me :)

PR #104, opened 2015-10-06
high: cibconfig: Allow node/rsc id collision in _set_update (bsc#948547)

krig:
I'm a bit nervous about this change, so in order to make it stand out a bit I am making it a pull request.

The core problem is this: Pacemaker no longer enforces unique ids for nodes, and remote nodes may have associated resources with the same name as the node.

crmsh, on the other hand, has no notion of scope and has always assumed that ids are globally unique. The bugs resulting from this change may be subtle. All tests are passing, but who knows.

### DIFF
--- a/modules/cibconfig.py
+++ b/modules/cibconfig.py
@@ -570,9 +570,7 @@ class CibObjectSetCli(CibObjectSet):
         coming from edit). The original CIB is preserved and no
         changes are made.
         '''
-        edit_d = {}
-        id_set = oset()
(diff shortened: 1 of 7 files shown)


dmuhamedagic:
On Tue, Oct 06, 2015 at 02:01:51AM -0700, Kristoffer Grönlund wrote:
> I'm a bit nervous about this change, so in order to make it stand out a bit I am making it a pull request.

I'm quite a bit nervous too.

PR #106, opened 2015-10-07
doc: fix some command syntax grammar in the man page

aspiers:
Some `<>` brackets were missing.  There may be others which still need to be fixed.


aspiers:
BTW is this grammar using a formal syntax?  Maybe [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_Form) would be better?

krig:
Thanks! And no, it's completely ad-hoc. Feel free to fix that :) We'd need some way to validate it..

PR #108, opened 2015-10-12
improve fencing-topology docs

aspiers:
Some self-explanatory changes.

### DIFF
--- a/doc/crm.8.adoc
+++ b/doc/crm.8.adoc
@@ -2773,7 +2773,7 @@ Example:
 ...............
 fencing_topology poison-pill power
 fencing_topology \
-    node-a: poison-pill power
+    node-a: poison-pill power \
     node-b: ipmi serial
 fencing_topology attr:rack=1 apc01,apc02
 ...............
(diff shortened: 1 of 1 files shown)


aspiers:
It's still not clear to me whether stonith resources within a single comma-delimited set are run near-simultaneously.  If not, how can it guarantee that apc01 is cutting the power at the same time as apc02?  apc01 might complete before apc02 cuts power, in which case, the node will never get fenced.  And wouldn't this example fence all nodes in that rack, which is not good :)

dmuhamedagic:
On Mon, Oct 12, 2015 at 07:26:45AM -0700, Adam Spiers wrote:
> It's still not clear to me whether stonith resources within a single comma-delimited set are run near-simultaneously.  If not, how can it guarantee that apc01 is cutting the power at the same time as apc02?  apc01 might complete before apc02 cuts power, in which case, the node will never get fenced.  And wouldn't this example fence all nodes in that rack, which is not good :)

No, they're not run simultaneously. I think that there's still no
way to do that, but Yan should know better.

krig:
Thanks :)

PR #114, opened 2016-01-25
Add the "push" Method to the "configure load" Command

msmith626:
This adds a new method called "push" to the "configure load" command of crmsh. This new method differs from replace and update in that it removes configuration lines from the running configuration that are not present in the imported content (file/URL). Similar to the behavior of "configure edit" except that you can specify an import file instead of using an interactive editor.

I've tested this new feature with Pacemaker 1.1.12 + Corosync 2.3.4 and it works as expected, and the current two methods (update and replace) still appear to work as well. Tested the auto-complete function too.

I also fixed the grammar of a line in the help text for "configure load" and provided an example line using "push" for completeness.

--Marc


krig:
Thank you, looks good to me!

PR #115, opened 2016-01-28
Patch queue/remove fix for debian

rikkotec:
The fix for Debian which was added to resolve packaging issues is not necessary,  so this pull request offers the commits to remove the respective logic.

 - The `AM_CONDITIONAL` added to `configure.ac` is removed
 - The extra logic found in `Makefile.am` for setting `--prefix` is removed


krig:
Thank you! It's always nice to get patches that reduce complexity :)

Issue #67, opened 2014-11-12
crmsh moaning about invalid configuration although everything is okay


zacha81:
hello

running crmsh 2.1 or 2.2-rc1 on debian with pacemaker 1.1.12 I get all the time errors like this:

crm configure property maintenance-mode=true
ERROR: crmd:metadata: got no meta-data, does this RA exist?
ERROR: cib-bootstrap-options: attribute symmetric-cluster does not exist
ERROR: cib-bootstrap-options: attribute stonith-action does not exist
ERROR: cib-bootstrap-options: attribute maintenance-mode does not exist

Although the config changes are valid and are accepted by the cluster... possibly a problem parsing output from cibadmin?

enabling debug does not give me any hints on what crmsh is missing...

PR #119, opened 2016-02-20
medium: config: add multiarch support for libdir

dhoffend:
Debian and Ubuntu doesn't use the directory /usr/lib64. They switched to
a different directory structure for better multiarch support.

This also fixes a problem where crmsh can't find crmd on recent
Debian/Ubuntu 64bit installations (same as in #67) because the
autodetection doesn't find the correct crm_daemon_dir path

References:
* https://wiki.debian.org/Multiarch/
* https://wiki.ubuntu.com/MultiarchSpec

### DIFF
--- a/modules/config.py
+++ b/modules/config.py
@@ -19,7 +19,9 @@ _PERUSER = os.getenv("CRM_CONFIG_FILE") or os.path.join(userdir.CONFIG_HOME, 'cr
 _PATHLIST = {
     'datadir': ('/usr/share', '/usr/local/share', '/opt'),
     'cachedir': ('/var/cache', '/opt/cache'),
-    'libdir': ('/usr/lib64', '/usr/libexec', '/usr/lib',
+    'libdir': ('/usr/lib64', '/usr/lib/x86_64-linux-gnu', '/usr/lib/i386-linux-gnu',
+               '/usr/libexec', '/usr/lib', '/usr/local/lib64',
+               '/usr/local/lib/x86_64-linux-gnu', '/usr/local/lib/i386-linux-gnu',
                '/usr/local/lib64', '/usr/local/libexec', '/usr/local/lib'),
     'varlib': ('/var/lib', '/opt/var/lib'),
     'wwwdir': ('/srv/www', '/var/www')


krig:
Ah, excellent! I knew that this detection wasn't right on Debian, but hadn't gotten around to figuring out what needs to be there. Thanks for taking care of it!

dhoffend:
@ChristophBerg you're right. Imo we should remove the i386 arch but keep the x86_64 arch to be in line with the lib64 folder. Basically every other locations should be take care by a packaging change or other default configuration.

It should not be the job of the upstream software to take care of dozen arch and distribution specific folders.

krig:
I'd like crmsh to be able to figure out the most common locations. Anything more esoteric can be fixed in packaging by patching the included `crm.conf` configuration file to set the locations as necessary there. The end user can also configure it themselves.

One thing I probably should add is a clear error message when the auto detection fails.

dhoffend:
What do you think about the new solution. I know that sysconfig is only available in newer python version, but tbh ... older pacemaker versions weren't packaged and compiled with multiarch directory structure in mind. 

On Debian 7 crmd was located in /usr/lib/pacemaker/crmd while in Debian8 it has moved to /usr/lib/<arch>/pacemaker/crmd.

We could argue about fixing this in the pacemaker debian package or fix this in crmsh to lookup the support or MULTIARCH on the system and adjust the include/search path during runtime.

Not sure what should be the right solution. It sounds a bit like the egg-chicken problem.

ChristophBerg:
Re: Daniel Hoffend 2016-02-21 <ClusterLabs/crmsh/pull/119/c186848586@github.com>
> What do you think about the new solution. I know that sysconfig is only available in newer python version, but tbh ... older pacemaker versions weren't packaged and compiled with multiarch directory structure in mind. 
> 
> On Debian 7 crmd was located in /usr/lib/pacemaker/crmd while in Debian8 it has moved to /usr/lib/<arch>/pacemaker/crmd.
> 
> We could argue about fixing this in the pacemaker debian package or fix this in crmsh to lookup the support or MULTIARCH on the system and adjust the include/search path during runtime.

Debian will most likely be moving the /usr/lib/<arch>/pacemaker/ files
to /usr/lib/pacemaker/. The multiarch support we have is meant for
libraries (e.g. for cross compilation or running legacy 32 bit apps on
a 64 bit OS). It's not (yet?) tailored for multiarch binary (program)
support (think "/usr/x86_64-linux-gnu/bin/ls"), so it doesn't make
much sense to try to get pacemaker or crmsh multiarched now when it
doesn't work anyway on the OS level anyway.

Christoph
-- 
cb@df7cb.de | http://www.df7cb.de/

krig:
The best thing from my perspective would be if crmsh autodetects crmd correctly both before and after pacemaker is updated, though I'm wondering if maybe we should try the multiarch dirs last. But that's a minor tweak.

PR #121, opened 2016-03-13
high: history: Reimplement history to improve performance

krig:
This replaces the log parsing in the history explorer.

When first loading a report, we do a single pass over all logfiles
and all lines in each log. The result of this pass is a complete
list of all transitions and all interesting events in each log.

This data is then saved to `report_dir/__meta.json`. The next time
crmsh is used to look at this report, we load this cached metadata
directly.

The `refresh` command is updated to re-create the metadata for
non-live reports.

A new command, `transitions`, is added, similar to `peinputs v`.

This commit may break the live report refresh. At the very least,
right now it completely re-scans all log files after a live report
refresh, even though optimally we would only need to examine the
additional log data. Adding this optimization should be fairly
trivial if it turns out to be necessary.

Preliminary tests show massive performance improvements.

The limiting performance factor now is the number of regexp
evaluations per log line to find events. Combining all of the
regexes into a single huge regex actually decreased performance,
so something more clever may need to be done here.

Perhaps it would be possible to do a single pass over each log line,
outputting the timestamp, node, transition edge or event data.

The regression tests indicate that there are some differences in
the log output compared to the previous implementation. However, from
what I can tell these seem to be due to bugs in the previous
implementation.


krig:
@aspiers: Please help test this :) Two things that I think I may have broken: live report refresh, session handling.

krig:
My tests are all passing, I'm merging this now.

PR #123, opened 2016-03-17
make history example consistent with timeframe deprecation

aspiers:
It seems "timeframe" is deprecated in favour of "limit", so the examples should reflect that.

### DIFF
--- a/doc/crm.8.adoc
+++ b/doc/crm.8.adoc
@@ -4359,7 +4359,7 @@ command.
 
 Example:
 ...............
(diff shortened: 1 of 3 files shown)


krig:
hm, deprecated? they alias to the same command at least, but I have to admit that I use timeframe myself..

dmuhamedagic:
On Thu, Mar 17, 2016 at 04:12:29PM -0700, Adam Spiers wrote:
> It seems "timeframe" is deprecated in favour of "limit", so the examples should reflect that.

IIRC, it is the other way around :)

PR #142, opened 2016-05-13
fixed version number

steenzout:
it seems you already tagged a `2.2.1` release but `setup.py` wasn't updated.


krig:
Good catch, thanks!

PR #143, opened 2016-05-13
ability to install crmsh using pip

steenzout:
Even though crmsh is not in pypi, I should be to install crm directly from github using pip.

With the current `setup.py` file the packages are installed but the command-line interface isn't.

Using this PR I can install locally using
```
$ cd ~/Projects/github/steenzout/crmsh
$ virtualenv env
$ source env/bin/activate
$ pip install -r requirements.txt
$ python setup.py install
$ which crm
~/Projects/github/steenzout/crmsh/env/bin/crm
```

Demo using my own branch and using pip pointing directly from github:
```
$ cd ~/
$ virtualenv env2
$ source env2/bin/activate

# I'll make another PR that will make this step obsolete
$ pip install -r ~/Projects/github/steenzout/crmsh/requirements.txt

$ pip install -e git+git://github.com/steenzout/crmsh.git@steenzout-setup-scripts#egg=crmsh

$ $ which crm
~/Projects/github/steenzout/crmsh/env2/bin/crm
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,4 +9,5 @@ setup(name='crmsh',
       author='Kristoffer Gronlund',
       author_email='kgronlund@suse.com',
       url='http://crmsh.github.io/',
-      packages=['crmsh'])
+      packages=['crmsh'],
+      scripts=['crm'])


krig:
Nice! There's a few things missing if installing like this, but I guess the basic functionality is there.

Off the top of my head, the main things missing are

- builtin help - copying doc/crm.8.adoc to the right place should fix this
- man page - this is a pain since it needs asciidoc to compile. probably not worth caring about if the builtin help works
- crm history - hb_report isn't installed - this one is tricky since there's some autoconf magic
- crm script - the scripts aren't copied in, this should be easy to add to setup.py I think though

steenzout:
@krig I'll compare to what comes with rpm/deb packages and add what I can figure out how.

PR #145, opened 2016-05-13
tox

steenzout:
- use [tox](http://tox.readthedocs.io) to setup Python 2.6 and 2.7 virtualenvs and easily run tests against both
- make travis-ci use tox to run its tests
- ignore .tox directory when virtualenvs are created

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ patches/*
 contrib/build/
 contrib/dist/
 contrib/pygments_crmsh_lexers.egg-info/
+
+.tox/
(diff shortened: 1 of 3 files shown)


steenzout:
#146 will fix the build by eliminating flaky test.

krig:
I haven't used tox before and I'm not quite sure I get the benefit yet.. I guess it's to make it easy to test in multiple python versions locally, not only on Travis?

But sure, seems fine to me, and I guess I merged all the others already ;)

steenzout:
yes that's the point.
ability to test multiple Python versions (and/or different virtualenv setups) locally.
it also enables N people to use the exact same test virtualenv setup for their tests.

krig:
I see, thanks.

Issue #144, opened 2016-05-13
test_gv test is not robust


steenzout:
This test sometimes passes and sometimes fails.

Example:

```
AssertionError:
'digraph G {\n\nfontname="Helvetica";\nfontsize="11";\ncompound="true";\n"1a" [label="1a"];\na [label="a"];\n}' !=
'digraph G {\n\nfontsize="11";\nfontname="Helvetica";\ncompound="true";\n"1a" [label="1a"];\na [label="a"];\n}'
```

Just because the order of the keys is different the test should pass nonetheless.

krig:
Tests are passing, so closing this. Thanks again!

PR #146, opened 2016-05-13
#144 fix

steenzout:
fix issue #144 by comparing output line by line.


krig:
Huh, interesting that it's been solid for me. Thanks!

PR #147, opened 2016-05-14
Feature: add version control to resources and simple constraints

igor-tsiglyar:
This extension provides interface to configure versioned resources and simple location, colocation and order constraints.
Please, review and leave feedback.

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -1614,6 +1614,12 @@ class CibLocation(CibObject):
         s = clidisplay.keyword(self.obj_type)
         ident = clidisplay.ident(self.obj_id)
         s = "%s %s %s" % (s, ident, rsc)
+        rsc_version = self.node.get("rsc-version")
+        if rsc_version:
+            s = "%s:%s" % (s, rsc_version)
+        rsc_op = self.node.get("rsc-op")
+        if rsc_op:
+            s = "%s:%s" % (s, rsc_op)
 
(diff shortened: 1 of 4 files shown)


krig:
Where do versioned resources come from? I tried looking in the pacemaker source code but can't find any mention of them. How does it work?

igor-tsiglyar:
It comes from my not yet accepted [pull request](https://github.com/ClusterLabs/pacemaker/pull/1001).
Sorry, maybe I've opened this request a little bit early.
Versioned resource starts only if version parameter you stated when configured resource (e.g. 

> crm configure primitive A ocf:pacemaker:Dummy params version=1.1.1

) matches with resource agent version parameter obtained by 
> /usr/lib/ocf/resource.d/pacemaker/Dummy version

When it comes to constraints, now you can specify versions and comparators. For example,
> crm configure colocation A-with-B inf: A::1.1.1 B::1.1.2:gt

only applies to resource A with version == 1.1.1 and resource B with version >= 1.1.2

krig:
Aaah, I see! OK, I'll wait to merge this.

It looks like a potentially useful feature, although version comparisons quickly become extremely complex - at least that's my experience from rpm and various programming language package managers.

igor-tsiglyar:
Right you are! That's why after discussing this with the Pacemaker maintainers we decided to use rules interface, and there's no need to add a new one. So, I'll close the request. Thanx!

krig:
Thanks for the update :)

PR #148, opened 2016-06-08
parse: Use original _TARGET_RE

n3world:
Commits 65fa115874f4fac5020eb4ffda9d8ee6db69dda1 and f971e47e248aee1649c3c04a3a53937bff894149
changed _TARGET_RE from r'([^:]+):$' to r'([\w=-]+):$'. The new regular expression no longer accepts
many character the old one would. One character for example is '.', which could be in the target
name if fully qualified domain names are used for the node names.

If it is more desired to keep the regex restrictive it could be changed to r'([\w=\.-]+):$' to allow '.' but that would still restrict characters which were once allowed but might not be legal in the node name.


krig:
Yeah, lets go back to the old regex. Thank you.

PR #149, opened 2016-07-22
low: Fix a setting example of the alert to become the error

yuusuke:
can not describe this setting example when I obey relaxng rule of Pacemaker .


krig:
Thank you!

yuusuke:
Thank you for merging.

PR #163, opened 2016-10-30
Fix crm path

vvidic:
crm moved from /usr/sbin to /usr/bin causing an error:

```
# crm report -f 2pm report_1
/usr/share/crmsh/hb_report/hb_report: 527: /usr/share/crmsh/hb_report/hb_report: /usr/sbin/crm: not found
sid1: ERROR: could not figure out a list of nodes; is this a cluster node?
```

### DIFF
--- a/hb_report/utillib.sh
+++ b/hb_report/utillib.sh
@@ -524,7 +524,7 @@ crmconfig() {
 		CIB_file=$1/$CIB_F crm configure show >$1/$CIB_TXT_F 2>&1
 }
 get_crm_nodes() {
-	/usr/sbin/crm node server
+	crm node server
 }
 get_live_nodes() {
 	if [ `id -u` = 0 ] && which fping >/dev/null 2>&1; then


vvidic:
Can you confirm if the crm is to be installed as ``/usr/sbin/crm`` or ``/usr/sbin/crm``?

The Debian package for v2.3 needs to be update but it is not that clear from the commit log what is the intention.

krig:
Distributions should install it to `/usr/sbin/crm` still, yes. For openSUSE we run setup.py and then move `crm` from `/usr/bin` to `/usr/sbin`. I know this seems weird but my goal is to eventually allow `crm` to be installed in `/usr/bin` instead, so that it can be installed purely using `setup.py`. The ultimate goal with that is to make `crmsh` completely installable via `PyPI`/`pip`.

I intend to merge your change to hb_report as well.

vvidic:
Thanks for the explanation, we should stick with ``/usr/sbin/crm`` than until everything is ready for the move.

Issue #164, opened 2016-11-02
crm report does not work with dash


vvidic:
Default shell in Debian is dash rather than bash and this does not work as expected with the hb_report:

    # crm report -f 2am -M -E /var/log/syslog -d -s -Z
    /usr/share/crmsh/hb_report/hb_report: 480: local: /var/log/syslog.1: bad variable name

The problem seems to be with backtick calls returning multiple values:

    local logf_set=`arch_logs $logf $from_time $to_time`

The script should either require bash or the backtick calls need to be fixed.

dmuhamedagic:
Yes, initialization won't work in local. Will fix that.

vvidic:
Thanks, quoting seems to help here:

    local logf_set="`arch_logs $logf $from_time $to_time`"

However there are other backtick calls that have the same problem, so they should also be reviewed and quoted if needed.

dmuhamedagic:
I just split out all initializations from local statements. If you have time, please test. Let us know if there are more problems with dash. Thanks!

vvidic:
Thanks for the quick fix, no more errors :)

krig:
Thank you both!

PR #165, opened 2016-11-02
Medium: hb_report: don't use backticks in local

dmuhamedagic:
dash dislikes initializing variables along with the local
statmement (and it's anyway bad practice).


dmuhamedagic:
Should fix #164.

krig:
Thanks!

PR #170, opened 2016-12-06
Rewrite hb_report use python

wenshizhang:
Hb_report is part of crmsh, it would be better if hb_report implement in
python like crmsh.

For now, hb_report can do most works:
 * parse the parameters user input
 * get environment variables
   * cluster nodes list
   * from time and to time
   * and so on
 * find logs and configure files location
 * create static environment variables xmlfile and send it to cluster
 nodes
 * collect logs and send it back
 * remove tempfiles
Base on features, there are few classes, please read README.md
Here are several things need to do in future:
 * analyzed logs that are collected from cluster nodes
   * check which logs aren't exist
   * check logs difference between different nodes
   * check other stuff
 * check logs if empty
 * create tarball at specific location

Signed-off-by: Shiwen Zhang <szhang@suse.com>


krig:
Hi Shiwen!

It is probably better to start with forking the `crmsh` repository on github (using the fork button).
Then you can add the new `hb_report` implementation files to that repository, and then create a pull request.

With this pull request, there is no common history between the repositories, so the merge does not succeed.

wenshizhang:
Yeah, I know what you means. I will create now directory named python_hb_report and store my code, because there are some unfinished feature and I have not test the code carefully.

PR #174, opened 2016-12-13
Low: ui: Fix vim highlightning support.

phmarek:
The "pacemaker-crm.vim" would highlight most of the data as "Error",
which isn't helpful.


krig:
Commits that remove code are the best commits.

PR #176, opened 2017-01-20
give the op's hint when type tab after 'op'

liangxin1300:
When I add a new primitive resource and when I want add 'op' to this resource,
I input 'tab' after 'op', but the shell does not give me the hints, for example, "start", "stop" and "monitor"
So, I make the change

Regards,
Xin


krig:
Thank you!

PR #179, opened 2017-02-01
strip '..' after the 'ls' command && Add '.' hints for 'cd' command

liangxin1300:
Hi @krig ：
       I think it's a better way to remove the '..' output of 'ls' command.
       Because I see the most outputs of 'ls' command are command,  let 'ls' show just command to call, that make sense. 
       '..' is the output of 'cd' command, showing places to go, that make  sense
      ###################################
      And, for the Commits -- Add '.' hints for 'cd' command:
      For example, type 'crm->node->cd' and the 'tab', '..' append to the 'cd' command, but when I continue type 'tab', many '..' append to the command line, 
at last, the command line became "crm->node->cd .. .. .. ..", that make no sense.
      Add '.' option to the 'cd' command can avoid these.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,43 @@
+* Tue Jan 31 2017 Kristoffer Grönlund <kgronlund@suse.com> and many others
+- Release 3.0.0
+- high: bootstrap: Add bootstrap commands (fate#321114)
+- high: logparser: Update transition RE (#168)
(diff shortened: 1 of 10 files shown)


krig:
Hmm, I have to test this and see if I understand what is going on.

liangxin1300:
I also plan to implement "-" option for "cd" command, all of these can make crmsh more like a shell ;-)

PR #181, opened 2017-02-06
Add completers for crm->corosync->reload/edit/show/log

liangxin1300:


krig:
Does adding the empty `completers()` calls actually make a difference? That sounds like a bug that should be fixed if so. The idea is that you can add custom completion code, but without any custom code it shouldn't do any difference.

liangxin1300:
Adding empty completers() means these sub-commands does not need any arguments;
Just complete these sub-commands themselves when type 'Tab'

Issue #182, opened 2017-02-16
Function 'is_process' always return True


liangxin1300:
crmsh.utils.is_process

even the processname is not running or illegal

I think "ps -ef|grep -v grep|grep -q %s" is better

krig:
oh, good catch!

liangxin1300:
But it still not work
Please check that
When crmd is running
ps -ef -o comm= | grep -v grep | grep -qsE '\bcrmd\b' ==> return 1

PR #185, opened 2017-03-09
low: utils: Use /proc for process discovery

krig:
Parsing ps output turns out to be tricky, so
use the /proc filesystem for process discovery
instead.

Fixes: #182

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -952,14 +952,25 @@ def is_int(s):
 
 
 def is_process(s):
-    cmd = "ps -ef -o comm= | grep -v grep | grep -qsE '\b%s\b'" % s
-    if options.regression_tests:
-        print ".EXT", cmd
-    proc = subprocess.Popen(cmd,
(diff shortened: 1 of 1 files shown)


krig:
@liangxin1300 please review :)

crmsh/utils.py
```diff
@@ -952,14 +952,14 @@ def is_int(s):
...
-    proc.wait()
-    return proc.returncode == 0
+    commands = set()
+    pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+    for pid in pids:
+        try:
+            commands.add(os.path.basename(open(os.path.join('/proc', pid, 'cmdline'), 'rb').read().replace('\x00', ' ').split(' ')[0]))
+        except os.error:
+            pass
+    return s in commands
```
krig: Yes, I think I prefer reading `/proc` since it is less error-prone than parsing the command output, but maybe I can make the code clearer :)

PR #186, opened 2017-03-16
find_files function can not find link files in hb_report

liangxin1300:
find_files function cannot find link files;
So corosync-blackbox would not run when corosync have fdata

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -952,14 +952,25 @@ def is_int(s):
 
 
 def is_process(s):
-    cmd = "ps -ef -o comm= | grep -v grep | grep -qsE '\b%s\b'" % s
-    if options.regression_tests:
-        print ".EXT", cmd
-    proc = subprocess.Popen(cmd,
-                            shell=True,
-                            stdout=subprocess.PIPE)
-    proc.wait()
-    return proc.returncode == 0
+    """
+    Returns true if argument is the name of a running process.
+
+    s: process name
+    returns Boolean
+    """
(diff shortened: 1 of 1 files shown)


krig:
No, this is incorrect: We don't care about the symlink, we care about the actual fdata files:
`find_files /var/lib/corosync $from_time $to_time | grep -w fdata` will find the actual data files within the `$from_time - $to_time` range.

The data files are named `fdata-$timestamp`, so `grep -w fdata` will find them correctly.

krig:
An example:

```console
$ find_files /var/lib/corosync $from_time $to_time | grep -w fdata
/var/lib/corosync/fdata-2017-03-16T17:23:21-1725
/var/lib/corosync/fdata-2017-03-16T17:23:32-1725
```

liangxin1300:
But corosync-blackbox actually call qb-backbox, which just care about /var/lib/corosync/fdata
```
$ tail -3 /usr/bin/corosync-blackbox
corosync-cmapctl -s runtime.blackbox.dump_state str $(date +%s)
corosync-cmapctl -s runtime.blackbox.dump_flight_data str $(date +%s)
qb-blackbox "/var/lib/corosync/fdata"
```

By the way, 
`[ -f "$inpf" ]` will never true when find more than one fdata file

krig:
Well, the idea is that we'd only want to run if there's relevant data: some fdata file for the selected date range.

But you are right about the second point: that test should probably just be `[ -n "$inpf" ]`.

PR #188, opened 2017-03-20
In corosync_blackbox function,  $inpf may include more than one file

liangxin1300:
In corosync_blackbox function,  $inpf may include more than one file

and,
`echo $inpf|awk '{print $NF}'` can filter the newest one

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -678,19 +678,13 @@ Configure Corosync (unicast):
         if not confirm("%s already exists - overwrite?" % (corosync.conf())):
             return
 
-    bindnetaddr = prompt_for_string(
-        'Network address to bind to (e.g.: 192.168.1.0)',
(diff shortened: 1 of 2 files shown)


dmuhamedagic:
Hi Xin,

Better just improve one line earlier:

```
inpf=`find_files /var/lib/corosync $from_time $to_time |
	grep -w fdata | awk '{print $NF}'`
```

Also, perhaps `tail -1` instead of awk?

Finally, find_files uses find(1) to produce the list of files.
I'm not sure if they are time sorted.

liangxin1300:
How about
```
inpf_=`find_files /var/lib/corosync $from_time $to_time | grep -w fdata`
inpf=`ls -1t $inpf_ | head -1`
```

dmuhamedagic:
Yes, that looks good. Or you can also:

```
inpf=$(ls -1t $(
	find_files /var/lib/corosync $from_time $to_time | grep -w fdata) |
	head -1)
```

PR #189, opened 2017-04-14
remove bindnetaddr for unicast(bsc#1030437)

liu4480:
In a two-node cluster, I 've one node configured with open-vswtich:
5: br-fixed: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue
state UNKNOWN group default
inet 192.168.124.88/24 scope global br-fixed
inet 192.168.124.87/24 scope global secondary br-fixed

In this case, if you use unicast, and set both nodelist and bindnetaddr,
corosync will mess up the address in some condition, and Jan Friesse
also pointed that specifiing both bind addr and nodelist is kind of "not
very often used" configuration.

This patch removes bindnetaddr for unicast.

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -484,7 +484,7 @@ totem {
 
     interface {
         ringnumber:     0
-        bindnetaddr:    %(bindnetaddr)s
+        %(bindnetaddr)s
 %(mcast)s
         ttl:        1
     }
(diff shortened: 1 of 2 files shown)


crmsh/corosync.py
```diff
@@ -542,10 +542,15 @@ def create_configuration(clustername="hacluster",
     if transport is not None:
         transport_tmpl = "    transport: {}\n".format(transport)
 
+    if bindnetaddr is None:
+        bindnetaddr_tmpl = ""
+    else:
+        bindnetaddr_tmpl = "bindnetaddr: %s" % bindnetaddr
```
krig: Could you add the indentation (like for transport above) as well? Just to make the file look a bit nicer when created :)

PR #190, opened 2017-04-20
complete change about 'crm ra list [classes]'

liangxin1300:
lsb, service, stonith and systemd don't have any providers;
so, it shouldn't be completed when type 'tab' after these ra classes.


krig:
Ah, good catch! Thank you :)

PR #198, opened 2017-06-15
Use python-prompt_toolkit to improve interactive mode

liangxin1300:
Hi @krig :

There are four major changes in this PR:
1) Append "; use_prompt_toolkit = no" option into "[core]" section in "/etc/crm/crm.conf";
   So, by default, crmsh will not use prompt_tookit;
   To use it, uncomment both "[core]" section and "use_prompt_toolkit" option, 
   and let "use_prompt_toolkit = yes"

2) Prompt_toolkit provides autocompletion.
   Apps which plan to use it, must implement Completer Class(like what I'm doing in pt_completer.py);
   When user input "Tab", method get_completions() calls Context.complete() method, 
   to fetch every command level's completion results, complete when matches the rules

   So it can make the completion of crmsh more fluent

3) Use ~/.config/crm/.crmsh_history to record the input history

4) For colors, I just use prompt_toolkit.token.Token to colorize the prompt message

Any suggestions are welcome :)

Regards,
Xin


krig:
Hi Xin,

Thank you very much! I will test and give feedback :)

Issue #178, opened 2017-02-01
Still have some hints which unnecessarily export to users?


liangxin1300:
Hi:
    I see some hints after input the "tab" behand "crm configure", 
    such as "_keywords/_objects/_regtest/_test", which is not documented,
    I think these hints unnecessarily export to users
    My version is 2.3.2
    Regards,
    Xin

krig:
Ah, those should probably be hidden from tab completion, yeah. Good catch!

PR #201, opened 2017-06-15
remove _keywords/_objects/_regtest/_test options from tab completion

liangxin1300:
for issue [178](https://github.com/ClusterLabs/crmsh/issues/178)


krig:
Thank you!

PR #202, opened 2017-06-16
Take off the completion limit for resource agent params help messages

liangxin1300:
Hi:
  I know the codes I removed means simplify the output,
  the help message wouldn't fill full of the terminal;

  What's I mean is that: **Completion rule should be unified.**

  For example:
  1) User input crm->configure, and type "Tab" many times,
     then configure complete results output every time

  2) User input crm->configure primitive vip ocf:Heartbeat:IPaddr,
     and just type "Tab", do not input space or "2", 
     complete results "ocf:Heartbeat:IPaddr   ocf:Heartbeat:IPaddr2" will be completed all the way

  So, when User input crm->configure primitive vip ocf:Heartbeat:IPaddr params ip=
  and just type "Tab", if there have help message, just let it show;
  We have inputs history, people would not lost.

  I think that can make user less confued:)

  Regards
  Xin

### DIFF
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -125,15 +125,8 @@ class CompletionHelp(object):
     '''
     Print some help on whatever last word in the line.
     '''
-    timeout = 60  # don't print again and again
-    laststamp = 0
-    lasttopic = ''
-
     @classmethod
     def help(cls, topic, helptxt):
-        if cls.lasttopic == topic and \
-                time.time() - cls.laststamp < cls.timeout:
-            return
         if helptxt:
             import readline
             cmdline = readline.get_line_buffer()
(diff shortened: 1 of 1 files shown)


liangxin1300:
I temporarily close this PR, 
because when input crm->configure->
and input primitive....... , things still not work 
maybe in ui_context.py, readline_completer() function "line != self._rl_line" block completion happen

@krig , I will look at how to fix that, if you agree with me

krig:
The rule that I tried to implement (but there are definitely bugs) is

* If the prompt is at the end of a word, then display completions for that word
* If the prompt is at the start of a line or after a space, then display completions for the word after the last word

There is a special case for when the prompt is at a `=`, but I'm not sure that should really be needed...

PR #205, opened 2017-06-21
Give hints where the next input need an unique id

liangxin1300:
Hi:
When I input these commands:
```
configure primitive
configure group
configure ms
configure clone
```
I don't have an idea to input an id if I'm not familiar with cluster
maybe I will get Error message when I type "Enter" or
maybe I will delete the line and type "help" to find useful info between the long help message

I think give hints here will be nice and fluent :)

Regards,
Xin

### DIFF
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -70,7 +70,7 @@ def ra_classes_or_tmpl(args):
     return ui_ra.complete_class_provider_type(args)
 
 def mini_help(args):
-    if args[-1]:
+    if not options.interactive or args[-1]:
         return
     print "\nHints: need an unique id after \"%s\"" % args[0]
     import readline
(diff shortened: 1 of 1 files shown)


liangxin1300:
BTW, if this PR make sense, other place use "compl.null" should be replaced by "mini_help"

krig:
I like the idea, the only issue is that when doing bash completion, the hint prints as part of the command. So then the user has to erase the hint in order to enter the ID.

liangxin1300:
Thank you for remind me that
I will check the option.interactive :)

liangxin1300:
hi @krig, what's happen about the travis-ci? 
I noticed that some recent comments you pushed not sucess in ci with the same reason

krig:
Hi, yeah, there is a problem with the Travis setup. For some reason, the build system started picking up the wrong python version inside the ubuntu environment.. I need to change the Travis configuration to adapt to whatever changed there.

PR #206, opened 2017-06-27
Improved cd completion:

liangxin1300:
Hi:
 These are changes about cd completion:
  * on the root level, complete sublevel names after "cd "
  * on the sublevel, complete other sublevel path after "cd ../"
  * prevent '..' completion happend many times when type "tab"

Regards,
xin


krig:
Looks good, thank you!

PR #208, opened 2017-06-29
Improve ls command outputs:

liangxin1300:
* sorted outputs
  * print results column by column, like other complete results


krig:
Great, thank you!

PR #209, opened 2017-06-29
Remove "up/back/end" option at root level

liangxin1300:


krig:
Can you explain this change, is it solving a bug?

Issue #116, opened 2016-01-31
configure/verify completion doesn't seem to work


pzwahlen:
Hi and congrats for 2.2.0!

I'm playing with it on a lab cluster running on CentOS 7.2

For some reasons, the completion for "verify" under "configure" doesn't work for me.

v-TAB-TAB correctly gives me "validate-all validate_all verify", but ve-TAB-TAB gives me nothing.

Very very minor annoyance,

krig:
Hm, odd. I'll look into it, thank you!

dmuhamedagic:
This also fixed?

krig:
It's fixed, but I'm not too happy about the fix.. Haven't figured out a good way to provide both abbreviation detection and bash completion without interference.

liangxin1300:
hi @krig 
  I'm using crmsh-3.0.0-371 , this question still in there
  using my patch [198](https://github.com/ClusterLabs/crmsh/pull/198) can fix it :)

PR #210, opened 2017-07-03
medium: ui_context: Make all the options can be completed

liangxin1300:
Currently, not all the options can be completed by 'Tab', for example:
    crm->configure->verify
    crm->configure->back
    crm->configure->master
    crm->cluster->init
  Maybe because the subcommand doesn't have completer yet;
  Maybe because the subcommand itself is an alias of other command;

  This commit will completed all of the options.


liangxin1300:
also for the issue #116

krig:
Good catch, thanks!

PR #211, opened 2017-07-03
medium: ui_ra: Improve resource agents completion

liangxin1300:
When need to show the agent supported by cluster,
grouped by different classes or providers, Tab by Tab,
instead of completing all the agents at one time.

First completion:  agent classes
Second completion: providers of ocf or types of other classes
Last completion:   types of ocf agents

### DIFF
--- a/crmsh/ui_ra.py
+++ b/crmsh/ui_ra.py
@@ -8,27 +8,60 @@ from . import utils
 from . import ra
 from . import constants
 from . import options
-
+import re
 
 def complete_class_provider_type(args):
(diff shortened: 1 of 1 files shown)


liangxin1300:
hi @krig:
    how about this commit? Is it useful?
    I have the impression that someone complain the ra complete result is too much,
    it's better grouped by different classes or providers

krig:
Ah yes, this seems useful, I just haven't tested it properly yet.. thank you for the reminder.

PR #214, opened 2017-07-05
medium:command:adjust the 'ls' print width for long option

liangxin1300:
like 'crm cluster geo-init-arbitrator'


krig:
Thank you!

PR #215, opened 2017-07-05
medium:utils:list_cluster_nodes:read nodes list from cib.xml when clu…

liangxin1300:
…ster service stopped

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1597,7 +1597,14 @@ def list_cluster_nodes():
     try:
         rc, outp = stdout2list(['crm_node', '-l'], stderr_on=False, shell=False)
         if rc != 0:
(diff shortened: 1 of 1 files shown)


krig:
It seems a bit strange to call back into crm again for this command..

liangxin1300:
So, shall we consider the situation that the service is stopped?

krig:
What  I mean is that basically all "crm node server" does is scan the CIB XML document for `//primitive[@id='%s']/instance_attributes/nvpair[@name='server']`.

It would be cleaner to make a utility function which takes a CIB XML document and returns the list of nodes, and call that function both from `crm node server` and in `list_cluster_nodes`.

Calling `crm` from crmsh then makes it a requirement to have the `crm` command in `PATH` for example, which makes it less useful as a python module.

Issue #175, opened 2016-12-21
[Question] About "cluster health" command.


HideoYamauchi:
Hi All,

After a cluster started in a single node, I executed "cluster health" command.

```
[root@rh73-01-crmsh ~]# crm_mon -1 -Af
Stack: corosync
Current DC: rh73-01-crmsh (version 1.1.15-11.el7_3.2-e174ec8) - partition with quorum
Last updated: Wed Dec 21 17:58:27 2016          Last change: Mon Dec 19 22:14:39 2016 by hacluster via crmd on rh73-01-crmsh

1 node and 0 resources configured

Online: [ rh73-01-crmsh ]

No active resources


Node Attributes:
* Node rh73-01-crmsh:

Migration Summary:
* Node rh73-01-crmsh:

[root@rh73-01-crmsh ~]# crm --version
2.1.9 (Build unknown)

[root@rh73-01-crmsh ~]# rpm -qa |grep pssh
pssh-2.3.1-4.1.x86_64
python-pssh-2.3.1-4.1.x86_64

[root@rh73-01-crmsh ~]# crm cluster health
INFO: Check the health of the cluster
INFO: Nodes: rh73-01-crmsh
OK: Collect cluster information
OK: Run HA health check
OK: Report cluster state
Processing collected information...
INFO: Cluster report:
Diff hostcache... no /tmp/.hb_report.workdir.8GSFbc/health-report/*/hostcache :/
Diff members.txt... OK
Diff cib.xml... OK
Diff crm_mon.txt... OK
Diff corosync.conf... OK
Diff logd.cf... no /tmp/.hb_report.workdir.8GSFbc/health-report/*/logd.cf :/
Diff sysinfo.txt... OK

Log patterns:

WARNING: rh73-01-crmsh: No log rotation configured for corosync.conf

```

After having executed a command, "health-report" was made nowhere.

```
[root@rh73-01-crmsh ~]# find / -name health-report -print

```

After having executed a command with d option, "health-report" was made.
Is the difference in this movement right?
```
[root@rh73-01-crmsh ~]# crm -d cluster health
DEBUG: pacemaker version: [err: ][out: CRM Version: 1.1.15-11.el7_3.2 (e174ec8)]
DEBUG: found pacemaker version: 1.1.15-11.el7_3.2
INFO: Check the health of the cluster
INFO: Nodes: rh73-01-crmsh
(snip)
[...]

PR #216, opened 2017-07-07
low:scripts:health: save health-report when run "crm cluster health"

liangxin1300:
in crm_clean.py, if the clean request came from scripts:health,
delete all except health-report

in report.py, print workdir after all things done

this came from crmsh github issue #175


krig:
Thanks!

PR #217, opened 2017-07-12
low: bootstrap: some patches about cluster remove

liangxin1300:
1) add "delete" alias for "remove" option

2) when given a node name which is not configured in cluster, program will crash
    so replace 'nodename' to 'seed_host' can resolve this

3) give a confirm message when remove node
    it will be nice hints


krig:
Looks good, thanks!

PR #218, opened 2017-07-12
medium: ui_cluster: disable/enable completion when bootstrap running/…

liangxin1300:
…leaving

When running crm->cluster->init(or other bootstrap function),
completion make no sense, it's better to disable it,
and enable it before these function return

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -101,6 +101,7 @@ class Cluster(command.UI):
         '''
         Initialize a cluster.
         '''
+        context.disable_complete()
         def looks_like_hostnames(lst):
             sectionlist = bootstrap.INIT_STAGES
             return all(not (l.startswith('-') or l in sectionlist) for l in lst)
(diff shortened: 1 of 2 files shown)


krig:
Hmm, why not enable completion for init etc.?

liangxin1300:
when user use init to config the cluster, 
for example,
input Network address to bind to, or input y/n to confirm something,
completion still work when type "TAB", the cluster sub-commands will all pop after these prompt string,
it's strange.

So, it's simple and clear to disable completion "temporarily" when enter these sub-commands(init, join, geo-init, geo-join, etc.); and enable completion when these commands done

krig:
Aha, I see! I need to check, but it sounds like completion was happening incorrectly for any yes/no prompts then, not just in bootstrap.

But actually the right fix should be to have relevant completion for those prompts, and not sub-commands.

liangxin1300:
I re-implement this in #252

PR #221, opened 2017-07-18
medium: bootstrap: add callback function to check valid port range

liangxin1300:
for ucast and mcast port, valid range should between 1024-65536


krig:
Good addition, thank you!

Issue #219, opened 2017-07-13
cluster rename command


krig:
There should be a command to rename the cluster. The cluster name is more likely to be used in geo cluster situations where differentiating between clusters having the same configuration is useful.

Renaming the cluster isn't as simple as just changing the `cluster-name` property in the CIB or changing the name in `corosync.conf`: we need to do both, and also update any rules that refer to the cluster name:

1. Update `/etc/corosync/corosync.conf` with the new name on all nodes
2. Change the `cluster-name` property in the CIB and update any references to the cluster name in the same transaction
3. Reload the corosync configuration on all nodes

krig:
Fixed by #222

liu4480:
when executing corosync-cfgtool -R, corosync will definitely read corosync.conf, and generate temporary cmap configuration, but there are some configurations can not be changed at runtime, corosync will call remove_ro_entries to remove these configurations from the newly temporary cmap.
here is the implementation of the function. cluster name will be removed. That is why corosync-cfgtool -R can not change to the new name, it is by design.

/*
 * Remove any keys from the new config file that in the new corosync.conf but that
 * cannot be changed at run time. A log message will be issued for each
 * entry that the user wants to change but they cannot.
 *
 * Add more here as needed.
 */
static void remove_ro_entries(icmap_map_t temp_map)
{
    delete_and_notify_if_changed(temp_map, "totem.secauth");
    delete_and_notify_if_changed(temp_map, "totem.crypto_hash");
    delete_and_notify_if_changed(temp_map, "totem.crypto_cipher");
    delete_and_notify_if_changed(temp_map, "totem.version");
    delete_and_notify_if_changed(temp_map, "totem.threads");
    delete_and_notify_if_changed(temp_map, "totem.ip_version");
    delete_and_notify_if_changed(temp_map, "totem.rrp_mode");
    delete_and_notify_if_changed(temp_map, "totem.netmtu");
    delete_and_notify_if_changed(temp_map, "totem.interface.ringnumber");
    delete_and_notify_if_changed(temp_map, "totem.interface.bindnetaddr");
    delete_and_notify_if_changed(temp_map, "totem.interface.mcastaddr");
    delete_and_notify_if_changed(temp_map, "totem.interface.broadcast");
    delete_and_notify_if_changed(temp_map, "totem.interface.mcastport");
    delete_and_notify_if_changed(temp_map, "totem.interface.ttl");
[...]

PR #222, opened 2017-07-18
medium: ui_cluster: Add cluster rename command

liangxin1300:
* Update /etc/corosync/corosync.conf with the new name on all nodes
  * Change the cluster-name property in the CIB
  * Reload the corosync configuration on all nodes(use corosync-cmapctl)
    
  other:
  * command corosync-cfgtool -R can not reload totem.cluster_name
  * I have no idea how to update any references to the cluster name
  * this is for the issue #219


krig:
Great, thank you!

PR #223, opened 2017-07-18
low: ui_cluster: when use -h/--help option, do not exit

liangxin1300:
just print help messages and return

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -159,7 +159,9 @@ Note:
     To use storage you have already configured, pass -s and -o to specify
     the block devices for SBD and OCFS2, and the automatic partitioning
     will be skipped.
-""")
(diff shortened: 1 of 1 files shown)


krig:
Can you explain why make this change?

liangxin1300:
For staying with the shell,
for example,
1) crm->report -h, then program will print the help info and stay with the prompt string
2) crm->ra -h, after showing the help info, program not exit, just stay and wait for next input

BTW, the code "return False", which at "do_init" and "do_join" function, will make the whole program exit other than stay with the shell, 
it's a little bit strict.

liangxin1300:
I just create another commit in this PR,
to handle the optparse error(e.g. crm->cluster->init -x), and stay at the shell

krig:
Ah, right.. yes. Makes sense. Thank you. :)

PR #230, opened 2017-07-20
low: completers: filter out ms resource when doing promote/demote

liangxin1300:
if the action came from promote or demote, just filter out master/slave resource to complete


krig:
Thanks!

PR #231, opened 2017-07-20
low: cibconfig: use refresh instead of reset after commit

liangxin1300:
if use reset, for example, after add a valid resource and commit it,
without leaving the configure level,
delete completer will not get any id to delete


liangxin1300:
There is another commit:
medium: ui_configure: do not remind user they can still commit when error

  when add a resource, meet a fatal error, such as required parameter xxx not defined,
  it's dangerous to remind user they also have an option "y" to continue,
  even at this time, error message have poped up.

  If we close this remind, maybe can lead user to do the right configuration.

  It's dangerous enough to have "force" option for commit:)

PR #233, opened 2017-07-23
medium: bootstrap: use strict regrex for bindnetaddr/mcastaddr/adminIP

liangxin1300:
Commit sets for more robust bootstrap config:
1) medium: bootstrap: use strict regrex for bindnetaddr/mcastaddr/adminIP
2) low: bootstrap: color the input error message
3) low: bootstrap: valid bindnetaddr must one of the local networks
4) medium: bootstrap: csync2 update after all things done at local

    So, call csync2_update after all things done will solve this problem

5) medium: bootstrap: add valid_adminIP to valid adminIP
    The administration virtaul IP must:
    - have a valid ipv4 formation
    - belongs one of local networks

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -676,6 +676,15 @@ def init_corosync_auth():
     invoke("corosync-keygen -l")
 
 
+def valid_network(addr):
+    all_ = utils.get_all_networks()
+    if all_ and addr in all_:
+        return True
+    else:
+        print term.render(clidisplay.error("    Must one of {}".format(all_)))
+        return False
+
+
 def valid_port(port):
     if int(port) >= 1024 and int(port) <= 65535:
         return True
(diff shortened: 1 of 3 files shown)


krig:
Hmm, for the validation, I wonder if this is breaking IPv6 support - not sure if it was possible to configure IPv6 before. If not, maybe it should be. I need to look at that...

liangxin1300:
for now, the regrex '([0-9]+\.){3}[0-9]+' does not support IPv6 either

liangxin1300:
append another commit:
medium: enable add the second heartbeat for mcast
    features include:
    1) skip second configure when local networks < 2
    2) bindnetaddrs must be different and
       the gap between ports must larger than 1
    3) give different default value for second config

krig:
Well, we don't want to make it impossible to use IPv6!

liangxin1300:
Ok, I will figure it out and split this commit set.
Thanks!

PR #238, opened 2017-07-26
low: bootstrap: suppress the error message

liangxin1300:
When run "crm->cluster->join", if peer node doesn't  configured any RA,
  there will pop up a error message:
  "Call cib_query failed (-6): No such device or address"

Thanks,
xin


krig:
Good catch, that has been annoying me!

PR #241, opened 2017-07-30
low: ui_cluster: run command on a specific node

liangxin1300:
to make "cluster run" more flexible

Thanks,
xin


liangxin1300:
append another commit
 medium: bootstrap: csync2 update after all things done at local

   So, call csync2_update after all things done will solve this problem

PR #242, opened 2017-07-31
low: ui_cluster: run command on a specific node

liangxin1300:
to make "cluster run" more flexible

Thanks,
xin


krig:
That's a great addition, thank you!

liangxin1300:
#248 just use this commit
Thanks for accept

PR #245, opened 2017-08-02
medium: bootstrap: run "csync2_update" for all files after new joinin…

liangxin1300:
…g node call csync2_remote

 So, call csync2_update for all files when join node call csync2_remote
 can reslove this problem


krig:
Interesting, thank you!

Could you file this as a SUSE bugzilla? This sounds like a fix that we would want to track there as well.

liangxin1300:
OK, 
Thanks!

liangxin1300:
https://bugzilla.suse.com/show_bug.cgi?id=1052045

PR #246, opened 2017-08-02
low: ui_cluster: Add two new lines before add a new node

liangxin1300:
it's easy to read the status log which log block belongs what node:)

Thanks,
xin


krig:
Sure :)

PR #248, opened 2017-08-03
medium: bootstrap: configure with IPv6(mcast)

liangxin1300:
Changes include:
  1) use "-I" option stand for IPv6 in ui_cluster
  2) totem.ip_version should set as "ipv6"
  3) choose one interface which has IPv6 address as default values
  4) each node must have an uniqe nodeid when using IPv6,
     the nodeid came from IPv6 address in default interface
  5) for IPv6 network computation, learn from
     https://github.com/tehmaze/ipcalc


liangxin1300:
close it temporarily because maybe have conflicts with #249

PR #250, opened 2017-08-04
high: bootstrap: expected votes wouldn't update in unicast mode

liangxin1300:
e.g. configure two nodes cluster use unicast,
  when second node join in the cluster,
  because command "cibadmin -Ql --xpath '/cib/status/node_state'" just return one node,
  the expected votes not changed to "2"

  I think there should be:
  1) sync config file(nodelist) after add node
  2) reload corosync after init_cluster_local() to apply the nodelist
  3) sleep enough time while inquire using "cibadmin":)

  BTW: if not do step1 and step2, cibadmin -Ql always return one node

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1315,6 +1315,7 @@ def join_cluster(seed_host):
     is_unicast = "nodelist" in open(corosync.conf()).read()
     if is_unicast:
         corosync.add_node(utils.this_node())
+        csync2_update(corosync.conf())
 
     # if no SBD devices are configured,
     # check the existing cluster if the sbd service is enabled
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -1327,16 +1328,24 @@ def join_cluster(seed_host):
...
+            rc, nodelist_text = utils.get_stdout("cibadmin -Ql --xpath '/cib/status/node_state'")
+            if rc == 0:
+                try:
+                    nodelist_xml = etree.fromstring(nodelist_text)
+                    nodelist = [n.get('uname') for n in nodelist_xml.xpath('//node_state') if n.get('remote_node') != 'true']
+                    if len(nodelist) >= 2:
+                        break
+                except Exception:
+                    pass
+            sleep(1)
```
krig: There should probably be a timeout here eventually, especially if `rc != 0`.

liangxin1300:
I just add a 10 seconds timeout in the loop
Thanks!

krig:
Alright, lets merge :)

Issue #243, opened 2017-07-31
bootstrap or cluster lock?


liangxin1300:
Hi @krig:
    When having more than two nodes to configure cluster, and, 
    when some nodes run "crm cluster join" to join a configured node simultaneously, 
    the csync2 will failed.

It's better to join a configured node **one by one**.
    To implement this, 
    (e.g. node A has been configured, node B and node C want to join)

1) touch a hidden lock file on the node A when node B want to join, and excluding node C until node B's join process done
2) remove the hidden lock file on node A when node B finished join process
3) leave the interface "do_lock" and "do_unlock" in ui_cluster, it's convenient to touch and remove the lock file manually 

Any suggestion will be appreciated:)
Regards,
xin

krig:
Ah yeah, that's a really good idea. The only issue is in the case of a failed join, the lock file might be left on the node. In that case, it would block future attempts to join.
I would say at the very least, there should be a clear message printed when blocking, and an option to ignore:
"Cannot join cluster while another node is joining. Lock found on remote node (override with --ignore-lock)."

liangxin1300:
Thanks for suggestion!

liangxin1300:
Close it
For latest implement, see https://github.com/ClusterLabs/crmsh/pull/677 and https://github.com/ClusterLabs/crmsh/pull/694

PR #251, opened 2017-08-07
medium: bootstrap: cluster join lock

liangxin1300:
This commit is the implementation of issue #243
 
When having more than two nodes to configure cluster,
It's better to join a configured node one by one, or
the csync2 will failed when some nodes join simultaneously

the differences from #243 are:
1) when have waiting every 30s, ask user to confirm whether still want to wait,
2) do not add "do_lock" and "do_unlock" in ui_cluster
3) no "--ignore-lock" option, asking user to confirm is enough, it can force to join in when always blocking:)

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1576,6 +1576,7 @@ def bootstrap_init(cluster_name="hacluster", nic=None, ocfs2_device=None,
         if watchdog is not None:
             init_watchdog()
         init_ssh()
+        join_lock()
         init_csync2()
         init_corosync()
         if template == 'ocfs2':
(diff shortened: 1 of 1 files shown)


liangxin1300:
Append commit "medium: bootstrap: cluster init lock"

When first node use init to configure,
blocking other nodes joining until this node finish init process

liangxin1300:
Append commit "medium: bootstrap: joining node self lock"

The node which is joining the configured node,
should create lock file in local until this node finish join process

e.g. There are 3 nodes A, B, C
A is in init process,
B want to join A, blocking...
C want to join B, blocking until B finish it's join process

PR #252, opened 2017-08-08
medium: bootstrap: disable completion and history when running bootstrap

liangxin1300:
When running crm->cluster->init(or other bootstrap function),
  completion makes no sense, it's better to disable it

I think this is a better and clear way against #218 
1) When meet multi_input and ask in bootstrap, still have completion handler, but the handler will complete nothing
2) remove the history item after inputs in bootstrap, because these items cannot use for other crmsh's commands and levels

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -28,6 +28,7 @@ from . import corosync
 from . import tmpfiles
 from . import clidisplay
 from . import term
+import readline
 
 
 LOG_FILE = "/var/log/ha-cluster-bootstrap.log"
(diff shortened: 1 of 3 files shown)


liangxin1300:
The issue is that, when other commands instead of bootstrap calling utils.ask, completion still not necessary

should we pass the context into the utils.ask to disable it's completion?

crmsh/bootstrap.py
```diff
@@ -105,9 +106,13 @@ def prompt_for_string(msg, match=None, default='', valid_func=None):
     if _context.yes_to_all:
         return default
     while True:
+        disable_completion()
```
krig: Shouldn't this be `_context.disable_completion()`?

liangxin1300: no, disable_completion() is a local **wrap** function in line 134:)

krig: Ah, I missed it :)

PR #257, opened 2017-08-14
medium: bootstrap: use strict regrex and valid function for bindnetad…

liangxin1300:
…dr/mcastaddr
  1) a valid bindnetaddr must be one of the local networks
  2) a valid mcastaddr must in a valid range and be a valid IPv4/v6 format


krig:
Looks good, thanks!

PR #258, opened 2017-08-15
medium: bootstrap: valid adminIP

liangxin1300:
commit medium: utils: extend "IP/Network" codes for IPv4/IPv6 calculation
and add test codes for testing:
1) Check if the given ip is part of the network
2) get ipv6 network

==============================
based on above commit,
second commit is -- medium: bootstrap: valid adminIP
The administration virtaul IP must:
1) have a valid IPv4/IPv6 formation
2) belongs one of local networks

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -739,6 +739,31 @@ def valid_v6_network(addr):
         return False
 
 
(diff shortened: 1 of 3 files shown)


krig:
Great, thank you!

PR #261, opened 2017-08-16
medium: utils: list_cluster_nodes: read nodes list from cib.xml

liangxin1300:
so this can get full node list no matter  what the node's status;

based on this, in ui_cluster, do_start/do_stop/do_enable/do_disable can use self.do_run to control all nodes


krig:
Good idea!

PR #262, opened 2017-08-16
low: ui_cluster: to control service on all nodes or a specific node

liangxin1300:
and update doc file

BTW: how about "cluster->status"? 
It's better to have a cluster-wide show on defaults;
or it can have a specific node status,
not only local status:)

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -30,6 +30,13 @@ def _remove_completer(args):
     return scripts.param_completion_list('remove') + n
 
 
+def _all_nodes(args):
+    try:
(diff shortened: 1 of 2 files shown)


krig:
Well, for cluster-wide status, there is the top-level status? :)

krig:
Actually, I think it is not good to make the default depend on SSH access to other nodes.

I think `crm cluster start` should only start the local node, like now.

Maybe add `crm cluster start all` which starts on all nodes?

Actually, starting or stopping all cluster nodes at the same time is a strange operation. The result is unpredictable. I don't know if it is even necessary to have a command to do that.

PR #264, opened 2017-08-16
low: main: add hostname in promptstr

liangxin1300:
is this necessary?

I think it's more clear when in cluster level


krig:
Sure, lets try it :)

PR #267, opened 2017-08-21
medium: enable adding the second heartbeat line for mcast

liangxin1300:
Changes include:
  1. skip second heartbeat configure when number of local networks < 2
  2. bindnetaddrs must be different and the gap between ports must larger than 1
  3. give different default value for second configuration
  4. IPv4 & IPv6 support


krig:
Nice, thank you!

PR #269, opened 2017-08-21
low: bootstrap: give error hints when use sbd without watchdog

liangxin1300:
It's better stop config process here,
instead of trying to init local cluster


liangxin1300:
append a commit: low: bootstrap: catch OSError except instead of crash
for inputting not exists device

liangxin1300:
append a commit: low: bootstrap: reset dev value when input of dev not valid

Issue #268, opened 2017-08-21
adjust unicast configure file template?


liangxin1300:
[corosync.conf.udpu.txt](https://github.com/ClusterLabs/crmsh/files/1237553/corosync.conf.udpu.txt)
above attach files is for 2 heartbeat lines;

I think it's more clear to use specific IP address instead of hostname for ringX_addr items in nodelist

PR #271, opened 2017-08-24
medium: enable add the second heartbeat line for unicast

liangxin1300:
Changes inlcude:
  1. IPv4 & IPv6 support
  2. user should specify local IP for ringX_addr when init and join both
  3. user input must be one of local IP addresses
  4. peer ringX_addr and local ringX_addr must in the same network

BTW: as mentioned in #268, it's more clear to use specific IP address instead of hostname for ringX_addr items in nodelist, and can avoid the situation like:
   User has 192.168.1.1/24 and 192.168.2.1/24;
   He writes 192.168.1.1 in /etc/hosts, but the real IP he wants to reserve for HA cluster is 192.168.2.1

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -745,6 +745,17 @@ def valid_v6_network(addr, prev_value=None):
         return False
 
 
+def valid_ucastIP(addr, prev_value=None):
+    if prev_value and addr == prev_value[0]:
+        print term.render(clidisplay.error("    {} has been used".format(addr)))
+        return False
+    ip_local = utils.ip_in_local(_context.ipv6)
+    if not addr in ip_local:
+        print term.render(clidisplay.error("    Must one of {}".format(ip_local)))
+        return False
+    return True
+
+
 def valid_adminIP(addr, prev_value=None):
     """
     valid adminIP for IPv4/IPv6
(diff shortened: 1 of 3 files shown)


krig:
I would prefer if multiple rings was an option only and by default only one ring is configured.

Multiple rings is not a recommended configuration.

liangxin1300:
but two rings are more safe 
I give a confirm to ask whether to add a second ring, after finish the first.
These like that we do ask user whether use watchdog, or ask whether use sbd, 
its a way to make a more reliable cluster :}

liangxin1300:
so, if user want a quick setup, they input "n", then got a one ring, no sbd, no adminIP cluster;
if they want a more reliable cluster, they just input "y", they still have a choice.
Put this function into option is not a convenient  way

krig:
No, two rings is not more safe, it is actually worse than one ring on top of a bonded network interface. With two rings, if a network critical to cluster services goes down the cluster may not react. We don't recommend configuring two rings.

liangxin1300:
hi @krig, you have merged #267, which is also a commit for two rings

liangxin1300:
or, I will try to change both of them as option-mode(give a patch to #267),
this will convenient for non-interactive mode(or for saltstack?)

BTW: 
we can not assume that user would use bonded network interface,
the number of network interface in server is at least 4,
user can use two of them to create two rings,
and use another one for Virtual-IP,
and use number 4 interface for DRBD/iscsi etc.

krig:
I don't mind making it possible to configure two rings, but I think it should be optional, yes :)

PR #274, opened 2017-08-28
Detect firewall by checking installed packages

vvidic:
For example, firewalld and ufw are available in Debian.


krig:
Thank you!

PR #275, opened 2017-08-28
Fix SBD configuration when using SBD device (Fixes #235)

### LINKED ISSUE #235: crm cluster init: sbd systemd service options
Cluster bootstrap updates the SBD config file with ``SBD_DEVICE`` and ``SBD_WATCHDOG`` options but [systemd service](https://github.com/ClusterLabs/sbd/blob/master/src/sbd.service.in) only takes ``SBD_OPTS`` from there:

    EnvironmentFile=-@sysconfdir@/sysconfig/sbd
    ExecStart=@sbindir@/sbd $SBD_OPTS -p @localstatedir@/run/sbd.pid watch

Is there a better service file available, or should the bootstrap script be updated to generate ``SBD_OPTS`` instead?

vvidic:
Disables pacemaker mode and enables watchdog device.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1237,7 +1237,12 @@ Configure SBD:
         error("Failed to initialize SBD device %s" % (_context.sbd_device))
     status_done()
 
-    utils.sysconfig_set(SYSCONFIG_SBD, SBD_DEVICE=_context.sbd_device, SBD_WATCHDOG="yes")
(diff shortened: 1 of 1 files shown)


krig:
Why disable pacemaker mode here?

vvidic:
Pacemaker mode was set explicitly in disk-less case, so I though it was not needed when SBD device is used. Updated the patch now to leave pacemaker check enabled.

krig:
Thank you!

PR #276, opened 2017-08-29
medium: ui_configure: in modgroup, add free id and remove id in group

liangxin1300:
when modgroup, 
  after "add", id should not from any groups, 
  after "remove", id must from the specific group,
  after "after/before", id must from the specific group


liangxin1300:
append a commit: 
medium: ui_configure: in modgroup, complete none after remove

krig:
Looks good!

Issue #272, opened 2017-08-24
configure modgroup switch bug


zafai:
Hello,

in the `configure modgroup` function is a bug.
```
crm(live)# configure
crm(live)configure# modgroup NFS add res_nfs_exportfs before res_nfs_mailto
crm(live)configure# show NFS
group NFS res_nfs_fs1 res_nfs_ip1 res_nfs_srv res_nfs_exportfs res_nfs_mailto \
	meta target-role=Started
crm(live)configure# commit
crm(live)configure# show NFS
group NFS res_nfs_fs1 res_nfs_ip1 res_nfs_srv res_nfs_mailto res_nfs_exportfs \
	meta target-role=Started
```

As you can see before the commit is looks correctly after the commit its switching the order and the wrong order is written into cib configuration.

liangxin1300:
hi @zafai, I have repeated that. 
Thank you!

liangxin1300:
@zafai ,
I don't think this is from crmsh
it's from pacemaker.
crmsh use "crm_diff" to get the diff patch, and use "cibadmin -p -P" to apply the diff patch,
I think the problem is in diff patch, which got a wrong "position" in group.

I will report a bug to pacemaker.
For now, I suggest you use "crm->configure->edit" to change the order of RAs in group :}

krig:
There is a known bug with the older diff format, which cannot handle reordering. This should be fixed with newer versions of pacemaker, though. Which version of pacemaker are you running?

Inventor1409:
Pacemaker Version from zafai was the Version from openSUSE Leap 42.3

```
Name        : pacemaker
Version     : 1.1.16
Release     : 3.6
Architecture: x86_64
Install Date: Sat Jul 22 22:16:06 2017
Group       : Productivity/Clustering/HA
Size        : 1168889
License     : GPL-2.0 and GPL-2.0+ and LGPL-2.1+
Signature   : RSA/SHA256, Sat Jul  8 02:54:08 2017, Key ID b88b2fd43dbdc284
Source RPM  : pacemaker-1.1.16-3.6.src.rpm
Build Date  : Sat Jul  8 02:53:16 2017
Build Host  : cumulus3
Relocations : (not relocatable)
Packager    : http://bugs.opensuse.org
Vendor      : openSUSE
URL         : http://www.clusterlabs.org
Summary     : Scalable High-Availability cluster resource manager
Description :
Pacemaker is an advanced, scalable High-Availability cluster resource
[...]

PR #277, opened 2017-08-29
medium: ui_configure: in modgroup, comlete none when remove for one r…

liangxin1300:
…a group

remove last resource in modgroup will cause syntax error,
so we just complete nothing when one resource left in group


liangxin1300:
These commits about modgroup are inspired from #272 issue:)

PR #278, opened 2017-08-30
medium: ui_node: node utilization is about node attr, not for resources

liangxin1300:
and define a new completer function "_find_utilization" to complete utilization attrs

BTW: in ui_node, do_attribute and do_status_attr seems have the same problem


krig:
Ah, nice catch!

liangxin1300:
OK, so I will fix the same problems left in ui_node.py later:)

krig:
Great :)

PR #279, opened 2017-08-30
medium: ui_node: node attribute/status-attr is about node attr, not f…

liangxin1300:
…or resources

just like #278 , apply the rules into attribute and status-attr;
and, custom the function "_find_attr" to handle attribute/utilization/status-attr


liangxin1300:
hi krig, maybe we should release a minor version as the end point of PY2 :)
and, please merge this commit as an integral patch for #278 problem

krig:
Hi, yes, makes sense! Merging.

PR #284, opened 2017-09-29
This is a proposed fix for bug 283.  The fix was to change the except…

handruin:
… object type to IOError when a directory cannot be found due to a process that has terminated.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1002,7 +1002,7 @@ def is_process(s):
             procname = basename(cmdline.replace('\x00', ' ').split(' ')[0])
             if procname == s:
                 return True
-        except IOError:
+        except EnvironmentError:
             # a process may have died since we got the list of pids
             pass
     return False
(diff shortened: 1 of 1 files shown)


krig:
I'm wondering if the correct fix would be to catch EnvironmentError, since it is the base exception for both IOError and OSError.

handruin:
If that's more desirable that's fine with me.  I can give this a try and rerun my test to see how it works under Python 2.6.

krig:
That would be great, thank you :)

handruin:
I made your proposed change and ran my test under Python 2.6 and this works fine.  What is the right process for making this change under github?  Should this PR be rejected and I create a new one with the proposed change?

This is the test I ran for 1000 iterations with success:

`#!/bin/bash

for i in {1..1000}; do
        /usr/sbin/crm corosync set logging.timestamp on &> /dev/null
        if [ $? -ne 0 ]; then
                echo "Error found with crm command"
                break
        else
                echo "crm return is good.  Attempt: $i"
        fi
done`

handruin:
I added my change to the same branch and it looks like the CI test passed.  Please let me know if you would like me to try anything else or make any other change for this.

krig:
Great, that looks good. I'll merge it :)

PR #286, opened 2017-10-05
'crm cluster stop' only stops pacemaker

vvidic:
Stopping corosync will also stop pacemaker via systemd dependencies.

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -90,9 +90,9 @@ class Cluster(command.UI):
         '''
         Stops the cluster services on this node
         '''
-        rc, out, err = utils.get_stdout_stderr('service pacemaker stop')
+        rc, out, err = utils.get_stdout_stderr('service corosync stop')
         if rc != 0:
-            context.fatal_error("Failed to stop pacemaker service: %s" % (err))
+            context.fatal_error("Failed to stop corosync service: %s" % (err))
         err_buf.info("Cluster services stopped")
 
         # TODO: optionally stop services on all nodes or specific node


liangxin1300:
but stop pacemaker will also cause corosync stopped

vvidic:
On Sun, Oct 08, 2017 at 10:29:00AM +0000, Xin Liang wrote:
> but stop pacemaker will also cause corosync stopped

Can you elaborate, I don't see this happening on the Debian side:

# crm cluster stop
INFO: Cluster services stopped

# systemctl status pacemaker
● pacemaker.service - Pacemaker High Availability Cluster Manager
   Loaded: loaded (/lib/systemd/system/pacemaker.service; enabled; vendor preset: enabled)
   Active: inactive (dead) since Sun 2017-10-08 14:39:37 UTC; 4s ago
     Docs: man:pacemakerd
           http://clusterlabs.org/doc/en-US/Pacemaker/1.1-pcs/html/Pacemaker_Explained/index.html
  Process: 795 ExecStart=/usr/sbin/pacemakerd -f (code=exited, status=0/SUCCESS)
 Main PID: 795 (code=exited, status=0/SUCCESS)
      CPU: 571ms

# systemctl status corosync
● corosync.service - Corosync Cluster Engine
   Loaded: loaded (/lib/systemd/system/corosync.service; disabled; vendor preset: enabled)
   Active: active (running) since Sun 2017-10-08 14:38:45 UTC; 1min 9s ago
     Docs: man:corosync
           man:corosync.conf
           man:corosync_overview
 Main PID: 438 (corosync)
    Tasks: 2 (limit: 4915)
   Memory: 48.9M
      CPU: 669ms
   CGroup: /system.slice/corosync.service
           └─438 /usr/sbin/corosync -f

-- 
Valentin

liu4480:
> but stop pacemaker will also cause corosync stopped
@liangxin1300 
this is because SUSE uses pacemaker to manager corosync(which is disabled by default) by dependency, with this line:

StopWhenUnneeded=yes

and if you enable corosync, crm cluster stop will not stop corosync at all, especially for other Linux distribution, this can be true since we have different service files.

vvidic:
On Mon, Oct 09, 2017 at 02:33:23AM +0000, Bin Liu wrote:
> this is because SUSE uses pacemaker to manager corosync(which is
> disabled by default) by dependency, with this line:
> 
> StopWhenUnneeded=yes
> 
> and if you enable corosync, crm cluster stop will not stop corosync at
> all, especially for other Linux distribution, this can be true since
> we have different service files.

Ok, I understand now and will try to handle this on the Debian side
than.

-- 
Valentin

PR #287, opened 2017-10-08
low: script: drbd: drbd also have some advanced params at Hawk

liangxin1300:
* drbd ocf script has some optional params,
  like adjust_master_score/stop_outdates_secondary/ignore_missing_notifications etc.,
  these should be shown as "Advanced" params in wizards at Hawk
* In wizard config page, for the simple reason, it's better to just keep required params,
  move all un-required params at "Advanced" select options.
  Less options, quick configure:)
* Leave these params' default value empty is right, these values are mutative

### DIFF
--- a/scripts/drbd/main.yml
+++ b/scripts/drbd/main.yml
@@ -7,20 +7,20 @@ longdesc: >-
   Also creates a multistate resource managing the state of DRBD.
 
   Does not create or modify the referenced DRBD configuration.
+include:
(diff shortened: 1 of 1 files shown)


liangxin1300:
Oh, travis-ci failed beacuse ocf:linbit:drbd not exits

krig:
Ah, this isn't right: Using `include` will create another primitive, so there will be two `ocf:linbit:drbd` primitives.

PR #294, opened 2017-10-30
medium: NewCommit: ui_configure: complete for ra actions

liangxin1300:
for now, it's not easy to use crmsh to add a primitive,
user must remeber add monitor with many attributes(role/interval/timeout),
so, it's nice to integrate these into op completer.

Changes include:
  * append action items which in agent default actions
    e.g. interval and timeout
  * monitor_Master will be mapped to "monitor role=Master"
  * monitor_Slave will be mapped to "monitor role=Slave"
  * remove action items which not in default actions
   e.g. promote/demote/notify do not exist in normal ra op list
  * remove action items which already completed
  * make sure all of default items can be completed


krig:
It seems this breaks the edit regression test. Hmm, I need to fix the Travis configuration so that it prints the output from the test case on failure...

liangxin1300:
haha, this commit passed test,
krig, please help me to review this when you are available:)

Regards, 
xin

krig:
Alright, perfect :) Travis should print the details when tests are failing now too, so it will be easier to figure out ;)

PR #296, opened 2017-10-30
low: bootstrap: change multi-heartbeats as option-mode

liangxin1300:
With "-M" option after cluster init, user can add a second connection line

Regards,
xin

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -952,6 +952,7 @@ Configure Corosync:
 
         if i == 1 or \
            len(default_networks) == 1 or \
+           not _context.second_hb or \
            not confirm("\nConfigure a second multicast ring?"):
             break
         two_rings = True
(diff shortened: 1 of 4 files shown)


liangxin1300:
Append commit: medium: enable add the second heartbeat line for unicast
Changes inlcude:
  1. IPv4 & IPv6 support
  2. user should specify local IP for ringX_addr when init and join both
  3. user input must be one of local IP addresses
  4. peer ringX_addr and local ringX_addr must in the same network

liangxin1300:
> user should specify local IP for ringX_addr when init and join both

It's better to specify  local IP address for ring0 and ring1, because the net interface may contain multi addresses;

> user input must be one of local IP addresses

program will give hints when user input address is a valid IP or out of local addresses range,
e.g. "Must one of ['10.67.18.134', '10.10.10.3', '20.20.20.3']";
       "10.10.10.3 has been used"(when used for ring0, input for  ring1 again)

> peer ringX_addr and local ringX_addr must in the same network

e.g node1 have ring0: 10.10.10.1
                          ring1: 20.20.20.1
      node2 must in the same network and in the same order
                          ring0: 20.20.20.2
                          ring1: 10.10.10.2
      is wrong.

PR #297, opened 2017-10-30
medium: ui_configure: configure->delete is dangerous, better to confirm

liangxin1300:


dmuhamedagic:
On Mon, Oct 30, 2017 at 01:30:31AM -0700, Xin Liang wrote:
>   * medium: ui_configure: configure->delete is dangerous, better to confirm

I think I'd disagree here. What's dangerous is commit. It also
changes the existing behaviour. People tend to get annoyed if
asked too many questions, at least I do ;-)

PR #298, opened 2017-10-30
low: ui_context: reset term when using "help command+ctrlC"

liangxin1300:
If dont do this, type "ctrl+C" when using "help command" will cause
  the term lost curses

### DIFF
--- a/crmsh/command.py
+++ b/crmsh/command.py
@@ -414,6 +414,7 @@ Examples:
         """usage: help topic|level|command"""
         h = help_module.help_contextual(context.level_name(), subject, subtopic)
         h.paginate()
+        context.command_name = ""
 
     def get_completions(self):
         '''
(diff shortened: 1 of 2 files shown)


liangxin1300:
Oh, it seems this change blocked on regression test.
Krig, I think you can skip these failed commits and continue doing the release work for factory,
we've done most of work after all.

krig:
No problem! I think if you rebase on master now, Travis should print the details for the failure.

liangxin1300:
My next plan for crmsh: bugfix, bundle, add some scripts for hawk wizard(like redis/nginx/tomcat/pgsql) :)

krig:
Sounds good! Having some new scripts would be great :)

liangxin1300:
This commit also passed test:)

Thanks!

PR #299, opened 2017-10-30
medium: ui_cluster: override OptParser.error to keep stay at shell wh…

liangxin1300:
…en error happend


liangxin1300:
append commit: medium: ui_cluster: valid clusters option for geo setup
* Geo cluster must have two sites
* valid the IP

liangxin1300:
append commit: medium: ui_cluster: valid cluster-node option for geo arbitrator setup

krig:
You should be able to `raise ValueError` on fatal errors which gets caught and printed at the interactive prompt. Maybe that can make the code simpler.

PR #302, opened 2017-11-01
medium: ui_context: Stop completing when an id is required

liangxin1300:
Hi krig,
   I have an idea popped when doing bundle work.
   "id" is required in many places, 
   e.g. in do_primitive, we can replace compl.null to compl.attr_id, 
   and stop completing when meeting "id=",
   this can make configure process more fluent because compl.attr_id give hints for users:)

   But, this can lead errors in test or in hawk, we can make it portable no matter "id=" is used or not.

   If this commit make sense, I will continue to do the replace work and port work:)

   Regards,
   xin

### DIFF
--- a/crmsh/completers.py
+++ b/crmsh/completers.py
@@ -16,7 +16,7 @@ def choice(lst):
 
 
 null = choice([])
-
+attr_id = choice(["id="])
 
 def call(fn, *fnargs):
     '''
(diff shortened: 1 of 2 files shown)


liangxin1300:
I'm confused why regression failed :(
I just define compl.attr_id, not even use it

krig:
Hi Xin, try rebasing on the latest `master` and force push to this branch again. It's probably something unrelated!

liangxin1300:
Thanks!

PR #304, opened 2017-11-03
medium: bootstrap: check addr/hostname in local's /etc/hosts when joi…

liangxin1300:
…ning

When nodeB want to join nodeA, it's better to check whether nodeB in /etc/hosts(local),
this can avoid csync2 error later which may hard to debug

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -347,6 +347,12 @@ def check_etc_hosts(node):
     return False
 
 
+def check_peer_etc_hosts(peer, node):
+    if invoke("ssh root@{} \"grep -E '(\s|^){}(\s|$)' /etc/hosts\"".format(peer, node)):
+        return True
+    return False
+
+
 def grep_output(cmd, txt):
     _rc, outp, _err = utils.get_stdout_stderr(cmd)
     return txt in outp
@@ -2048,6 +2054,8 @@ def bootstrap_join(cluster_node=None, ui_context=None, nic=None, quiet=False, ye
             _context.cluster_node = cluster_node
 
         join_ssh(cluster_node)
+        if not check_peer_etc_hosts(cluster_node, utils.this_node()):
+            error("Need configure \"{}\" in /etc/hosts on {} if want to join it".format(utils.this_node(), cluster_node))
         join_csync2(cluster_node)
         join_ssh_merge(cluster_node)
         join_cluster(cluster_node)
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -339,6 +339,14 @@ def check_tty():
         error("No pseudo-tty detected! Use -t option to ssh if calling remotely.")
 
 
+def check_etc_hosts(node):
+    with open("/etc/hosts") as f:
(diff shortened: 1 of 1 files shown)


liangxin1300:
Append commit: medium: bootstrap: check addr/hostname in peer's /etc/hosts when joining

For the same reason, it's better to make sure nodeB in nodeA's /etc/hosts file

krig:
Ah no, this is not a good idea. It is not a strict requirement that peers are configured in /etc/hosts. It is possible to use an internal DNS server for example.

zzhou1:
I do urge to improve User Experience.

I rush to run ha-cluster-join before preparing /etc/hosts. The tool exits with error and  left a mess state behind. I don't have immediate clue where to recover. Naturally, I don't want to redo all those previous steps of the system.

The error handling can be smarter: 

a) try to diagnose the environment how hostname can be resolved, and even add addr/hostname if suitable.

b) the minimum expectation is to protect a clean state, don't let additional steps to leave a mis-configuration behind. User will feel much better to continue from where it fails. 

Redo from the scratch hurts user experience a lot ;)

zzhou1:
Here is the error before I prepare a valid /etc/hosts.

sle15sp0-2:~ # ha-cluster-join
! Failed to detect firewall: Could not open ports tcp=30865|5560|7630|21064, udp=
  Join This Node to Cluster:
  You will be asked for the IP address of an existing node, from which
  configuration will be copied.  If you have not already configured
  passwordless ssh between nodes, you will be prompted for the root
  password of the existing node.

  IP address or hostname of existing node (e.g.: 192.168.1.1) []192.168.100.101
  Retrieving SSH keys - This may prompt for root@192.168.100.101:
Password:
/root/.ssh/id_rsa already exists - overwrite (y/n)? y
  One new SSH key installed
  Configuring csync2...
! csync2 run failed - some files may not be sync'd
done
  Merging known_hosts
! Failed to get known_hosts from sle15sp0-1: Exited with error code 255, Error output: ssh: Could not resolve hostname sle15sp0-1: Name or service not known

ERROR: 1: cluster.join: [Errno 2] No such file or directory: '/etc/corosync/corosync.conf'

krig:
Well, there are two problems I think:

1. Missing to configure `/etc/hosts` can cause problems later. I think the best thing we could do here is to try to verify that all the nodes are reachable by hostname before doing anything else, and aborting if the configuration seems incorrect. To configure `/etc/hosts` automatically was something that the bootstrap scripts used to do, but it was removed due to the issues it caused. So any solution would need to solve the reasons why that code was removed too. That was done before I took over the scripts, so I don't know the exact reason. That's why I am a bit careful.

2. csync2 is very fragile and it is difficult to recover once it has problems. The best solution here is to stop using csync2 and use a more robust method for synchronizing the configuration between the nodes. In fact, csync2 should not be necessary, we should be able to use the parallax library to copy all the needed configuration files from the existing cluster nodes. So I would prefer to remove the use of csync2 to solve this.

liangxin1300:
I suggest that we do some check at very beginning like ping self hostname, if faild, exit;
we don't need to change /etc/hosts

PR #306, opened 2017-11-08
low: doc: change usage and doc for group

liangxin1300:
1. In http://clusterlabs.org/doc/en-US/Pacemaker/1.1-pcs/html-single/Pacemaker_Explained/index.html#_changed, "The attributes container tag was removed"

2. According to http://clusterlabs.org/doc/en-US/Pacemaker/1.1-pcs/html-single/Pacemaker_Explained/index.html#group-resources,
These is no params, so I delete it from usage

### DIFF
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -742,7 +742,6 @@ class CibConfig(command.UI):
     @command.completers_repeating(compl.attr_id, _group_completer)
     def do_group(self, context, *args):
         """usage: group <name> <rsc> [<rsc>...]
-        [params <param>=<value> [<param>=<value>...]]
         [meta <attribute>=<value> [<attribute>=<value>...]]"""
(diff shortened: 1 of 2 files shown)


krig:
2. is not correct, you have misunderstood the statement :) It means that params set on the group are applied to all members of the group.

krig:
Actually 1. is also not correct, the comment is referring to an `attributes` tag, tag meaning XML tag (like nvpair or instance_attributes), not to the `container` attribute.

krig:
For 1., it is very poorly documented unfortunately, perhaps it is worth improving. Anyway, some hints about the feature can be found in the code:

https://github.com/ClusterLabs/pacemaker/search?utf8=%E2%9C%93&q=XML_RSC_ATTR_CONTAINER&type=

liangxin1300:
Thank you for explanation:)

PR #307, opened 2017-11-08
low: cibconfig: use refresh instead of reset after commit

liangxin1300:
if use reset, for example, after adding a valid resource and commit it,
without leaving the configure level,
delete completer will not get the id to delete


liangxin1300:
hi, krig, please consider this commit, without it, maybe will cause confuse sometime:)

PR #308, opened 2017-11-09
medium: filter exist args

liangxin1300:
Apply this commit to avoid this completion situation:

1 primitive vip ocf:Heartbeat:IPaddr2 params ip=10.10.10.123 nic=eth1 nic=xxxx ip=xxxx
2 primitive vip ocf:Heartbeat:IPaddr2 meta is-managed=xxx is-managed=xxx

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -42,6 +42,14 @@ def to_ascii(s):
         return s
 
 
+def any_startswith(iterable, prefix):
+    """Return first element in iterable which startswith prefix, or None."""
+    for element in iterable:
+        if element.startswith(prefix):
+            return element
+    return None
+
+
 def memoize(function):
     "Decorator to invoke a function once only for any argument"
     memoized = {}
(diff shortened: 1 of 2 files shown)


crmsh/utils.py
```diff
@@ -42,6 +42,14 @@ def to_ascii(s):
...
 
 
+def completed(item, args, use='='):
+    """Check whether item has been completed in args"""
+    for arg in args:
+        if arg.startswith(item+use):
+            return True
+    return False   
+
+
```
krig: how about a more general function `any_startswith`:

```
def any_startswith(iterable, prefix):
    """Return first element in iterable which startswith prefix, or None."""
    for element in iterable:
        if element.startswith(prefix):
            return element
    return None
```

and then use like this:

```
return [s+'=' for s in constants.rsc_meta_attributes if utils.any_startswith(args, s + '=') is None]
```

liangxin1300:
Thank you, it looks better:)

liangxin1300:
Hi krig, I've changed the code in local days ago, but I forgot to push it:)

krig:
Looks good, thanks :)

PR #311, opened 2017-11-13
Low: ui_context: Continue completing when input is an alias

liangxin1300:
For example:
   "master" is the alias name of "ms",
   input "crm configure master" will not complete, until input a space ' ';
   use this commit, "crm configure master" will continue completing after input "Tab"
  
This can make using crmsh more fluently:)

Regards,
xin


krig:
Ah, thank you! I have noticed that before but never got around to making a fix.

PR #312, opened 2017-11-14
low: ui_configure: required params first

liangxin1300:
When completing, complete required parameters firstly, which is more important.
When all of required parameters completed, then complete non-required parameters.
It's clear and can reduce configure errors

Regards,
xin

### DIFF
--- a/crmsh/ra.py
+++ b/crmsh/ra.py
@@ -390,7 +390,7 @@ class RAInfo(object):
         self.broken_ra = False
         return self.ra_elem
 
-    def params(self, completion=False):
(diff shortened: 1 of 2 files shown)


krig:
Hmm, this seems overly complicated and strange to me, I don't think it is appropriate for completion to contain complicated logic like this. It won't be transparent to the user. What if the user is trying to remember the name of a particular parameter?

liangxin1300:
I have tried to mark required parameters with colors, but not work

krig:
Yeah, I don't think that would work with bash completion either...

PR #313, opened 2017-11-14
low: utils: convert bytes to str(bsc#1067823)

liangxin1300:
bsc#1067823 still not work
apply this commit, some errors disappeared, 
but some strange status shown in wizards page
e.g. I use wizards page to install apache2 and configure vip,
apache2 package can be successfully installed,
the resource group also started successfully,
but some strange status like: b''
b'true\n'
b'"status" already present\n'
shown in page

when I use command line "crm script run apache virtual-ip:id=vip virtual-ip:ip=10.10.10.123 virtual-ip:cidr_netmask=24 install=true id=fs"
it works well


krig:
The issues may be with the `json` command.

PR #315, opened 2017-11-15
medium: scripts: make sure gfs2 can be configured using hawk(bsc#1067…

liangxin1300:
…123)

First, we can use ocfs2 script as template, make sure gfs2 can be configured in hawk;

Then there are two questions need to discuss:
1. we can add clvm2 as non-required step both in ocfs2 and gfs2, as what apache script does;
2. we can integrate mkfs.ocfs2 and mkfs.gfs2 commands in crmsh or/and hawk so that
    user do not have to remember to run these commands with many options manually.

### DIFF
--- a/scripts/gfs2/main.yml
+++ b/scripts/gfs2/main.yml
@@ -9,8 +9,6 @@ longdesc: >-
   The file system should be on the device, unless cLVM is used.
 
 category: File System
-include:
-  - script: gfs2-base
 parameters:
   - name: id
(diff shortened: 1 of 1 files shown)


krig:
For part 1, perhaps you could discuss this with zren since there are changes regarding lvmlockd. Maybe we need to rewrite all the scripts using LVM.

2. I agree, though we have to make sure the user would not lose data accidentally.

PR #316, opened 2017-11-16
low: ui_configure: Improve do_clone completer

liangxin1300:
We should complete clone meta keywords like "interleave" because user wouldn't remember these.

BTW, It' nice that if we add some help messages for meta attributes like primitive params,
give some hints for suggestion value


krig:
Nice, thank you!

I'm not sure about help in completions, I've always thought that was a bit of an odd feature and it doesn't work well with bash completion :(

PR #323, opened 2017-11-30
low: ui_configure: fix for 309d2e, remove "id=" in a save way

liangxin1300:
When inputs like: id=ids, all 'id' words will be stripped


krig:
Good catch, yes

liangxin1300:
hi krig, I have created a PR in https://github.com/krig/salt-module-crmsh/pull/1, which project maybe you need add watching:)

PR #328, opened 2017-12-11
low: ui_configure: use filter_keys replace any_startswith

liangxin1300:
to keep code simple and clean,
and to improve _advanced_completer for group/clone/ms resource type

### DIFF
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -57,6 +57,7 @@ _node_id_list = compl.call(cib_factory.node_id_list)
 _rsc_template_list = compl.call(cib_factory.rsc_template_list)
 _container_type = compl.choice(constants.container_type)
 
+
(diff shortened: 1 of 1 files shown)


crmsh/ui_configure.py
```diff
@@ -57,25 +57,28 @@ def _tag_completions():
...
     meta completers for group/ms/clone resource type 
     '''
     key_words = ["meta", "params"]
     completing = args[-1]
     resource_type = args[0]
+    return_list = []
     if completing.endswith('='):
         # TODO add some help messages
-        return []
+        return return_list
```
krig: `return_list` is always `[]` here. Why make this change? It seems confused to me.

liangxin1300: I've changed it back:)

krig:
OK :)

PR #330, opened 2017-12-14
medium: ui_configure: Improve location completer

liangxin1300:
Hi, krig:

This is a draft commit.
This commit can make adding location command more readable, like:
`location id=loc1 resource1 on node1 score=100`
with score help message to explain what score value means for location.

Todo list include:
rules completer;
parse process;
configure show;
and document change.

My purpose to do this primarily is a user complain he had trouble with configuring constraints,
do you remember that? So, I think we should at least do some improvement for completer of location/order/colocation, to lead user configure constraints more easily.

Regards,
xin

### DIFF
--- a/crmsh/constants.py
+++ b/crmsh/constants.py
@@ -473,4 +473,14 @@ container_helptxt = {
     }
 }
 
+score_helptxt = """
+score:
+valid value include: (integer)|INFINITY|+INFINITY|-INFINITY
+
+Positive values indicate a preference for running the affected resource(s)
+on this node — the higher the value, the stronger the preference. Negative
+values indicate the resource(s) should avoid this node (a value of -INFINITY
+changes "should" to "must").
+"""
+
 # vim:ts=4:sw=4:et:
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -80,6 +80,26 @@ def _advanced_completer(args):
     return return_list + key_words
 
 
+def _location_completer(args):
+    completing = args[-1]
+    token = args[-2]
+    if len(args) == 4:
+        return ["on"]
+    if len(args) == 5:
+        return cib_factory.node_id_list()
+
+    if completing.endswith('='):
+        if len(completing) > 1 and options.interactive:
+            topic = completing[:-1]
+            if topic == "score":
+                CompletionHelp.help(topic, constants.score_helptxt)
+        return []
+
+    if len(args) == 6:
+        return ("score=", 1)
+
+
+
 def _list_resource(args):
     if len(args) > 3:
         if args[2] == "remove":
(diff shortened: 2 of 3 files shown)


krig:
Hm, just wondering: The syntax of location is

`location loc1 resource1 100: node1`

not

`location id=loc` resource1 on node1 score=100`

I only see changes to the completer.. or you mean that you want to change the syntax?

I don't think we should change the syntax, this would only confuse readers and require changing documentation (imagine random tutorials on the internet etc. that would still use old syntax).

However, maybe we can provide a second syntax which is easier to use. In that case, I think it is possible to go further... Maybe something like

`location resource1 on node1 score=100`

The ID could be implicit and generated, like: `resource1-on-node1`

liangxin1300:
Currently, location command line like:
`location id=loc1 dummy 100: salt-2` ("id=", see #303, will be strip before parse process )

configure show:
`location loc1 dummy 100: salt-2`

Now, I do want to change the synatx as:
`location id=loc1 resource1 on node1 score=100`

and configure show like:
```
location loc1 resource1 on node1 
              score=100
```

This is more clear to understand, why not change? 
Dropping old syntax like dropping python2 is nature for improvement.

Also, old syntax of order like:
`order ord1 Mandatory dummy dummy2`
vs:
`order ord1 dummy before dummy2`
The second one is more clear and easy to understand.

Regards,
xin

krig:
Changing the syntax requires users to learn something new, to update their own notes, and to forget old knowledge. It also requires documentation changes both upstream and downstream.

For a mature project, this is a big change. To motivate such a big change, there are some requirements:

* The old syntax must still be parsed and converted to the new syntax
* The change must be a big enough improvement to motivate inflicting pain on users
* The change must be documented clearly.

Moving from Python 2 to 3 is also a
major change, and should only be done with care. It is already causing pain in the user community. But it was necessary to do so since Python 2 will become unmaintained soon.

This is my thinking: Changing the syntax is fine, but the change must repay its cost by being significantly better. I agree that the proposed new syntax is better, but I think it can be even better still.

When changing the syntax for constraints, lets consider the syntax of all constraints at the same time, and see if we can make it more consistent, for example.

liangxin1300:
OK, this thing need more well thinking and well design, with all of the constraints, also with rules.

I will write a first version for that, in days. 

Regards,
xin

krig:
There is one problem which we should try to fix with new syntax: colocation and order constraints are not symmetrically defined. To create a common type of constraint which is also the constraint implemented by groups, a user creates two constraints: a colocation constraint, and an order constraint. This is to define simple dependency: Two resources have to start on the same node, and in a certain order.

To do this now, one needs to define these constraints:

```
colocation c resource2 resource1
order o resource1 resource2
```

This is because the colocation says "place resource2 with resource1", meaning resource1 is placed first.
The second constraint says "start resource1 before resource2", meaning resource1 is started first.

This is confusing and difficult to understand. In fact, a recent public case where HA looked bad was due to this confusion: https://gocardless.com/blog/incident-review-api-and-dashboard-outage-on-10th-october/

If we can make this clearer, that would be good. The colocation constraint seems "backwards" to me, but I find it difficult to express colocation in a different order in the english language.

PR #331, opened 2018-01-02
fix: ui_resource: Using crm_failcount(bsc#1074127)

liangxin1300:


krig:
Great, thanks!

krig:
Hi Xin,

I am also new to salt and still not sure what the best approach to using it for HA is. I don't see anything fatal with the experiment, but at the same time, is there improvement by using salt for this part of bootstrap? Maybe that is the question to ask :)

Cheers,
Kristoffer

liangxin1300:
What we want is using salt to replace the whole bootstrap.py, is that right?

For now, what I have done is about configure corosync and start pacemaker service;
There are still many functions need to map from origin bootstrap.py into salt.

We can certainly get benefits from this transformation(less code, more flexible and stable, easy to maintain, easy to integrate with other applications)

krig:
What I mean is that maybe we can make bootstrap better by rethinking how it works, not just making it use salt to do the same thing.

My thinking is that a more declarative approach would be better in the long run:

Instead of a command "bootstrap cluster with admin IP and SBD and OCFS2" as a one-shot command, apply a cluster description: "the cluster should have an admin IP and consist of two nodes". Then changes to the configuration would be made as changes to the description. Then apply the new description.

This is more like how salt works.

Then, the bootstrap would only be to configure the salt master and minions to prepare for this..

So...

- Setup salt master to manage HA clusters
- Start and connect salt minions to master
- Apply cluster orchestration / state

PR #333, opened 2018-01-03
low: ra: Strip repeated codes

liangxin1300:
So that when using "ra validate fdsf:fdsfl:fsf", command line can pop error message:)


krig:
Hmm, I don't think this is a good change: Better to change `do_validate` to check the return code; if it is `< 0` `do_validate` could report that the agent was not found or does not have the validate action.

liangxin1300:
But can_validte_agent function do the same thing with other parts of code in find_agent function

PR #334, opened 2018-01-03
low: ui_configure: Complete rsc template correctly

liangxin1300:
For now, configure->primitive->xxx->@ will not complete rsc template names;
It's better to show them directly:)


krig:
Looks good, yes :)

PR #336, opened 2018-01-04
fix: test: 2.0 branch has dropped "requires"

liangxin1300:
in pacemaker 2.0 branch, commit 014a543d53d4cc31739a618279685357862749a0

### DIFF
--- a/test/testcases/history.exp
+++ b/test/testcases/history.exp
@@ -5,7 +5,7 @@
 .EXT tar -tj < history-test.tar.bz2 2> /dev/null | head -1
 .EXT tar -xj < history-test.tar.bz2
 Source: history-test.tar.bz2
-Created on: Fri 14 Dec 19:08:38 UTC 2012
+Created on: Sat 15 Dec 03:08:38 CST 2012
 By: unknown
 Period: 2012-12-14 20:06:34 - 2012-12-14 20:08:44
 Nodes: xen-d xen-e
--- a/test/bugs-test.txt
+++ b/test/bugs-test.txt
@@ -1,6 +1,5 @@
 node node1
 primitive st stonith:null params hostlist=node1
-property default-action-timeout=60s
 group g1 gr1 gr2
 group g2 gr3
 group g3 gr4
--- a/test/crm-interface
+++ b/test/crm-interface
@@ -33,7 +33,6 @@ primitive p3 ocf:pacemaker:Dummy
 primitive st stonith:null params hostlist=node1
 clone c1 p1
 ms m1 p2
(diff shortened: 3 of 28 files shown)


liangxin1300:
Hi krig,
    I'm trying to change related testcases which pacemaker-2.0 has dropped.
    Should we merge these commits into master or some other branch?
    And, I think you should update pacemaker packages(https://build.opensuse.org/package/show/home:yan_gao:branches:network:ha-clustering:Factory:Test/pacemaker) for travis-CI

liangxin1300:
Append commit: fix: test: 2.0 branch has dropped "default-action-timeout"

krig:
Hi Xin, instead of removing the use of `default-action-timeout`, replace it with `op_defaults timeout=<value>`

krig:
Also, maybe it is better to make the changes to the tests first only, not to the expected output. Then run the tests in the Travis environment to get new expected output.

I will need to update the base docker image to have the new version of Pacemaker, I will do that.

test/testcases/history.exp
```diff
@@ -5,7 +5,7 @@
 .EXT tar -tj < history-test.tar.bz2 2> /dev/null | head -1
 .EXT tar -xj < history-test.tar.bz2
 Source: history-test.tar.bz2
-Created on: Fri 14 Dec 19:08:38 UTC 2012
+Created on: Sat 15 Dec 03:08:38 CST 2012
```
krig: This is a strange change, why has it switched from UTC to local time zone? Hrm..

liangxin1300: I just use “regression.sh prepare” to generate it

krig: Might have to set the locale or TZ in the container

krig:
Hi Xin, apologies, but I will close this pull request since it is now outdated. I updated the test cases to remove use of `requires` as an operation attribute (replaced with `requires` as meta attribute) and use of `default-action-timeout` with `op_defaults timeout`.

krig:
Actually I don't use `regression.sh prepare` when updating the expected test case output. I run the tests as usual inside the container using `test/containerized-regression-tests.sh` and then use a small script to update them like this:

```
#!/bin/bash
crmtestout="$1"

for f in $crmtestout/*.diff; do
	fil=$(grep -- --- $f | awk '{print $2}' | sed 's/\/usr\/share\/crmsh\/tests/\/test/g')

	cat $f | awk "NR==1{\$2=\"a$fil\"}1" | awk "NR==2{\$2=\"b$fil\"}1"


done
```

Maybe I should add this script to the `test` folder.

PR #338, opened 2018-01-08
high: bootstrap: Add QDevice/QNetd support (bsc#1070961)

krig:


crmsh/corosync.py
```diff
@@ -584,9 +587,39 @@ def create_configuration(clustername="hacluster",
...
+    two_node: 0
+    device {
+      votes: 1
+      model: net
+      net {
+        tls: off
+        host: %(ip)s
+        port: %(port)s
+        algorithm: %(algo)s
+        tie_breaker: %(tie_breaker)s
```
liu4480: the tie_breaker should be:

auto_tie_breaker: 1
auto_tie_breaker_node: lowest

auto_tie_breaker = 1 mean to enable auto_tie_breaker, this option is disabled by default, we should enable it.
while auto_tie_breaker_node is the preference that qnetd to decide who should have the votes if there is brain split

crmsh/corosync.py
```diff
@@ -57,6 +57,14 @@ def __repr__(self):
         return '%s: %s' % (self.key, self.value)
 
 
+class QDevice(object):
+    def __init__(self, ip, port=5403, algo="ffsplit", tie_breaker="lowest"):
+        self.ip = ip
+        self.port = port
+        self.algo = algo
+        self.tie_breaker = tie_breaker
```
liu4480: the tie_breaker should be:

auto_tie_breaker: 1
auto_tie_breaker_node: lowest

auto_tie_breaker = 1 mean to enable auto_tie_breaker, this option is disabled by default, we should enable it.
while auto_tie_breaker_node is the preference that qnetd to decide who should have the votes if there is brain split

crmsh/bootstrap.py
```diff
@@ -1742,16 +1749,27 @@ def update_expected_votes():
...
+                nodecount = int(v) + 1 + qnetd_votes
                 corosync.set_value("quorum.expected_votes", str(nodecount))
                 corosync.set_value("quorum.two_node", 1 if nodecount == 2 else 0)
         else:
-            nodecount = len(nodelist)
-            corosync.set_value("quorum.expected_votes", str(nodecount))
+            nodecount = len(nodelist) + qnetd_votes
+            if corosync.get_value("quorum.expected_votes"):
+                corosync.set_value("quorum.expected_votes", str(nodecount))
             corosync.set_value("quorum.two_node", 1 if nodecount == 2 else 0)
```
liu4480: two_node can not be enable when using qdevice, even if the real nodes are two

crmsh/bootstrap.py
```diff
@@ -1742,16 +1749,27 @@ def update_expected_votes():
         # Increase expected_votes
         # TODO: wait to adjust expected_votes until after cluster join,
         # so that we can ask the cluster for the current membership list
+        # Have to check if a qnetd device is configured and increase
+        # expected_votes in that case
+        qnetd_votes = 1 if corosync.get_value("quorum.device.model") == "net" else 0
```
liu4480: Hi, Kristoffer 
Sorry to make qdevice unclear. 

in fact the qnetd_votes is the same thing with device_votes, 

without qdevice, normally expected_vote is number of nodes.
assume we have a cluster with N nodes. expected_votes will be N without qdevice.
But qdevice, expected_vote is sum of "number of nodes" and device_votes.

there are two algorithms so far, ffsplit and lms

for ffsplit,  device_votes is always on, so the expected_votes is N + 1.
as described on bsc#1070961 comment#15, we will not add the device_votes to expected_votes at first init,
but  on the first join(indeed the second node join the cluster) by judging "device_votes==0 and expected_votes==1", we can set  device_votes to 1, while adding device_votes to expected_votes only this time.

liu4480: while for lms, device_votes is N - 1, and the expected_votes is (N-1) + N
we can get number of nodes: num = (expected_votes+1)/2,  
and increase number of nodes: num += 1
get the new  device_votes:  device_votes = num - 1
and get the new expected_votes:  expected_votes = (num -1) + num

this is recalculated every time a node joins

liu4480:
@krig since beta6 is releasing, is there anything block this PR to be merged

krig:
No, lets merge it now.

PR #343, opened 2018-01-11
medium: ui_cluster: Stop corosync when stopping pacemaker (bsc#1066156)

krig:


krig:
@gao-yan @liu4480 Is this what you meant? Start only pacemaker, but stop both.

liu4480:
@krig  stoppting both could work, but what I mean is stop only by using "service corosync stop", pacemaker will also stop

gao-yan:
On 2018/1/11 9:42, Kristoffer Grönlund wrote:
> @gao-yan <https://github.com/gao-yan> @liu4480 
> <https://github.com/liu4480> Is this what you meant? Start only 
> pacemaker, but stop both.
Or simply stop corosync ;)

krig:
Hm, I see :) OK, will do that instead.

PR #345, opened 2018-01-18
high: scripts: ra validate with params for hawk verify

liangxin1300:
hi krig:
     In https://github.com/ClusterLabs/hawk/issues/134, I think we should find a way detecting whether application installed;
     So, this is one option, using "crm ra validate + ra_params" to do the detacting, then, if application not installed locally, click "Verify" button in hawk wizards will give error messages instead of success messages.
    BTW, there is a sudo problem which because in hawk lib/crm_script.rb, cmdline was called by hawk_invoke, so crm ra validate command will not trigger without sudo. Then, I had to add "hacluster ALL=(ALL) NOPASSWD: ALL" into /etc/sudoers to avoid password input in hawk web console.
     Is there other better solution to make this permissions problem work?
     
Regards,
xin

### DIFF
--- a/crmsh/scripts.py
+++ b/crmsh/scripts.py
@@ -2178,6 +2178,28 @@ def verify(script, params, external_check=True):
 
     Return a list of actions to perform.
     """
+    def do_validate(params):
+        result = ""
+        for item in params:
+            if re.search("primitive ", item):
+                result = ' '.join(item.split())
+                break
+        if not result:
+            return None
+
+        err_res = []
+        pattern = re.compile(r'(primitive .*?) op')
+        m = pattern.findall(result)
+        for item in m:
+            _res = item.split()
+            if "params" in _res:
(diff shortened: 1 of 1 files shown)


krig:
I added this before, but had to remove it due to several issues:

* Sudo is required to run validate but not to run crmsh so running validate is not an option
* Validate requires full runtime environment including dependent resources: For example, to validate a database that stores its data on a file system mounted using the Filesystem resource, it is necessary to first mount the Filesystem resource and only then run validate.

So this cannot work unfortunately.

liangxin1300:
OK, well, alternatively we should find a temporary suitable way for detecting whether application installed, before we using SALT to figure this out

krig:
Using salt is exactly the approach we should take! Since salt can be idempotent, it can simply ask for the application to be installed, and it would be installed if this is required.

PR #346, opened 2018-01-24
fix: hb_report: collect sbd info(bsc#1076389)

liangxin1300:
Changes:
- sbd rpm version in sysinfo.txt
- collect /etc/sysconfig/sbd
- add some new packages in constants.PACKAGES

Thanks for reviewing!


hb_report/constants.py
```diff
@@ -74,12 +74,11 @@
 start_stop Configuration.validated..Starting.heartbeat|Corosync.Cluster.Engine|Executive.Service.RELEASE|Requesting.shutdown|Shutdown.complete
 """
 
-PACKAGES = """pacemaker libpacemaker3
-pacemaker-pygui pacemaker-pymgmt pymgmt-client
-openais libopenais2 libopenais3 corosync libcorosync4
+PACKAGES = """pacemaker libpacemaker3 pacemaker-cli pacemaker-remote
+corosync libcorosync4
```
liu4480: actually libcorosync4 was splited by opensuse review team [1] like:
libcfg6
libcmap4
libcorosync_common4
libcpg4
libquorum5
libsam4
libtotem_pg5
libvotequorum8

and we also have new packages:
corosync-qdevice
corosync-qnetd

[1]https://build.opensuse.org/request/show/523762
@krig @liangxin1300 I think you should also add these packages as well

hb_report/utillib.py
```diff
@@ -1439,6 +1440,25 @@ def say_ssh_user():
         return constants.SSH_USER
 
 
+def sbd_info():
+    """
+    sbd information
+    """
+    out_string = "#####SBD info:\n\n"
+    sbd_device = grep("(^|^\s*)SBD_DEVICE=.*", infile=constants.SBDCONF)
```
liu4480: the sbd device can support multiple devices up to 3, it is like: 
SBD_DEVICE="/dev/sda1;/dev/sdb1;/dev/sdc1"

this need to be considered as well

hb_report/utillib.py
```diff
@@ -1439,6 +1440,25 @@ def say_ssh_user():
...
+    if sbd_device:
+        device = sbd_device[0].split('=')[-1]
+        for cmd in ["dump", "list"]:
+            out_string += "## sbd {}:\n".format(cmd)
+            out_string += get_command_info("sbd -d {} {}".format(device, cmd))[1] + '\n'
+        shutil.copy2(constants.SBDCONF, constants.WORKDIR)
```
liu4480: just save the conf file if it exists. since sbd-1.3.0 or later supports diskless mode

hb_report/constants.py
```diff
@@ -74,12 +74,11 @@
...
 resource-agents cluster-glue libglue2 ldirectord libqb0
-heartbeat heartbeat-common heartbeat-resources libheartbeat2
```
krig: Please don't remove these: For debugging purposes it would be interesting to include information about these packages if they are installed even though they should not be.

krig: That applies for all old packages: Please don't remove them from the list!

hb_report/utillib.py
krig: Yes, I am not sure dumping the actual SBD device is a good idea for crm report... I would suggest to just save the configuration file in the report.

PR #348, opened 2018-02-06
fix: hb_report: Don't create *.log.info file if not in timespan

liangxin1300:
If not match any log content in given timeframe for a log file, 
collect this log.info is not necessary.

I will fire a bug if this commit make sense


krig:
Seems sensible to me, yes!

PR #349, opened 2018-02-06
fix: hb_report: Collect irregular log file

liangxin1300:
Collect irregular time-format log file such as ha-cluster-bootstrap.log.

I will fire a bug if this commit make sense

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -294,6 +294,12 @@ def collect_info():
         if l == constants.HA_LOG and l != constants.HALOG_F:
             os.symlink(constants.HALOG_F, os.path.join(constants.WORKDIR, os.path.basename(l)))
             continue
+        if is_our_log(l, constants.FROM_TIME, constants.TO_TIME) == 4:
+            log_warning("found irregular log file %s" % l)
+            outf = os.path.join(constants.WORKDIR, os.path.basename(l))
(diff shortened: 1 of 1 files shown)


krig:
We should definitely collect ha-cluster-bootstrap.log. But I wonder if this might collect too many files...

liangxin1300:
> But I wonder if this might collect too many files...

No, in code, this branch is from config.report.collect_extra_logs or from command-line option, 
so, hb_report will collect what user want it to collect:)

krig:
OK, great :)

hb_report/utillib.py
```diff
@@ -294,6 +294,12 @@ def collect_info():
         if l == constants.HA_LOG and l != constants.HALOG_F:
             os.symlink(constants.HALOG_F, os.path.join(constants.WORKDIR, os.path.basename(l)))
             continue
+        if is_our_log(l, constants.FROM_TIME, constants.TO_TIME) == 4:
+            log_warning("found irregular log file %s" % l)
```
krig: Should this really be a warning? Maybe info is more appropriate. After all we expect ha-cluster-bootstrap.log to be irregular, for example.

krig:
Merging as it is now, we can change warning to info later if that makes sense. Thanks!

PR #351, opened 2018-02-09
fix: hb_report: got the right file decompressor(bsc#1077553)

liangxin1300:
Command 'file /var/log/messages' return 'data', 
So, this will cause lose /var/log/messages :(

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -531,16 +531,13 @@ def events(destdir):
 
 
 def find_decompressor(log_file):
-    decompressor = "echo"
+    decompressor = "cat"
     if re.search("bz2$", log_file):
         decompressor = "bzip2 -dc"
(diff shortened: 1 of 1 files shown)


krig:
Hmm, "echo" can never be the right response from get_compressor, right?

Instead of doing this fix, how about just changing the default compressor to "cat" and always returning that if the input is not a compressed file?

krig:
I mean:

```
def find_decompressor(log_file):
    decompressor = "cat"
    if re.search("bz2$", log_file):
        decompressor = "bzip2 -dc"
    elif re.search("gz$", log_file):
        decompressor = "gzip -dc"
    elif re.search("xz$", log_file):
        decompressor = "xz -dc"
    return decompressor
```

liangxin1300:
OK!

PR #353, opened 2018-02-11
fix: bootstrap: Create pacemaker_remote authkey(#bsc 1077389)

liangxin1300:
Changed:
1. Using dd to create /etc/pacemaker/authkey when cluster->init
2. sync this key to other nodes when other nodes run cluster->join

other:
Maybe need copy this authkey to the remote node manually.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -37,6 +37,8 @@ COROSYNC_AUTH = "/etc/corosync/authkey"
 SYSCONFIG_SBD = "/etc/sysconfig/sbd"
 SYSCONFIG_FW = "/etc/sysconfig/SuSEfirewall2"
 SYSCONFIG_FW_CLUSTER = "/etc/sysconfig/SuSEfirewall2.d/services/cluster"
+PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
+
 
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin")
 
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -2138,11 +2163,22 @@ def bootstrap_join(cluster_node=None, ui_context=None, nic=None, quiet=False, ye
...
 
 
+def join_remote_auth(node):
+    invoke("rm -f {}".format(PCMK_REMOTE_AUTH))
+    pcmk_remote_dir = os.path.dirname(PCMK_REMOTE_AUTH)
+    if not os.path.exists(pcmk_remote_dir):
+        invoke("mkdir -p --mode=0750 {}".format(pcmk_remote_dir))
+        invoke("chgrp haclient {}".format(pcmk_remote_dir))
+
+    invoke('ssh root@{remote} "csync2 -mr {file}; csync2 -fr {file}; csync2 -rxv {file}"'.format(remote=node, file=PCMK_REMOTE_AUTH))
```
krig: This seems convoluted and unnecessary, the authkey should be included in the basic HA csync2 group so that it is synced automatically.

liangxin1300: No, without this last line, /etc/pacemaker/authkey wouldn't sync to the join-node.

krig: There is another problem if that is the case, I don't think this is the right solution.

liangxin1300:
Hi krig, I have adjusted the code, authkey will sync between nodes, no warning popped, please check.

BTW, I will back at 2/23:)

krig:
I will also be on vacation for the spring festival :) 新年快乐 !

krig:
Looks good, thank you!

Issue #350, opened 2018-02-06
method to check whether to enable sbd or not


liu4480:
if /etc/sysconfig/sbd exists and is configure, bootstrap will enable sbd anyway, even like:
SBD is already configured to use /dev/sdb - overwrite (y/n)? n

#tail /var/log/ha-cluster-bootstrap.log
++ systemctl enable hawk.service
##   Log in with username 'hacluster'
++ systemctl enable sbd.service
Created symlink /etc/systemd/system/corosync.service.requires/sbd.service \u2192 /usr/lib/systemd/system/sbd.service.
Created symlink /etc/systemd/system/pacemaker.service.requires/sbd.service \u2192 /usr/lib/systemd/system/sbd.service.
Created symlink /etc/systemd/system/dlm.service.requires/sbd.service \u2192 /usr/lib/systemd/system/sbd.service.
++ systemctl enable pacemaker.service
++ systemctl start pacemaker.service

my expectation we should base on  the answer is 'y', and also /etc/sysconfig/sbd is configured, then bootstrap can enable sbd.

liu4480:
just noticed, if /etc/sysconfig/sbd exists, it does not ask if you want to use sbd, but just thinking you are using sbd. In this condition, I think bootstrap firstly ask if you want to use sbd, then warns if you want to overwrite sbd configuration files. 

@krig do you think this could make it clearer?

liu4480:
thanks, close this issue since #354 was merged.

PR #354, opened 2018-02-13
low: bootstrap: Always ask whether to use sbd

liangxin1300:
Hi krig,

As bin mentioned in #350,
bootstrap should always ask whether to use sbd, even sbd is already configured before.


krig:
Perfect, thank you :)

PR #356, opened 2018-02-23
 low: "master-max" and "master-node-max" deprecated in 2.0

liangxin1300:


krig:
It would be better to add `promoted-max` etc. to the list and not remove `master-max`, since crmsh can still be used with pacemaker < 2.0.

dmuhamedagic:
Agreed!

Dejan
(the regression-avoidance-please-keep-compatibility freak)

PR #357, opened 2018-02-26
[crmsh 2] crmsh cluster init : support GCE

dud225:
Hello

I'm using crmsh to setup a cluster on GCP. Unfortunately GCE instances are configured in a pretty unsual way in that VM are all set a /32 IP, that is all LAN-destined trafic goes through the GCP gateway. So we may not configure Corosync to bind on the network address, we have to request it to use the local IP address.

I've noticed that these init scripts are now deprecated in the latest release, so I'd like to know if there is an interest in this code before spending time to port it to the newer init mechanism. If so, let me known what should be reworked or improved.

Regards 
dud

### DIFF
--- a/scripts/init/configure.py
+++ b/scripts/init/configure.py
@@ -50,6 +50,11 @@ def run_install():
 def make_bindnetaddr():
     host = crm_script.host()
     hostinfo = crm_script.output(2)[host]
+
+    # Bind to the local IP address
+    if crm_script.param('bindipaddr'):
+        return hostinfo['net']['hostname']['ip']
+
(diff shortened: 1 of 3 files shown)


krig:
Hi! Yes, the init scripts have been replaced by the bootstrap module in crmsh now, to gain more flexibility and enable interactive setup.

We're definitely interested in enabling bootstrap on GCE as well.

dud225:
Hello

you can look for the code for crmsh 3 in PR #369 

Regards
dud

krig:
Thank you!

PR #358, opened 2018-02-27
Update man-3.adoc

schwindp:
fixed typo in resource agent name


schwindp:
might need to be fixed in `man-1.2.adoc`, `man-2.0.adoc` as well.

krig:
Nice catch, thanks!

PR #366, opened 2018-04-24
low: terminal will lose cursor after type ctrl+c(bsc#1090626)

liangxin1300:
refer to #298,
this is a common way to fix this problem:)


krig:
Looks good to me, thanks!

PR #368, opened 2018-05-10
New constraints rules syntax

liangxin1300:
#### Motivation
- current syntax is not very clear
```
e.g.
  
location l1 g1 100: node1
  
location l3 m5 rule inf: #uname eq node1 and pingd gt 0

order o-2 Serialize: A ( B C )

collocation c2 inf: m5:Master d1:Started

rsc_ticket ticket-C_master ticket-C: m6 m5:Master loss-policy=fence
```
- no rules/resource_set/options completer, input manually is hard because xml.rng is complicate and huge
- no options help messages by interactive

#### Features:
- Explicit syntax
```
e.g.
location id=l1 g1 on node1 score=100

location l3 m5 \
        rule score=inf \
        expression attribute=#uname operation=eq value=node1 and \
        expression attribute=pingd operation=gt value=0

order o-2 \
        resource_set A \
        resource_set B C sequential=false \
        options kind=Serialize

colocation c2 m5 role=Master with d1 with_role=Started options score=inf

rsc_ticket ticket-C_master ticket=ticket-C \
        resource_set m6 \
        resource_set m5 role=Master \
        options loss-policy=fence
```
- complete each steps which from latest constraints.rng and rule.rng
- all options have help messages in interactive mode

#### Demo
- [location](https://asciinema.org/a/172520)
- [order](https://asciinema.org/a/172521)
- [colocation](https://asciinema.org/a/172522)

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -577,7 +577,7 @@ class CibObjectSetCli(CibObjectSet):
         comments = []
         for cli_text in lines2cli(s):
             err_buf.incr_lineno()
-            node = parse.parse(cli_text, comments=comments)
+            node, _ = parse.parse(cli_text, comments=comments)
             if node not in (False, None):
                 rc = rc and diff.add(node)
             elif node is False:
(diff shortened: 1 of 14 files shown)


krig:
Hi @liangxin1300 , first of all I am very impressed with the work you have done to implement the new syntax! It think it is a lot clearer and easier to understand than the earlier syntax.

I have one major concern with the change, and that is backwards compatibility. We need to have some way to parse the old syntax into the new syntax. For example, the Openstack HA support calls crmsh to create resources and constraints, and I don't want to break all of their code. :(

Maybe we even need a way for them to switch to the old syntax for output also...

krig:
Yes, that should work I think. Thank you :)

liangxin1300:
Hi @krig :
    I've added the compatibility for the old syntax(updated the commits in this PR),
    Leaving current test codes nearly unchanged and added some test cases for the new syntax.
    So, for the codes already used in some scripts, it's **totally** **transparent**:)

   But I still wandering how to change the document? Do we need **replace** the old syntax in it?
   Or, just append the new contents for the new sytanx?

   And,
   Maybe we need open a Fate to trace this? 
   And which OS version do we plan to put this feature into? 15SP1?
   I don't think it's necessary to put these into 12sp4

PR #369, opened 2018-05-11
[crmsh 3] crmsh cluster init : support GCE

dud225:
Hello

This PR permits the use of crmsh to configure cluster hosted on GCE.
GCE instances are peculiar because the network interface is configured with a /32 address.

Regards
dud


krig:
Thanks!

PR #371, opened 2018-05-21
High: hbreport: fix UnicodeEncodeError while print(bsc#1093564)

liangxin1300:


krig:
I really wonder if this is a good idea... If the terminal is not ready to receive UTF-8, I wonder if this really works as intended.

liangxin1300:
in https://github.com/ClusterLabs/crmsh/pull/362/commits/8b0eb08e2d2c6c094a69bf4428be95e0fb03b6ac
we meet the same error while write in a file

I have no other ideas how to fix it. Could you give me some suggestions?:)

PR #373, opened 2018-05-22
Low: Travis-CI: make sure exit value is not 0 while unittest failed

liangxin1300:
Hi krig:
    I create this request because I want the whole Travis-CI return failed status when unittest failed and testcases success


krig:
makes sense, thanks :)

PR #375, opened 2018-06-28
dev: Using ipaddress instead of ipcalc in Python3

liangxin1300:
ipaddress is from std library, remove ipcalc code, keep simple and stable


krig:
Very nice, thanks!

PR #376, opened 2018-07-25
Low: bootstrap: No warning message when using '-q'

liangxin1300:
Warning message should be shown whether quiet option has been set.


krig:
The purpose of `-q` is to not print to the command line unless really necessary. The warning is written to the log still?

liangxin1300:
e.g
   when using "-q" option, it just popped:
   "Do you want to continue anyway (y/n)",
   the warning message "No watchdog device found. If SBD is used, the cluster will be unable to start without a watchdog" will not show.

   It's a little strange

krig:
Aaah... that does look strange indeed... OK, merging.

PR #379, opened 2018-08-06
Fix: bootstrap: "-i" option doesn't work(bsc#1103833, bsc#1103834)

liangxin1300:
ha-cluster-init command has option "-i" to bind ip address to the specific interface,
I use -i option to change the default network address(bindnetaddr in corosync.conf), but it doesn't work


krig:
Oh, that's not good. :/ Thank you!

PR #381, opened 2018-08-15
dev: a conservative way for commit 715f27

### LINKED ISSUE #367: Detect -h for crm configure property
When using the command `crm configure property -h` it gives the following warning:

```
# crm configure property -h
WARNING: cib-bootstrap-options: unknown attribute '-h'
```

I would have expected that it prints a help message similar to `crm configure help property`. Or at least recognize that `-h` isn't a valid property name. :wink:

However, what's worse is, the command writes `-h` into the CIB:

```
# crm configure show
[...]
property cib-bootstrap-options: \
        have-watchdog=true \
        dc-version="1.1.18+20180313.2c83c6872-1.1-2c83c6872" \
        cluster-infrastructure=corosync \
        cluster-name=hacluster \
        stonith-enabled=true \
        placement-strategy=balanced \
        enable-acl=true \
        have-watchdog \
        no-quorum-policy \
        -h
[...]
```

As you can see, the `-h` appears in the CIB. I'm pretty sure, you don't want that. :wink:

liangxin1300:
I found commit #378, to resolve #367, have some problems:
1. commands in ui_cluster like init, join, add, remove, and geo_* already accept "-h" and "--help" option,
    #378 will overwrite them
2. var line_list maybe from root level like ["configure", "property", "-h"], its length not always be 2

I can not find a good way to solve this.
So I just focus on the "property" command temporary.

### DIFF
--- a/crmsh/ui_context.py
+++ b/crmsh/ui_context.py
@@ -49,8 +49,9 @@ class Context(object):
         '''
         def trans_to_help(line):
             line_list = line.split()
-            if len(line_list) == 2 and line_list[1] in ["-h", "--help"]:
-                return "help " + line_list[0]
+            if line_list[-1] in ["-h", "--help"] and \
+               line_list[-2] == "property":
+                return " ".join(line_list[:-2] + ["help", "property"])
             else:
                 return line
 


krig:
Alright.

PR #382, opened 2018-08-22
support ocfs2 log collecting

losemyheaven:
Hi Krig,

This pr intends to add support for collecting ocfs2 logs.

Sometimes, our customer need this to report log to us.

And I also need your suggestion.

Thanks.

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -283,6 +283,7 @@ def collect_info():
     time_status()
     corosync_blackbox()
     get_ratraces()
+    dump_ocfs2()
 
     for p in process_list:
         p.join()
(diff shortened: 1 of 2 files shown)


hb_report/utillib.py
```diff
@@ -1629,6 +1630,20 @@ def time_status():
     time_f = os.path.join(constants.WORKDIR, constants.TIME_F)
     crmutils.str2file(out_string, time_f)
 
+def dump_ocfs2():
+    ocfs2_f = os.path.join(constants.WORKDIR, constants.OCFS2_F)
+    f = open(ocfs2_f, "w")
```
krig: Two recommendations:

1. Make sure that you check and handle the case if opening the file fails for whatever reason.

2. Use the Python `with` syntax to close the file:

```
with open(ocfs2_f, "w") as f:
   ...
```

hb_report/utillib.py
```diff
@@ -1629,6 +1630,20 @@ def time_status():
...
+
+    #dump all tasks stack into dmesg
+    cmd = "echo t > /proc/sysrq-trigger"
+    crmutils.get_stdout(cmd)
```
krig: Hmm, this looks odd to me. Maybe use `os.system("echo t > /proc/sysrq-trigger")` instead?

losemyheaven: Thanks krig,
I'll do it.
: )

losemyheaven:
Thanks krig,
I'll do it.
: )

PR #383, opened 2018-08-27
hbreport: process name change for pacemaker 2.0(bsc#1106052)

liangxin1300:
Got "STOPPED" flag file in hbreport contents, 
while the cluster is running on this node,
This is because the processes name in pacemaker has been changed
(crmd to pacemaker-controld, but hbreport code still check crmd)


krig:
Please instead check for both process names, since older versions of pacemaker use the old name.

liangxin1300:
Changed:)

krig:
Thanks :)

PR #387, opened 2018-09-21
fix: bootstrap: fix non interactive unicast cluster init and join(bsc…

liangxin1300:
…#1109172)

* Add "-l" and "--iplist" option in for init and join process, to specific IP address when using unicast on non interactive mode
* Validate these IP addresses provide by this new option

example:
* init process could be:
   `crm cluster init -y -u -l 10.10.10.11`
  while join process should be:
   `crm cluster join -y -c peer_ip -l 10.10.10.12`

* init
 `crm cluster init -yuM -l 10.10.10.11,20.20.20.11`
 join
 `crm cluster join -yc peer_ip -l 10.10.10.12,20.20.20.12`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1713,12 +1713,11 @@ def join_cluster(seed_host):
     is_unicast = "nodelist" in open(corosync.conf()).read()
     if is_unicast:
         ringXaddr_res = []
-        print("")
         for i in 0, 1:
             while True:
                 ringXaddr = prompt_for_string('Address for ring{}'.format(i),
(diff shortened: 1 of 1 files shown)


krig:
Being able to pass in the IP list may be useful when configuring multiple rings and unicast, but it should not be necessary for unicast. The join process should be able to get the ring0 IP address based on the interface name alone, using `utils.network_defaults` just like init.

I would prefer a fix like that for this issue and adding the new parameter as a separate unrelated feature.

liangxin1300:
So, simple fix is coming:)
Provide default IP address when joining

krig:
Perfect, thank you :)

I still think adding the `-l` parameter is a good idea, please submit this as a separate pull request :)

PR #388, opened 2018-09-25
High: bootstrap: change default ip address way for both mcast and unicat

liangxin1300:
Problems solved:
* For mcast
  `crm cluster init -M -y` 
   will failed, cause the second default ip value is None

* For unicat
  `crm cluster join -c peer_ip -y`


liangxin1300:
Do we need fire a new bug for this if this commit make sense?

liangxin1300:
sorry, maybe need more test for this PR

PR #389, opened 2018-09-26
High: bootstrap: change default ip address way for both mcast and unicat

liangxin1300:
Problems solved:

* For mcast
`crm cluster init -M -y`
will failed, cause the second default ip value is None

* For unicat
`crm cluster join -c peer_ip -y`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -235,6 +235,24 @@ def wait_for_cluster():
     status_done()
 
 
+def pick_default_value(vlist, ilist):
+    # give a different value for second config items
+    default_value = ""
+
+    if not vlist:
+        return default_value
+
+    default_value = vlist[0]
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -235,6 +235,24 @@ def wait_for_cluster():
     status_done()
 
 
+def pick_default_value(vlist, ilist):
```
krig: Hm, this function is a bit strange to me.. only the first element of `ilist` is ever used, and `vlist` is modified.
I would prefer it if the function returned two return values: The chosen default value, and the remaining values from vlist.. and it would be clearer if the `ilist` parameter was a single element called `preferred_value` or something like that..

Also I try to document all functions with a documentation string, especially functions that modify their parameters :)

PR #391, opened 2018-09-27
Fix: bootstrap: change default ip address way for both mcast and unic…

liangxin1300:
…at(bsc#1109975,bsc#1109974)
Problems solved:
**in mcast mode:**
run command `crm cluster init -M -y` failed
error messages are:
"ERROR: cluster.init: No value for bindnetaddr"

**in unicast mode:**
run command `crm cluster init -M -y` failed
error messages are:
"ERROR: cluster.init: No value for ring1"

**These because we forgot provide the second default value for non interactive mode**

And for unicat
crm cluster join -c peer_ip -y

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -235,6 +235,21 @@ def wait_for_cluster():
     status_done()
 
 
+def pick_default_value(default_list, prev_list):
+    """
+    Provide default value for function 'prompt_for_string'.
+    Make sure give different default value in multi-ring mode.
+
+    Parameters:
+    * default_list - default value list for config item
+    * prev_list    - previous value for config item in multi-ring mode
+    """
+    for value in default_list:
+        if value not in prev_list:
+            return value
+    return ""
+
+
 def start_service(service):
     """
     Start and enable systemd service
@@ -886,11 +901,6 @@ def valid_port(port, prev_value=None):
 
 
 def init_corosync_unicast():
-    def pick_default_value(vlist, ilist):
-        # give a different value for second config items
-        if len(ilist) == 1:
-            vlist.remove(ilist[0])
-        return vlist[0]
 
     if _context.yes_to_all:
         status("Configuring corosync (unicast)")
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -235,6 +235,22 @@ def wait_for_cluster():
...
+    Parameters:
+    * default_list - default value list for config item
+    * prev_list    - previous value for config item in multi-ring mode
+    """
+    if not default_list:
+        return ""
+    for value in default_list:
+        if value not in prev_list:
+            return value
+
```
liangxin1300: hi krig, for this function, I've tried to make it more simple:)
just make sure to give different default value in multi-ring config

e.g.
```
ring0[10.10.10.1] 20.20.20.1
ring1[10.10.10.1]
```
for ring0, user not pick the default ip "10.10.10.1", input the original second default ip "20.20.20.1",
so for ring1, if we still give the default value as "20.20.20.1", this will not work, the default value must always not be already configured

krig: Right now this could return `None` if `value` is not in `default_list`: It would be better to `return ""` as the last line in the function just in case :)

crmsh/bootstrap.py
```diff
@@ -911,27 +921,19 @@ def pick_default_value(vlist, ilist):
...
-        else:
-            default_networks = [_context.ip_network]
-    if not default_networks:
+    local_iplist = utils.ip_in_local(_context.ipv6)
+    len_iplist = len(local_iplist)
+    if len_iplist == 0:
         error("No network configured at {}!".format(utils.this_node()))
 
+    local_iplist.remove(_context.ip_address)
+    default_ip = [_context.ip_address, "" if not local_iplist else local_iplist[0]]
```
krig: How about

```
default_ip = [_context.ip_address] + [ip for ip in local_iplist if ip != _context.ip_address]
```

Modifying local_iplist is confusing to me, there is a risk that it would be used later but now it does not contain `_context.ip_address`.

Same thing for line 1012 below.

crmsh/bootstrap.py
```diff
@@ -235,6 +235,21 @@ def wait_for_cluster():
...
+    Make sure give different default value in multi-ring mode.
+
+    Parameters:
+    * default_list - default value list for config item
+    * prev_list    - previous value for config item in multi-ring mode
+    """
+    for value in default_list:
+        if value not in prev_list:
+            return value
+    return ""
```
liangxin1300: like this?

PR #400, opened 2018-10-22
auto-commit enabling/disabling maintenance mode for a whole cluster

cupracer:
This fix ensures that enable/disable actions via the command "crm maintenance" are committed automatically.


krig:
Thank you!

PR #401, opened 2018-10-24
ui_cluster: restart cluster is added

aleksei-burlakov:


ChrisKowalczyk:
Good change, Aleksei, we have been missing this feature for some time..!

krig:
Looks good, thanks for updating the documentation as well :)

gao-yan:
Good thinking. Not sure if it depends on how ExecStop is defined in the service file, it's not clear if it's really by default synchronous or asynchronous,  given that there are "systemctl --no-block/--wait" options.

Of course another solution is to invoke "systemctl restart corosync".

aleksei-burlakov:
I checked that `crm cluster stop` really wait's until the cluster has stopped. Besides, if an exception was thrown, then crm won't try to start the cluster.

PR #404, opened 2018-11-14
dev: Suggestion: Try to centralize command options in one file

liangxin1300:
hi krig, 

How about making change like this?
I create this PR because I want to get options value and help info when completing, before this sub-command run("cluster init" command here).

Btw1, maybe putting all sub-command's options in one file is easy to maintain? I don't know exactly.

Btw2, in crmsh, we are using three option parse function: argparse, optparse and getopt(in hbreport),
    i think it's better to unify them into one, like argparse? cause optparse is deprecated.

### DIFF
--- a/crmsh/options.py
+++ b/crmsh/options.py
@@ -16,3 +16,77 @@ shadow = ""
 scriptdir = ""
 # set to true when completing non-interactively
 shell_completion = False
+
+
+arg_parameter_key = ["action", "dest", "metavar", "default", "type", "help"]
+
+cluster_init_base_options_value = [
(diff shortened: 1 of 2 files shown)


krig:
Hi Xin,

Yes, agreed. We should try to move everything to argparse since this is the maintained option for Python.

liangxin1300:
Hi @krig @diegoakechi , please consider to revert this commit.
At the time when I want to create this commit, I'm investigating [python prompt toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit), the benefit to do this is for completion.

But to be honest, the biggest defect for this commit is uneasy to maintain each option because they are at different files:(

PR #406, opened 2019-01-03
Fix: bsc#1120554, bsc#1120555 crmsh crashed when using template>apply

liangxin1300:

PR #408, opened 2019-01-07
High: Testcases: update the testcases/bugs for bsc#1120554

liangxin1300:
Hi, krig:
   For pr #406 , I have updated the testcases in testcases/bug


krig:
Great, thanks! Maybe you could add this commit to that branch, before the commit with the fix? And then add a commit following the fix with the expected test output on success?

PR #409, opened 2019-01-13
low: utils: convert string contstants to bytes

vvidic:
Make sure return type for stdout and stderr is always bytes
or stdout/stderr.decode('utf-8') fails.


krig:
Thank you :)

PR #410, opened 2019-01-13
dev: log_patterns: fix regexp warning for [[]

vvidic:
Python 3.7 gives 'FutureWarning: Possible nested set' so
replace any regexp [[] with \[.


krig:
...and thank you again. That’s a readability improvement as well.

PR #414, opened 2019-01-30
brace expansion with single quotation marks in scp command is not safe

liangxin1300:
for bsc#1123028, affter updating openssh, this bug can be reproduced in my own environment
this is most likely caused by bsc#1122695,


krig:
Why did you close this PR? Was the fix not working?

liangxin1300:
I closed this because in https://bugzilla.suse.com/show_bug.cgi?id=1123028, people updated openssh again and our code works after updated openssh

PR #422, opened 2019-02-27
reject append ipaddress to config file if already have(bsc#1127095, 1127096)

liangxin1300:
I have two nodes, node1 using 'crm cluster init -u' and node2 using 'crm cluster join', it works;

But then I run 'crm cluster stop; crm cluster join' on node2, corosync.conf will become 3 nodes cluster config file, cause node2's ip address appended it again

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1745,7 +1745,10 @@ def join_cluster(seed_host):
                 break
         print("")
         invoke("rm -f /var/lib/heartbeat/crm/* /var/lib/pacemaker/cib/*")
-        corosync.add_node_ucast(ringXaddr_res)
+        try:
+            corosync.add_node_ucast(ringXaddr_res)
+        except ValueError as e:
+            warn(e)
         csync2_update(corosync.conf())
         invoke("ssh -o StrictHostKeyChecking=no root@{} corosync-cfgtool -R".format(seed_host))
 
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -386,6 +386,15 @@ def add_node_ucast(IParray, node_name=None):
     f = open(conf()).read()
     p = Parser(f)
 
+    # to check if the same IP already configured
+    exist_iplist = []
+    for path in p.all_paths():
+        if re.search('nodelist.node.ring[0-9]*_addr', path):
+            exist_iplist.extend(p.get_all(path))
+    for ip in IParray:
+        if ip in exist_iplist:
+            raise ValueError("IP {} was already configured".format(ip))
+
     node_id = get_free_nodeid(p)
     node_value = []
     for i, addr in enumerate(IParray):
--- a/test/unittests/test_corosync.py
+++ b/test/unittests/test_corosync.py
@@ -92,6 +92,27 @@ class TestCorosyncParser(unittest.TestCase):
         _valid(p)
         self.assertEqual(p.count('nodelist.node'), nid - 1)
 
+    def test_add_node_ucast(self):
+        from crmsh.corosync import add_node_ucast, get_values
+
+        os.environ["COROSYNC_MAIN_CONFIG_FILE"] = os.path.join(os.path.dirname(__file__), 'corosync.conf.2')
+
+        exist_iplist = get_values('nodelist.node.ring0_addr')
+        try:
+            add_node_ucast(['10.10.10.11'])
+        except ValueError:
+            self.fail("corosync.add_node_ucast raised ValueError unexpectedly!")
+        now_iplist = get_values('nodelist.node.ring0_addr')
+        self.assertEqual(len(exist_iplist) + 1, len(now_iplist))
+        self.assertTrue('10.10.10.11' in get_values('nodelist.node.ring0_addr'))
+
+        # bsc#1127095, 1127096; address 10.10.10.11 already exist
+        with self.assertRaises(ValueError) as err:
+            add_node_ucast(['10.10.10.11'])
+        self.assertEqual("IP 10.10.10.11 was already configured", str(err.exception))
+        now_iplist = get_values('nodelist.node.ring0_addr')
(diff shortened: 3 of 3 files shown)


krig:
I think a better solution would be to detect and avoid duplicates in `corosync.add_node_ucast`, and to use the functionality in the `corosync` module to check the node address property only, instead of grepping the whole config file for `:<ip>`.

arbulu89:
@liangxin1300 Agree with @krig . It has more sense to add this check to corosync module.
And create a new UT for this new use case. Let us know if you need any help on that

crmsh/bootstrap.py
```diff
@@ -1745,9 +1745,15 @@ def update_nodeid(nodeid, node=None):
                 break
         print("")
         invoke("rm -f /var/lib/heartbeat/crm/* /var/lib/pacemaker/cib/*")
-        corosync.add_node_ucast(ringXaddr_res)
-        csync2_update(corosync.conf())
-        invoke("ssh -o StrictHostKeyChecking=no root@{} corosync-cfgtool -R".format(seed_host))
+        already_have = False
+        for ip in ringXaddr_res:
```
arbulu89: Python has a cleaner way to manage this kind of situation where you need to control a boolean in a loop.
It would be something like that:
```
for i in my_data:
    if my_check ==true:
        break
else:
    # if break didn't happen
    code_in_case_the_last_condition_didnt_match
```
More info here: http://book.pythontips.com/en/latest/for_-_else.html

crmsh/corosync.py
```diff
@@ -386,6 +386,11 @@ def add_node_ucast(IParray, node_name=None):
     f = open(conf()).read()
     p = Parser(f)
 
+    exist_iplist = p.get_all('nodelist.node.ring0_addr')
+    for ip in IParray:
+        if ip in exist_iplist:
+            return
+
```
krig: Ah, this doesn't look quite right: The IParray contains [ring0_addr, ring1_addr, ..., ringN_addr], right?
Really, this should get all configured IP addresses, and if any of the input IP addresses is already in use, raise an exception.

So I'd write a function `get_all_configured_addresses` which returns a list of all configured addresses, and then check if any of the input addresses is in that list. If so, `raise ValueError("IP {} was already configured")` - and then make sure to catch and handle that.

liangxin1300: Thank you!

crmsh/bootstrap.py
liangxin1300: Thank you for suggestion!

crmsh/corosync.py
liangxin1300: I have changed the logic to make sure extract all configured IP.

And I think if this case does happen, we should better keep silence and complete the remain bootstrap process to make sure join node will start cluster successfully, 
So just return, no raise error

krig: What you can do is raise the error here (since that is the correct response in the general case), and catch the error at the usage site to ignore it there.

liangxin1300: OK, changed

liangxin1300:
Hi @krig , please consider to merge this, thanks!

liangxin1300:
Hi @krig @diegoakechi , I recommend to merge this cause I will trigger this bug while working on #464
Thanks!

liangxin1300:
hi @diegoakechi , please consider to merge this
Thank you!

diegoakechi:
@liangxin1300 I have approved, but as @krig and @arbulu89 also have requested changes, they need to approve it. Please, can you guys take a look?

test/unittests/test_corosync.py
```diff
@@ -92,6 +92,26 @@ def test_add_node_no_nodelist(self):
...
+        os.environ["COROSYNC_MAIN_CONFIG_FILE"] = os.path.join(os.path.dirname(__file__), 'corosync.conf.2')
+
+        exist_iplist = get_values('nodelist.node.ring0_addr')
+        try:
+            add_node_ucast(['10.10.10.11'])
+        except ValueError:
+            self.fail("corosync.add_node_ucast raised ValueError unexpectedly!")
+        now_iplist = get_values('nodelist.node.ring0_addr')
+        self.assertEqual(len(exist_iplist) + 1, len(now_iplist))
+
```
arbulu89: Shouldn't we check if the new ip address is in the list? The exact value

crmsh/corosync.py
```diff
@@ -386,6 +386,15 @@ def add_node_ucast(IParray, node_name=None):
     f = open(conf()).read()
     p = Parser(f)
 
+    # to check if the same IP already configured
+    exist_iplist = []
+    for path in p.all_paths():
+        if re.search('nodelist.node.ring[0-9]*_addr', path):
+            exist_iplist += p.get_all(path)
+    for ip in IParray:
+        if ip in set(exist_iplist):
```
arbulu89: I don't really think that we need the `set` method here, but I guess is not harmfull

crmsh/corosync.py
```diff
@@ -386,6 +386,15 @@ def add_node_ucast(IParray, node_name=None):
     f = open(conf()).read()
     p = Parser(f)
 
+    # to check if the same IP already configured
+    exist_iplist = []
+    for path in p.all_paths():
+        if re.search('nodelist.node.ring[0-9]*_addr', path):
+            exist_iplist += p.get_all(path)
```
arbulu89: I would use `append` instead of `+=`, it's a bit more visual

liangxin1300: get_all will return a list
and here I want to extend the list `exist_iplist`, not append

crmsh/corosync.py
liangxin1300: Thanks!
Changed

crmsh/corosync.py
krig: You could use `exist_iplist.extend(p.get_all(path))`, but I am fine with `+=`

liangxin1300:
@diegoakechi Please merge this, thanks!

PR #429, opened 2019-03-19
low: msg: add timestamp for DEBUG messages(bsc#1129380)

liangxin1300:
This is from yan's requirement

> Many times when we asked users to provide the output of "crm -d ..." for the purpose of debugging, we'd almost always ask them to also run "date" and tell us the timestamp so that we could later dig into the corresponding specific context of the logs  meanwhile for example from hb_report, rather than getting lost in the dozens of thousands lines of them.
It'd be very convenient if "crm -d" could include a timestamp itself for example at the beginning of the output.

### DIFF
--- a/crmsh/main.py
+++ b/crmsh/main.py
@@ -360,6 +360,8 @@ def run():
             err_buf.reset_lineno()
             options.batch = True
         user_args = parse_options()
+        if config.core.debug:
+            print(utils.debug_timestamp())
         term.init()
         if options.profile:
             return profile_run(context, user_args)
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2159,4 +2159,8 @@ def iplist_for_cloud():
     return []
 
 
+def debug_timestamp():
+    return datetime.datetime.now().strftime('%Y/%m/%d %H:%M:%S')
+
+
 # vim:ts=4:sw=4:et:


krig:
Looking at what is asked for, there is no need to add a timestamp to each debug message. Just printing the date once at startup (when running with the -d flag) would be enough.

liangxin1300:
OK, changed:)

crmsh/main.py
```diff
@@ -303,6 +303,8 @@ def compgen():
 
 def parse_options():
     opts, args = option_parser.parse_args()
+    if opts.debug:
+        print(utils.debug_timestamp())
```
diegoakechi: @liangxin1300 I am not familiar with crmsh coding, but sounds weird to me to have this print into a function that has the main purpose of parsing the command line options.

krig: yes, this is not the best place for printing the timestamp. Please move the timestamp printing out of the options parser to a more appropriate location :)

crmsh/main.py
```diff
@@ -360,6 +360,8 @@ def run():
             err_buf.reset_lineno()
             options.batch = True
         user_args = parse_options()
+        if config.core.debug:
+            print(utils.debug_timestamp())
```
liangxin1300: How about here?:)

diegoakechi: Sounds reasonable to me...

PR #430, opened 2019-03-19
Fix: bsc#1129719: check command and related files exist

liangxin1300:
to solve this bug: https://bugzilla.suse.com/show_bug.cgi?id=1129719


krig:
For context, pasting the contents of the bug here:

```
when running "hb_report", there are some error message, like
"/bin/sh: mounted.ocfs2: command not found
cat: /sys/fs/ocfs2/cluster_stack: No such file or directory" 

maybe need to check if command and file exists.
```

PR #431, opened 2019-03-22
 low: hb_report: collect output of "sbd dump" and "sbd list"(bsc#1129383)

liangxin1300:
- Output of "sbd dump"
- Output of "sbd list"

On diskless sbd mode, no output in sbd.txt


krig:
Right now we don't have tests for hb_report at all, since it needs a real cluster to work. It's not great, but that's how it is...

liangxin1300:
Maybe recently I can write some unit test codes to test specific functions which do not need cluster running.
And make some improvements.

PR #434, opened 2019-03-27
medium: parse: op params detect and error ... (bsc#1129210)

krig:
Detect illegal ordering of attributes and error out early instead of triggering a more obscure validation error when trying to commit.

Unfortunately there is no way to allow mixing of instance_attributes
with op node attributes, as reconstructing the line-based syntax
from the XML is impossible.

This fix detects when the ordering is illegal and errors out.


krig:
@angelabriel Thank you!

PR #436, opened 2019-03-28
high: cibconfig: Use correct CIB as patch base (bsc#1127716)

krig:
crmsh was passing a sanitized CIB version as the basis for
patches. This works as long as the CIB is already sanitized
(apparently almost always the case) but generates strange
diffs otherwise.

### DIFF
--- a/crmsh/xmlutil.py
+++ b/crmsh/xmlutil.py
@@ -87,9 +87,17 @@ def cibdump2file(fname):
     return None
 
 
-def cibdump2tmp():
+def cibdump2tmp(filterfn=None):
     try:
         _, outp, _ = sudocall(cib_dump)
+        if filterfn is not None:
+            try:
+                cib_elem = etree.fromstring(outp)
+            except etree.ParseError as msg:
+                common_err(msg)
+                return None
+            filterfn(cib_elem)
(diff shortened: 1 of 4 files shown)


crmsh/xmlutil.py
```diff
@@ -675,6 +684,22 @@ def sanitize_cib(doc):
...
+    """
+    Custom version of sanitize_cib which
+    doesn't sort container children, to use
+    for processing the original CIB when
+    generating a patch to apply using crm_diff.
+    """
+    xml_processnodes(doc, is_status_node, rmnodes)
+    # xml_processnodes(doc, true, printid)
+    # xml_processnodes(doc, is_emptynvpairs, rmnodes)
+    # xml_processnodes(doc, is_emptyops, rmnodes)
```
diegoakechi: I would maybe remove the commented lines to make it easier to read and visualize the difference between the two functions.

krig: alright, makes sense, I'll do that.

diegoakechi:
Wow! Nice and very complex new test case...  Well done @krig

liangxin1300:
Great Job!

PR #437, opened 2019-04-09
hb_report unittest

liangxin1300:
Changes include:
  * unittest: add test_report.py to test hb_report module
      - cover most functions hb_report related
      - use special characters(see test/unittests/pacemaker_unicode.log) and invalid utf-8 in test case
  * setting error='replace' to handle UnicodeDecodeError caused by invalid utf-8 characters(bsc#1130715)
  * try to catch UnicodeDecodeError and print traceback
  * Using Tempfile class to manage tempfiles

### DIFF
--- a/hb_report/constants.py
+++ b/hb_report/constants.py
@@ -49,7 +49,6 @@ SSH_USER = ""
 SUDO = ""
 THIS_IS_NODE = 0
 TMP = None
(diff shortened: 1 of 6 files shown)


liangxin1300:
Well, @krig , I think we should consider merge this for closing bug bsc#1130715

PR #443, opened 2019-04-25
Fix: hb_report: bsc#1121912 hb_report stuck on named pipe in /var/log/

liangxin1300:


krig:
Hi Xin, perhaps you could change the commit message to follow the usual format. Something like:

```
medium: hb_report: Report gets stuck on named pipe in /var/log/ (bsc#1121912)

hb_report tried to read all files in /var/log/, and if there was a named pipe
in that folder it would get stuck. Fix by making sure to only read proper
files.
```

liangxin1300:
OK, changed!

krig:
Thank you!

PR #445, opened 2019-05-06
medium: ui_node: Check corosync state before clearstate (bsc#1129702)

krig:
Sometimes nodes are in "pending" state which is not really "offline".
But "stonith_admin --confirm" will tell pacemaker the node is down...
So basically we need narrow down the cases that we run
"stonith_admin --confirm" for.

### DIFF
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -259,9 +259,11 @@ class NodeMgmt(command.UI):
             cib_elem = xmlutil.cibdump2elem()
             if cib_elem is None:
                 return False
-            node_state = cib_elem.xpath("//node_state[@uname=\"%s\"]/@crmd" % node)
-            if node_state == ['online']:
+            if cib_elem.xpath("//node_state[@uname=\"%s\"]/@crmd" % node) == ["online"]:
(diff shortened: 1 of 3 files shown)


liangxin1300:
Looks good to me:)

krig:
@arbulu89 XPATH is like a query language for XML, although lxml only supports a subset of it. But some of the syntax is `//foo` to find all `<foo>` tags anywhere in the tree, `[@bar="value"]` to get all the tags where the bar attribute has the value, and `@bar` to return the value of the bar attribute. So the output of the xpath method is a list of matches, where the match is the value of that attribute.

arbulu89:
> @arbulu89 XPATH is like a query language for XML, although lxml only supports a subset of it. But some of the syntax is `//foo` to find all `<foo>` tags anywhere in the tree, `[@bar="value"]` to get all the tags where the bar attribute has the value, and `@bar` to return the value of the bar attribute. So the output of the xpath method is a list of matches, where the match is the value of that attribute.

Thanks! I was confused as I don't see any `node_state` entry nor `crmd` attribute in the new test case, that's it

krig:
@arbulu89 ah, the test case is not very obvious, since it only specifies the input command and the expected output, not the actual contents of the CIB

PR #447, opened 2019-05-15
Remove landscape.io

krig:
Since landscape.io doesn't seem to work very well anymore, we should see if we can use some other linting service.


liangxin1300:
Looks good to me:)

PR #428, opened 2019-03-19
Dev: log_patterns: update patterns for pacemaker version 2.0+

liangxin1300:
## Problem
`crm history peinputs` and `crm history peinputs v` are failed to give outputs
```
15sp4-1:~ # crm history peinputs
INFO: Retrieving information from cluster nodes, please wait...
15sp4-1:~ # crm history peinputs v
INFO: Fetching updated logs from cluster nodes. Please wait...
Date       Start    End       Filename      Client     User       Origin      Tags
====       =====    ===       ========      ======     ====       ======      ====
```
And the debug info are meaningless like `DEBUG: +Event events: error`
```
15sp4-1:~ # crm -d history peinputs
2022/11/25 15:57:10
DEBUG: found pacemaker version: 2.1.2+20211124.ada5c3b36-150400.4.3.1
DEBUG: setting source to live
DEBUG: setting report times: <2022-11-25 06:57:10> - <None>
DEBUG: Found description.txt, created 1669362779.811167
DEBUG: history: setting year to 2022 (ts: 1669362779.811167)
DEBUG: Transition metadata loaded from /var/cache/crm/history-root/live/__meta.json
INFO: Fetching updated logs from cluster nodes. Please wait...
DEBUG: Candidate logs: ['/var/cache/crm/history-root/live/15sp4-1/ha-log.txt', '/var/cache/crm/history-root/live/15sp4-2/ha-log.txt']
DEBUG: Updating 15sp4-1 : /var/cache/crm/history-root/live/15sp4-1/ha-log.txt : /var/log/messages : 1663586
DEBUG: Updating 15sp4-2 : /var/cache/crm/history-root/live/15sp4-2/ha-log.txt : /var/log/messages : 1636294
DEBUG: updating /var/log/messages from 15sp4-1 (pos 1663586)
DEBUG: updating /var/log/messages from 15sp4-2 (pos 1636294)
DEBUG: found pacemaker version: 2.1.2+20211124.ada5c3b36-150400.4.3.1-2.1.2+20211124.ada5c3b36 in cib: /var/cache/crm/history-root/live/15sp4-2/cib.xml
DEBUG: parsing /var/cache/crm/history-root/live/15sp4-1/ha-log.txt
DEBUG: +Event node: 15sp4-1
DEBUG: +Event node: 15sp4-1
DEBUG: +Event node: 15sp4-1
DEBUG: +Event node: 15sp4-1
DEBUG: +Event events: error
DEBUG: +Event events: error
[...]

PR #449, opened 2019-05-21
Fix: hb_report: analysis.txt should include important messages

liangxin1300:
>Comparing with the hb_report's result in 12sp3, analysis.txt in hb_report should include warning, error, critical messages if they do exist in log files

Because missing suitable reg-patterns and wrong parameters order while finding log files, these messages lost in analysis.txt


liangxin1300:
@krig , this is a blocker for #428 , we should consider resolve this firstly :)

PR #450, opened 2019-05-21
Fix: bindnetaddr should accept both network and specific IP

liangxin1300:
> When running 'crm cluster init' to try to configure cluster with mcast,
in  step "Network address to bind to", that is "bindnetaddr" in corosync.conf, when inputting specific IP address which is a real address in local node, crmsh considers it's invalid.
While according to 'man corosync.conf', this way is totally valid.


krig:
Good catch, and the patch looks good!

PR #452, opened 2019-06-11
medium: bootstrap: Pick first match for multiple routes (bsc#1106946)

liangxin1300:
also for bsc#1137783, on 12sp3, so we need back port


krig:
Looks good

PR #461, opened 2019-07-26
High: cibconfig: Correctly sanitize the original CIB as patch base (bsc#1127716, bsc#1138405)

gao-yan:
Rather than retrieving the latest CIB again to pass it to "crm_diff -o",
we should stick with what "self.cib_elem" is originally based on no
matter if there's any new changes from the live CIB in the meantime.
Otherwise there's race condition that the generated CIB diff reverts the
changes made right before.

self.cib_orig should be sanitized with sanitize_cib_for_patching() based
on the original self.cib_elem that hasn't yet been sanitized.


gao-yan:
@liangxin1300 Please review and test. Thanks.
@krig Could you please double-check? Thanks.

liangxin1300:
@gao-yan Great work! Looks good to me.
And I also reproduce these two bugs in my own environment, after applying above commits, it both work:)
Thanks!

gao-yan:
Thanks, @liangxin1300. I'm back-porting it for crmsh-3.0 branch.

PR #464, opened 2019-08-13
Feature: Improve qdevice/qnetd management

liangxin1300:
## Motivation
As the major HA configuration tool set, crmsh must support qdevice/qnetd management.

At currently master branch, crmsh already have some options to configure qdevice when doing ha-cluster-init process, but need to improve for validation, qnetd management,  making qdevice process as a bootstrap stage, non-TLS process and TLS certification process.

Note: `stage` here means run qdevice process separately, no need to re-run all bootstrap process

#### 1. Setup qdevice/qnetd during cluster init process
When deploying cluster service on first node `hanode1`, using
`crm cluster init --qnetd-hostname=qnetd-node`
(`qnetd-node` is the node used for qnetd service) can get a node running both cluster service and `corosync-qdevice` service
[related demo](https://asciinema.org/a/273458)
#### 2. Setup qdevice during cluster join process
Assuming first node `hanode1` finished init process with qdevice service running, on second node `hanode2`, running
`crm cluster join -c hanode1`
will get `corosync-qdevice` service running on `hanode2`
[related demo](https://asciinema.org/a/273461)
#### 3. Setup qdevice/qnetd on an running cluster
This also means configuring qdevice/qnetd could be a `stage` of bootstrap;
When cluster running on `hanode1` and `hanode2`, running
`crm cluster init qdevice --qnetd-hostname=qnetd-node`
Then `corosync-qdevice` service running on these two nodes.
[related demo](https://asciinema.org/a/273463)
#### 4. Remove qdevice from a cluster
running
`crm cluster remove --qdevice`
can remove qdevice configuration and stop corosync-qdevice service from all cluster nodes
[related demo](https://asciinema.org/a/273464)
#### 5. Show status
using
`crm corosync status [ring|quorum|qnetd]`
```
15sp1-1:~ # crm corosync status quorum 
Quorum information
------------------
Date:             Fri Oct 11 10:11:02 2019
Quorum provider:  corosync_votequorum
Nodes:            2
Node ID:          168430131
Ring ID:          168430131/12
Quorate:          Yes

Votequorum information
----------------------
Expected votes:   3
Highest expected: 3
Total votes:      3
Quorum:           2  
Flags:            Quorate Qdevice 

Membership information
----------------------
    Nodeid      Votes    Qdevice Name
 168430131          1    A,V,NMW 10.10.10.51 (local)
 168430132          1    A,V,NMW 10.10.10.52
         0          1            Qdevice
```
```
15sp1-1:~ # crm corosync status qnetd
[...]

PR #465, opened 2019-08-14
Fix: ui_cluster: Improve crm cluster run command(bsc#1145520)

liangxin1300:
**Assumption:** Two nodes cluster(15sp1-1 and 15sp1-2)

---
**Problem:** For a running two nodes cluster, run `crm cluster run "ls"`, will give this error:
```
ERROR: [15sp1-1]: Exited with error code 255, Error output: The authenticity of host '15sp1-1 (10.10.10.51)' can't be established.
ECDSA key fingerprint is SHA256:gOm0UDlChIQvNXXWn0N+bzVa+yj5Y80krR/i0boe/Z8.
Are you sure you want to continue connecting (yes/no)? 
Host key verification failed.
```
**Suggested Solution:** Add `'StrictHostKeyChecking=no'` ssh_options for parallax
**Related commit:** 2f84449238f0212958813bf5ab5584f75ff47e3b

---
**Problem:** When cluster service is totally stopped, run `crm cluster run "ls" 15sp1-1`, crmsh crashed:
```
ERROR: running cibadmin -Ql: Connection to the CIB manager failed: Transport endpoint is not connected
Init failed, could not perform requested operations
  File "/usr/sbin/crm", line 46, in <module>
    rc = main.run()
  File "/usr/lib/python3.6/site-packages/crmsh/main.py", line 369, in run
    return main_input_loop(context, user_args)
  File "/usr/lib/python3.6/site-packages/crmsh/main.py", line 249, in main_input_loop
    rc = handle_noninteractive_use(context, user_args)
  File "/usr/lib/python3.6/site-packages/crmsh/main.py", line 205, in handle_noninteractive_use
    if context.run(' '.join(l)):
  File "/usr/lib/python3.6/site-packages/crmsh/ui_context.py", line 86, in run
    rv = self.execute_command() is not False
  File "/usr/lib/python3.6/site-packages/crmsh/ui_context.py", line 276, in execute_command
    rv = self.command_info.function(*arglist)
  File "/usr/lib/python3.6/site-packages/crmsh/ui_cluster.py", line 610, in do_run
    if node and node in hosts:
TypeError: argument of type 'NoneType' is not iterable
```
**Reason:** That is because function `list_cluster_nodes` trying to use `cibadmin` to get the node list from cib.xml, and return `None` when failed, while we miss handling `None` result here

**Suggested Solution:**
* Changing the implementation of function `list_cluster_nodes`:
   - when pacemaker running, using `crm_node` to fetch node list
   - when corosync running, using `corosync-cmapctl` to fetch the member IP address
   - static situation, get env  CIB_file or parse file `/var/lib/pacemaker/cib/cib.xml` to get the node list
* Handle the `None` situation, give error message like `failed to get node list from cluster`

**Related commit:** 0dfa747bd9db3235990f5a6510dfd21ca55597c0

---

**Problem:** Can not run command for non-cluster node
**Suggested Solution:** If specific node in arguments, no matter whether it's a cluster node, enable running command on this node(s); Should also test whether need password, if yes, set `askpass` to `True` for parallax options
**Related commit:** f3cd676438f16b1ae2c96bdcb10194da439332df

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2189,4 +2189,13 @@ def get_member_iplist():
             ip_list.append(match.group(1))
     return ip_list
 
+
+def check_ssh_passwd_need(hosts):
+    ssh_options = "-o StrictHostKeyChecking=no -o EscapeChar=none -o ConnectTimeout=15"
+    for host in hosts:
+        ssh_cmd = "ssh {} -T -o Batchmode=yes {} true".format(ssh_options, host)
+        rc, _, _ = get_stdout_stderr(ssh_cmd)
+        if rc != 0:
+            return True
+    return False
 # vim:ts=4:sw=4:et:
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1760,13 +1760,23 @@ def list_cluster_nodes():
         return None
 
     try:
+        # when pacemaker running
         rc, outp = stdout2list(['crm_node', '-l'], stderr_on=False, shell=False)
         if rc == 0:
             return [x for x in [getname(line.split()) for line in outp] if x and x != '(null)']
 
+        # when corosync running
+        ip_list = get_member_iplist()
+        if ip_list:
+            return ip_list
+
+        # static situation
+        cib_path = os.getenv('CIB_file', '/var/lib/pacemaker/cib/cib.xml')
+        if not os.path.isfile(cib_path):
+            return None
         from . import xmlutil
         node_list = []
-        cib = xmlutil.cibdump2elem()
+        cib = xmlutil.file2cib_elem(cib_path)
         if cib is None:
             return None
         for node in cib.xpath('/cib/configuration/nodes/node'):
@@ -2165,5 +2175,18 @@ def iplist_for_cloud():
 def debug_timestamp():
     return datetime.datetime.now().strftime('%Y/%m/%d %H:%M:%S')
 
+
+def get_member_iplist():
+    rc, out, err= get_stdout_stderr("corosync-cmapctl -b runtime.totem.pg.mrp.srp.members")
+    if rc != 0:
+        common_debug(err)
+        return None
+
+    ip_list = []
+    for line in out.split('\n'):
+        match = re.search(r'ip\((.*?)\)', line)
+        if match:
+            ip_list.append(match.group(1))
+    return ip_list
 
 # vim:ts=4:sw=4:et:
--- a/test/unittests/test_utils.py
+++ b/test/unittests/test_utils.py
@@ -12,6 +12,70 @@ from crmsh import config
 from crmsh import tmpfiles
 
 
+def test_check_ssh_passwd_need_True():
+    with mock.patch('crmsh.utils.get_stdout_stderr') as mock_get_stdout_stderr:
+        mock_get_stdout_stderr.side_effect = [(0, None, None), (1, None, None)]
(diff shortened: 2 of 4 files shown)


krig:
Maybe you could split the change into two commits, first the update to `list_cluster_nodes` and another for changing the commands to take multiple nodes? This would make the commit messages clearer too :)

liangxin1300:
Hi @krig @diegoakechi , please review this
And I also created commit e86b8548113a0d4a3385819d5076709be037ebfc to update doc

crmsh/utils.py
```diff
@@ -2166,4 +2176,26 @@ def debug_timestamp():
...
+
+    ip_list = []
+    for line in out.split('\n'):
+        match = re.search(r'ip\((.*?)\)', line)
+        if match:
+            ip_list.append(match.group(1))
+    return ip_list
+
+
+def test_ssh_need_passwd(hosts):
```
diegoakechi: @liangxin1300 I would avoid to use `test` prefix on a real function as it can cause confusion with unit tests. I would suggest here `check_ssh_passwd_need`.

liangxin1300:
> @liangxin1300 I requested a small change.
> 
> Also, please, avoid to use only the Bug number on the commit title. Please, use more meaning description of what the real change is, as not all the bug information is public available.
> 
> I would like also to see some unit tests for these new scenarios. Can you, please, include it?
> 
> Thanks.

@diegoakechi Thanks for your suggestions, I've changed the function name `test_ssh_need_passwd` as `check_ssh_passwd_need`, changed the commit messages, and append unittest code
Please review
Thanks!:)

test/unittests/test_utils.py
```diff
@@ -12,6 +12,58 @@
 from crmsh import tmpfiles
 
 
+def test_check_ssh_passwd_need():
+    with mock.patch('crmsh.utils.get_stdout_stderr') as mock_get_stdout_stderr:
+        mock_get_stdout_stderr.side_effect = [(0, None, None), (1, None, None)]
+        assert utils.check_ssh_passwd_need(["node1", "node2"]) == True
```
diegoakechi: @liangxin1300 we should also assert a False return case too.

test/unittests/test_utils.py
```diff
@@ -12,6 +12,58 @@
...
+
+
+def test_get_member_iplist_None():
+    with mock.patch('crmsh.utils.get_stdout_stderr') as mock_get_stdout_stderr:
+        mock_get_stdout_stderr.return_value = (1, None, "Failed to initialize the cmap API. Error CS_ERR_LIBRARY")
+        assert utils.get_member_iplist() is None
+    mock_get_stdout_stderr.assert_called_once_with('corosync-cmapctl -b runtime.totem.pg.mrp.srp.members')
+
+
+def test_get_member_iplist():
```
diegoakechi: @liangxin1300 We should have a test case for the case when the command failst too.

test/unittests/test_utils.py
liangxin1300: Yes, I defined this `def test_check_ssh_passwd_need_Flase():`

test/unittests/test_utils.py
liangxin1300: @diegoakechi we have this function to test the command failed: `def test_get_member_iplist_None`

liangxin1300:
@diegoakechi , please consider to merge this, cause it became a blocker for me when working on #464 
Thanks!

PR #467, opened 2019-08-22
Revert "dev: Suggestion: Try to centralized command options in one file"

liangxin1300:
This reverts commit 52a44fdce0eb17ae36b8da9c430437054f6ee476.

Hi @krig @diegoakechi , please consider to revert this commit.
At the time when I want to create this commit, I'm investigating python prompt toolkit, the benefit to do this is for completion.

But to be honest, the biggest defect for this commit is uneasy to maintain each option because they are at different files:(


liangxin1300:
Issues about qdevice might have changes for options,
and bsc#1141976 also need to adjust options,
it's not easy to maintain this.

I really suggest to revert it

krig:
Reverting is OK by me, I'll let @diegoakechi decide :)

PR #468, opened 2019-08-22
Dev: ui_resource: maintenance: stop using crm_resource

aleksei-burlakov:
crm_resource doesn't work consistently when applying maintenante
attribute to a group of primitives.
https://github.com/ClusterLabs/pacemaker/pull/1858
The RscMgmt._commit_meta_attr works correctly.


aleksei-burlakov:
@krig , could you please look the change and give your feedback. The PR fails the test_maintenance in the unittests/test_resource.py, because it doesn't use the crm_resource anymore. 
I think the whole unittests/test_resource.py can be removed now. There should be however some test for the `$ crm resource maintenance` scenario. I think I should add them in the testcases/resource and the testcases/resource.exp. I've created a shotgun test, which (surprise) fails. Could you please also give a hint about the test?

liangxin1300:
Hi @aleksei-burlakov, looks good to me! Great job!

aleksei-burlakov:
> Hi @aleksei-burlakov, looks good to me! Great job!
> BTW, I think this can be merged after [ClusterLabs/pacemaker#1858](https://github.com/ClusterLabs/pacemaker/pull/1858) got merged

I think it's unnecessary. The purpose of this commit is exactly to stop using the crm_resource. This  PR will not be affected by https://github.com/ClusterLabs/pacemaker/pull/1858

krig:
Looks good to me as well!

PR #469, opened 2019-08-24
Dev: replace optparse with argparse

liangxin1300:
Hi:
Since `optparse` is **deprecated**, I suggest we should replace it with `argparse`.

There are some differences between `optparse` and `argparse` that should be noticed in replacing:
- in usage message, `optparse` using `%prog` as format specifier, while `argparse` using `%(prog)s`
- `argparse` using `version` in `add_argument` function
- to keep code compatible,  in `argparse`, using `parse_known_arg` instead of `parse_args`
- `optparse` using `add_help_option` while `argparse` using `add_help`
- `optparse` using `disable_interspersed_args` to set parsing to stop on the first non-option.
   For example, if don't using `disable_interspersed_args`, `crm cluster init -h` will be parsed as `crm -h` and print `crm` help message;
   In `argparse`, we don't have interface `disable_interspersed_args`, even [parse_intermixed_args](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_intermixed_args) would not be existed until python3.7, so I use this `parser.add_argument("SUBCOMMAND", nargs=REMAINDER)` to restore the non-option extra options

### DIFF
--- a/crmsh/main.py
+++ b/crmsh/main.py
@@ -84,58 +84,58 @@ def cib_prompt():
 
 
 def make_option_parser():
-    from optparse import OptionParser
-    parser = OptionParser(usage="""%prog [-h|--help] [OPTIONS] [SUBCOMMAND ARGS...]
-or %prog help SUBCOMMAND
+    from argparse import ArgumentParser, REMAINDER
+    parser = ArgumentParser(prog='crm', usage="""%(prog)s [-h|--help] [OPTIONS] [SUBCOMMAND ARGS...]
+or %(prog)s help SUBCOMMAND
 
-For a list of available subcommands, use %prog help.
+For a list of available subcommands, use %(prog)s help.
 
-Use %prog without arguments for an interactive session.
+Use %(prog)s without arguments for an interactive session.
 Call a subcommand directly for a "single-shot" use.
-Call %prog with a level name as argument to start an interactive
+Call %(prog)s with a level name as argument to start an interactive
 session from that level.
 
-See the crm(8) man page or call %prog help for more details.""",
-                          version="%prog " + config.CRM_VERSION)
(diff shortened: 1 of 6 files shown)


liangxin1300:
Thanks @diegoakechi , I think it's more suitable to test it with regression test instead of unit test, maybe we could use python-behave to assert right help messages printed

liangxin1300:
@diegoakechi @arbulu89  With functional tests, I think this PR could be merged.
BTW, bsc#1141976 depend on this PR;
And I also want to add option dependency for qdevice using "required" which only exists in argparse

liangxin1300:
hi @arbulu89 ,
Thanks!

> Remove new options from this PR (`qdevice` and the new things) and we can merge this PR. And after that, create individual new PRs adding `qdevice` and everything that we have added new.

qdevice and no-overwrite-sshkey options have already in master code, we could not change part of options by using argparse while left new options still using optparse, I think it's not possible.

liangxin1300:
> Now we want to backport the `no-overwrite-sshkey` option and the that PR depends on this one. So, to backporting that would carry qdevice options, something that I don't know if we want to do...

@arbulu89 , the backport PR #487 doesn't inlcude qdevice options.

liangxin1300:
@diegoakechi @arbulu89 I prefer to merge this PR.
Then for master branch, python3.x, we using argparse;
for 3.0 branch, python2.7, we keep using optparse

liangxin1300:
@arbulu89 test results are here: https://travis-ci.org/ClusterLabs/crmsh/jobs/634648781?utm_medium=notification&utm_source=github_status

liangxin1300:
@diegoakechi Please review and approve this, thanks!

PR #470, opened 2019-08-27
'configure show changed' includes deleted elements

liangxin1300:
**Problem:**
Our user complain that after  delete resource(before commit), 'crm configure show changed' not list the elements just deleted.
> Working with crm in SLES12 SP4 I noticed that when you delete resources in "crm configure", a > "show changed" just displays nothing.
> How hard would it be to show commands like "delete xyz" for each deleted resource?

**Solution:**
So I append this line:
```
obj_set |= orderedset.oset(self.remove_queue)
```
in `filter_objects` function, cibconfig.py file.

**Test:**
I also append lines in `test/testcases/commit`:
```
delete d3
show changed
```
to test this scenarios.
Behave seems not suitable to test interactive scenarios.

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -3223,7 +3223,7 @@ class CibFactory(object):
         obj_set = orderedset.oset([])
         and_filter, and_set = False, None
         for spec in filters:
-            if spec == "or":
+            if spec in ["or", "from_show"]:
                 continue
             elif spec == "and":
                 and_filter, and_set = True, obj_set
@@ -3231,6 +3231,8 @@ class CibFactory(object):
                 continue
             if spec == "changed":
                 obj_set |= orderedset.oset(self.modified_elems())
+                if "from_show" in filters:
+                    obj_set |= orderedset.oset(self.remove_queue)
             elif spec.startswith("type:"):
                 obj_set |= orderedset.oset(self.get_elems_on_type(spec))
             elif spec.startswith("tag:"):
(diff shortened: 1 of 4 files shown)


liangxin1300:
> @liangxin1300 I don't really understand the change.
> Shouldn't be the expected behaviour that the `delete` command is shown after `show changed`?
> At least in the test is not shown...
> Instead we print `primitive d2 Dummy`. Is this what you want?

@arbulu89 , `show changed` will show both modified and deleted elements.
Current `show changed` doesn't include `deleted` elements, just show modified

liangxin1300:
@arbulu89 maybe we could add `show changed modified` and `show changed deleted`?
I think it looks like a bit of complicated.
I think `show changed`(without more sub-parameters) is enough.

arbulu89:
> @arbulu89 maybe we could add `show changed modified` and `show changed deleted`?
> I think it looks like a bit of complicated.
> I think `show changed`(without more sub-parameters) is enough.

That's OK for me, but I don't know if matches with that user expectation. He won't find any output like `delete resource` (what he comments). He will find `resource changed` what is different. Maybe, when the resource is deleted, instead of printing `primitive d2 Dummy` he would prefer `primitive d2 Dummy deleted`. But I don't know if this is something easy to do

liangxin1300:
@arbulu89 I will collect suggestions from this user

liangxin1300:
New impl is #1730

PR #471, opened 2019-08-28
Dev: ui_resource: resolve maintenance vs is-managed conflict

aleksei-burlakov:
The maintenance and is-managed are conflicting atributes. When setting
one of them, the user is proposed to remove another.

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -44,9 +44,26 @@ def get_children_with_different_attr(node, attr, value):
             l.append(p)
     return l
 
+def get_children_with_attr(node, attr):
+    l = []
+    for p in node.xpath(".//primitive"):
+        diff_attr = False
+        for meta_set in xmlutil.get_set_nodes(p, "meta_attributes", create=False):
+            p_value = xmlutil.get_attr_value(meta_set, attr)
+            if p_value is not None:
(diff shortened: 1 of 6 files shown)


aleksei-burlakov:
@krig, @liangxin1300, please review the PR.

aleksei-burlakov:
> Please add a unit test that verifies the change :)

@krig , I have added a regression test.

aleksei-burlakov:
@krig, could you please check it out?

krig:
It would be good to include a link to the documentation or some kind of clarification in the info message, I think just the question like that is going to confuse people who aren’t clear about the difference between is-managed and maintenance mode.

aleksei-burlakov:
> It would be good to include a link to the documentation or some kind of clarification in the info message, I think just the question like that is going to confuse people who aren’t clear about the difference between is-managed and maintenance mode.

https://github.com/ClusterLabs/pacemaker/blob/13b7e117dfba0603791d9d72dc0a4d3737a256ec/doc/Pacemaker_Explained/en-US/Ch-Resources.txt#L337

This exists only on the github now.

aleksei-burlakov:
@krig, I have changed the output message and the help for crm maintenance and crm manage.

krig:
Perfect, thank you!

PR #476, opened 2019-10-11
Low: bootstrap: don't overwrite ssh key if already exists

liangxin1300:
## Problem
Using **--yes/-y** option with **ha-cluster-init** will cause re-generate/overwrite ssh key if ssh key already exists.
That may cause problem or inconvenient for user.
#### Problem 1
One colleague complain to me that he can not access the node passwordless after configured HA cluster because ssh key changed
I think this is the inconvenient point other user might have
#### Problem 2
That will cause inconvenient during regression test for #464.
In [.travis.yml](https://github.com/liangxin1300/crmsh/blob/b8198e60405473c3201730adbbc212c17f53930a/.travis.yml#L42), I start two nodes for cluster, another one node for qnetd.
Since both setup cluster and setup qdevice/qnetd need to input password, I deploy ssh key firstly in image ([Dockerfile](https://github.com/liangxin1300/myDockerfiles/blob/0b3d55e110f3d3902e1123032efa91341b02118c/HA/habase/tumbleweed/Dockerfile#L20)), to make sure the whole process are passwordless/non-interactive.
But the process will block when setup qdevice/qnetd because my initial ssh key has been overwrite.
## Solution
I suggest if ssh key already exists before HA cluster configuring, we do not overwrite it, just return and continue to do next init steps.
## Affect
From my view, there is no side effect for any interfaces

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -670,9 +670,11 @@ def init_ssh():
     start_service("sshd.service")
     invoke("mkdir -m 700 -p /root/.ssh")
     if os.path.exists("/root/.ssh/id_rsa"):
-        if not os.path.exists("/root/.ssh/authorized_keys"):
-            append("/root/.ssh/id_rsa.pub", "/root/.ssh/authorized_keys")
-        return
+        if not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
+            return
+        if _context.yes_to_all and _context.no_overwrite_sshkey:
+            return
+        rmfile("/root/.ssh/id_rsa")
     status("Generating SSH key")
     invoke("ssh-keygen -q -f /root/.ssh/id_rsa -C 'Cluster Internal' -N ''")
     append("/root/.ssh/id_rsa.pub", "/root/.ssh/authorized_keys")
@@ -2053,7 +2055,7 @@ def remove_localhost_check():
 
 def bootstrap_init(cluster_name="hacluster", ui_context=None, nic=None, ocfs2_device=None,
                    shared_device=None, sbd_device=None, diskless_sbd=False, quiet=False,
-                   template=None, admin_ip=None, yes_to_all=False,
+                   template=None, admin_ip=None, yes_to_all=False, no_overwrite_sshkey=False,
                    unicast=False, second_hb=False, ipv6=False, watchdog=None, qdevice=None, stage=None, args=None):
     """
     -i <nic>
@@ -2094,6 +2096,7 @@ def bootstrap_init(cluster_name="hacluster", ui_context=None, nic=None, ocfs2_de
     _context.watchdog = watchdog
     _context.ui_context = ui_context
     _context.qdevice = qdevice
+    _context.no_overwrite_sshkey = no_overwrite_sshkey
 
     def check_option():
         if _context.admin_ip and not valid_adminIP(_context.admin_ip):
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -670,9 +670,8 @@ def init_ssh():
     start_service("sshd.service")
     invoke("mkdir -m 700 -p /root/.ssh")
     if os.path.exists("/root/.ssh/id_rsa"):
-        if not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
-            return
-        if _context.yes_to_all and _context.no_overwrite_sshkey:
+        if _context.yes_to_all and _context.no_overwrite_sshkey or \
+                not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
             return
         rmfile("/root/.ssh/id_rsa")
     status("Generating SSH key")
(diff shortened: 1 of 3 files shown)


diegoakechi:
@liangxin1300 We have a conception problem here. Look what the `-y` stands for:

```
-y, --yes             Answer "yes" to all prompts (use with caution, this is
                        destructive, especially during the "storage" stage)
```

That basically means you are answering YES to all the questions, including the one that asks if you want to override the keys or not. Removing it will change completely the behavior including the interactive one, just to fulfill your use case.

My suggestion here is that you improve the help message to describe that the ssh will be overwritten too. 

To fulfill your use case, I think we could have a complementary option to not overwrite the SSH keys, so this way it does not change the behavior of `-y` and on interactive mode.

liangxin1300:
Thanks @diegoakechi , I've added "--no-overwrite-sshkey" option to avoid ssh key be overwritten when using "-y/--yes" option, and improved related help messages.

I will add test case here later using python-behave, also as the start to test the whole bootstrap process

liangxin1300:
While working on this, I found another problem here:
When doing ha-cluster-init, if ssh key exists and authorized_keys not exists or not contain public key, then we skip overwrite ssh key,  at last, join node will failed with setup passwordless.
So I append commit dbee1c4 to resolve this problem

test/features/bootstrap.feature
```diff
@@ -0,0 +1,29 @@
+Feature: HA bootstrap process
```
arbulu89: Maybe we can add some documentation to the Feature

test/features/environment.py
```diff
@@ -0,0 +1,94 @@
...
+    will be skipped.'''
+
+def before_feature(context, feature):
+    if feature.name == "HA bootstrap process":
+        context.init_help_message = CLUSTER_INIT_HELP_MESSAGE
+
+def before_scenario(context, scenario):
+    if utils.check_service_active('corosync.service', 'local'):
+        utils.run_command('crm cluster stop')
+    if scenario.name == "No overwrite ssh key but append to authorized_keys":
```
arbulu89: This looks quite risky. I wouldn't like to have specific conditions for particular test scenarios. Maybe we can perform this action in the test itself

test/features/environment.py
```diff
@@ -0,0 +1,94 @@
...
+    in sequence, with prompts for required information.
+  - If using the ocfs2 template, the storage stage will partition a block
+    device into two pieces, one for SBD, the remainder for OCFS2.  This is
+    good for testing and demonstration, but not ideal for production.
+    To use storage you have already configured, pass -s and -o to specify
+    the block devices for SBD and OCFS2, and the automatic partitioning
+    will be skipped.'''
+
+def before_feature(context, feature):
+    if feature.name == "HA bootstrap process":
```
arbulu89: I don't think it's a good idea to use this kind of conditions here... 
I don't know how to manage that anyway

test/features/steps/utils.py
```diff
@@ -0,0 +1,18 @@
...
+def run_command(cmd, stdout=False, stderr=False):
+    rc, out, err = utils.get_stdout_stderr(cmd)
+    if out and stdout:
+        print(out)
+    if err and stderr:
+        print(err)
+    return rc, out, err
+
+def check_cluster_status(nodelist):
+    if nodelist == "local":
```
arbulu89: Why are we using this `local` variable. All the tests are executed in local so far

test/features/steps/utils.py
```diff
@@ -0,0 +1,18 @@
+import os
+from crmsh import utils, bootstrap
+
+def run_command(cmd, stdout=False, stderr=False):
+    rc, out, err = utils.get_stdout_stderr(cmd)
+    if out and stdout:
+        print(out)
```
arbulu89: I would avoid the `print` usage... Maybe we can implement a proper logger

liangxin1300:
Remove python-behave part

crmsh/bootstrap.py
```diff
@@ -670,17 +670,26 @@ def init_ssh():
     start_service("sshd.service")
     invoke("mkdir -m 700 -p /root/.ssh")
     if os.path.exists("/root/.ssh/id_rsa"):
-        if not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
+        if _context.yes_to_all and _context.no_overwrite_sshkey or \
+           not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
```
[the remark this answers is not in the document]
liangxin1300: Ok
Changed

liangxin1300:
> ~@liangxin1300 I think there is some miss match in the logic. The condition to remove or not is not the good one. Could you have a look there? At least the UT you added don't follow the behaviour they should. Thanks!~
> I have rechecked, and after that I think the logic of the code is good, what is failling is the UT, they are not done properly, the use cases must be fixed.
> Thanks

Thanks @arbulu89 I've changed the UT, key point is `mock_confirm` should return True when:
`bootstrap._context = mock.Mock(yes_to_all=True, no_overwrite_sshkey=False)`

PR #477, opened 2019-10-16
Set resource failcount correctly(bsc#1144241)

liangxin1300:
#### Problem
Can't set the resource failcount correctly.
e.g.
Dummy resource `d` failed on monitor on node1, using
`crm resource failcount d show node1`
shows the failcount number
If we want to use this:
`crm resource failcount d set node1 0`
to reset the failcount to zero, it doesn't work 
#### Reason

> from SLE12SP3 and SLE15,  fail-counts are no longer recorded "per resource", instead they are recorded "per resource operation per node". 
> So if to make "crm resource failcount <rsc> set" correctly work again, the interface needs to be completed/extended, for example like:
> crm resource failcount \<rsc\> set \<node\> \<operation\> \<interval\> \<value\>

#### Solution
I changed this set sub-command like this:
`failcount <rsc> set <node> <value> [operation]`
If user want to set the failcount, program fetch the current failcount status from cib firstly, to get the operations and their interval.
If only one operation failure happened, option `operation` is no need to specify;
If there are multiple operation failures, user must specify option `operation` ;
Interval of operation always get internally.

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -203,7 +203,8 @@ class RscMgmt(command.UI):
         'get': "crm_resource --meta --resource '%s' --get-parameter '%s'",
     }
     rsc_failcount = {
-        'set': "crm_attribute -t status -n 'fail-count-%s' -N '%s' -v '%s' -d 0",
+        'set': "crm_attribute -t status -n 'fail-count-%s' -N '%s' -v '%s'",
+        'set_p': "crm_attribute -t status -P 'fail-count-%s' -N '%s' -v '%s'",
         'delete': "crm_failcount -D -r %s -N %s",
         'show': "crm_failcount -G -r %s -N %s",
         'get': "crm_failcount -G -r %s -N %s",
@@ -483,11 +484,75 @@ class RscMgmt(command.UI):
 
     @command.wait
     @command.completers(compl.resources, _attrcmds, compl.nodes)
-    def do_failcount(self, context, rsc, cmd, node, value=None):
+    def do_failcount(self, context, rsc, cmd, node, value=None, operation=None, interval=None):
         """usage:
-        failcount <rsc> set <node> <value>
+        failcount <rsc> set <node> <value> [operation] [interval]
         failcount <rsc> delete <node>
         failcount <rsc> show <node>"""
+        def sec_to_ms(s):
+            return s + '000'
+
+        def ms_to_sec(m):
+            return m[:len(m)-3]
+
+        if rsc not in compl.resources():
+            context.fatal_error("Resource {} not exists in this cluster".format(rsc))
+        valid_cmd_list = ["set", "delete", "show"]
+        if cmd not in valid_cmd_list:
+            context.fatal_error("{} is not valid command(should be one of {})".format(cmd, valid_cmd_list))
+        nodeid = utils.get_nodeid_from_name(node)
+        if nodeid is None:
+            context.fatal_error("Node {} not in this cluster".format(node))
+
+        if cmd == "set":
+            # query the current failcount status
+            query_cmd = "cibadmin -Ql --xpath '/cib/status/node_state[@id='{}']'".format(nodeid)
+            rc, out, err = utils.get_stdout_stderr(query_cmd)
+            if rc != 0:
+                context.fatal_error(err)
+
+            # try to get failcount dict {operation:interval}
+            import re
+            failcount_res = re.findall(r'fail-count-{}#(.*)_([0-9]+)'.format(rsc), out)
+            if not failcount_res:
+                context.fatal_error("No failcount on node {} for resource {}".format(node, rsc))
+            failcount_dict = dict(failcount_res)
+
+            # validate for operation and interval
+            if operation and operation not in failcount_dict.keys():
+                context.fatal_error("Usage: failcount <rsc> set <node> <value> [operation] [interval]\n\
+                           Should specify operation between \"{}\"".format(' '.join(failcount_dict.keys())))
+            if (operation and interval) and (operation, sec_to_ms(interval)) not in failcount_res:
+                context.fatal_error("Usage: failcount <rsc> set <node> <value> [operation] [interval]\n\
+                           Should specify (operation, interval) between {}".\
+                           format([(op, ms_to_sec(inter)) for op, inter in failcount_res]))
+
+            # just one failcount entry
+            if len(failcount_res) == 1:
+                operation = failcount_res[0][0]
+                interval = failcount_dict[operation]
+                rsc = '{}#{}_{}'.format(rsc, operation, interval)
+
+            # multiple failcount entries for this resource and node
+            if len(failcount_res) > 1:
+                if operation and interval:
+                    rsc = '{}#{}_{}'.format(rsc, operation, sec_to_ms(interval))
+                elif int(value) == 0:
+                    # using '-P' option of 'crm_attribute' command
+                    cmd = "set_p"
+                    if operation:
+                        op_interval_str = '|'.join(['{}_{}'.format(operation, inter) for op, inter in failcount_res if op==operation])
+                    else:
+                        op_interval_str = '|'.join(['{}_{}'.format(op, inter) for op, inter in failcount_res])
+                    rsc = '{}#({})'.format(rsc, op_interval_str)
+                else:
+                    # value != 0
+                    if operation and len([op for op, _ in failcount_res if op == operation]) == 1:
+                        rsc = '{}#{}_{}'.format(rsc, operation, failcount_dict[operation])
+                    else:
+                        context.fatal_error("Should specify (operation, interval) between {}".
+                                format([(op, ms_to_sec(inter)) for op, inter in failcount_res]))
+
         return ui_utils.manage_attr(context.get_command_name(), self.rsc_failcount,
                                     rsc, cmd, node, value)
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,12 @@ jobs:
       script:
         - $FUNCTIONAL_TEST qdevice run validate
 
+    - name: "functional test for resource subcommand"
+      before_install:
+        - $FUNCTIONAL_TEST resource before_install
+      script:
+        - $FUNCTIONAL_TEST resource run
+
     - stage: delivery
       if: type != pull_request AND branch = master
       env:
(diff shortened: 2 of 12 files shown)


liangxin1300:
Testing is WIP

liangxin1300:
@diegoakechi @arbulu89 Would you take a look at testing part of #476 ?
If it make sense, please merge it and then I could base on that to do regression test for this PR.
Otherwise there will have some file conflicts

gao-yan:
Thanks for the nice work. The idea is quite sensible to me. I like the user friendliness about being intelligent on automatically determining the "operation" field.

One additional thing needs to think about is, users usually configure two monitors for a promotable resource (master/slave type)  respectively  for role=Master or role=Slave.

Of course if there's only one fail-count entry, crmsh still can be smart enough without even requiring 'operation' to be specified by user.  But if multiple monitor fail-count entries are existing, users would have to be specific about which monitor with what "role"/"interval" they want to set failcount for.

liangxin1300:
@gao-yan , do you think it's possible that multiple monitor fail-count entries existing on the same node?
As far as I know, master and slave are running two nodes, when both failures happened, two fail-count entries will be created separated by each node;
And for this `set` sub-command, user must specific the node to set the fail count

liangxin1300:
@gao-yan I've updated the code, try to keep intelligent/friendly by using '-P' option for `crm_attribute` command.
When multiple entries exist for this specific resource and node, if `operation` or `interval` not provided,
we use this format to set the value:
`crm_attribute -t status -P 'fail-count-<rsc>#(op1_inter1|op1_inter2|op2_inter2)' -N '<node>' -v '<value>' -d 0`

crmsh/ui_resource.py
```diff
@@ -218,6 +218,7 @@ class RscMgmt(command.UI):
     }
     rsc_failcount = {
         'set': "crm_attribute -t status -n 'fail-count-%s' -N '%s' -v '%s' -d 0",
+        'set_p': "crm_attribute -t status -P 'fail-count-%s' -N '%s' -v '%s' -d 0",
```
gao-yan: `-d <value>` is meaningful only when being used with "--query/-G" though, meaning if there's no such an entry in cib, the caller of this command gives himself a `default_value` defined by himself rather than getting an error.

crmsh/ui_resource.py
```diff
@@ -497,11 +498,54 @@ def do_constraints(self, context, resource):
...
+            failcount_res = re.findall(r'fail-count-{}#(.*)_([0-9]+)'.format(rsc), out)
+            if not failcount_res:
+                context.fatal_error("No failcount on node {}".format(node))
+            failcount_dict = dict(failcount_res)
+
+            # validate for operation and interval
+            if operation and operation not in failcount_dict.keys():
+                context.fatal_error("Should specify operation between \"{}\"".format(' '.join(failcount_dict.keys())))
+            if (operation and interval) and (operation, interval) not in failcount_res:
+                context.fatal_error("Should specify (operation, interval) between {}".format(failcount_res))
```
gao-yan: We don't need to prevent users from setting a fail-count that is not existing yet.

crmsh/ui_resource.py
```diff
@@ -497,11 +498,54 @@ def do_constraints(self, context, resource):
...
+                    rsc = '{}#{}_{}'.format(rsc, operation, interval)
+                else:
+                    # using '-P' option of 'crm_attribute' command
+                    cmd = "set_p"
+                    if operation:
+                        op_interval_str = '|'.join(['{}_{}'.format(operation, inter) for op, inter in failcount_res if op==operation])
+                    else:
+                        op_interval_str = '|'.join(['{}_{}'.format(op, inter) for op, inter in failcount_res])
+                    rsc = '{}#({})'.format(rsc, op_interval_str)
+
```
gao-yan: I'd go with some logic like the following. Although it looks a little sophisticated, but it'd be very friendly ;-)
```
if operation == None and interval == None:
    if len(existing_failcount_entries) == 0:
        if rsc_has_only_one_monitor_configured(rsc):
            # Set fail-count for monitor with the _configured_ interval (don't forget *1000)
        else:
            # Error: Please specify operation and interval
    elif len(existing_failcount_entries) == 1:
        # Set the only existing fail-count entry to value
    elif len(existing_failcount_entries) > 1:
        if value == 0:
            # Set all the existing fail-count entries to 0
        else:
            # Error: Please specify operation and interval
elif operation != None:
    if not is_valid_op(rsc, operation): #start/stop/monitor, promote/demote if rsc is promotable
        # Error: Please specify a valid operation
    elif interval == None:
        if len(failcount_entries_matching_operation) == 0:
            if is_valid_one_shot_op(rsc, operation):  #start/stop, promote/demote if rsc is promotable 
                # Set fail-count for the operation with interval 0
            elif operation == "monitor" and rsc_has_only_one_monitor_configured(rsc):
                # Set fail-count for monitor with the _configured_ interval
            else:
                # Error: Please specify operation and interval
        elif len(failcount_entries_matching_operation) == 1:
            # Set the only fail-count entry matching the operation to value
        elif len(failcount_entries_matching_operation) > 1:
            if value == 0:
                # Set all the fail-counts matching the operation to 0
            else:
                #  Error: Please specify operation and interval
    else: # interval != None
        if len(failcount_entries_matching_operation_interval) > 0:
            # Set fail-count for operation with the specified interval anyway
        elif is_valid_one_shot_op(rsc, operation) and interval != 0:
                # Error: interval should be 0 or omitted for operation %s
        elif operation == "monitor" and not rsc_has_monitor_configured_with_interval(rsc, interval):
                # Error: Please specify a configured monitor interval for rsc
        else:
                # Set fail-count with the specified operation and interval
else: # operation == None and interval != None -- is it possible?
    ...
```    
Another thing is, we accept interval in "seconds" specified at command line interface, but we need to convert it to "ms" as suffix of the fail-count entry name.

crmsh/ui_resource.py
liangxin1300: But setting a non-exist entry make no sense.
And validate here also will be helpful for misspelling.

crmsh/ui_resource.py
liangxin1300: OK, I've removed `-d` for both 'set' and 'set_p'

crmsh/ui_resource.py
liangxin1300: @gao-yan Thanks for your suggestions. 
I might keep my opinion.
1) This bugs' requirement from our customer is simple: 
`Keep this ticket to track "crm resource failcount <rsc> set <node> 0" fix.`
That will work if cib has one or more failcount entries;
In doc, I give the syntax like:
`failcount <rsc> set <node> <value> [operation] [interval]`
That will also work if user input the `<operation> and <interval>`, I will always do validation work if they were not null, to lead user give the right and exists `<operation> and <interval>` pair.
2) I think it make no sense to create a new fail-count entry if this fail-count for this operation not exist, and that is not the requirement from customer. So if there is no fail-count entry exists while running this command, I will give error like:
`No failcount on node {} for resource {}`

crmsh/ui_resource.py
```diff
@@ -497,11 +498,54 @@ def do_constraints(self, context, resource):
...
+            if len(failcount_res) > 1:
+                if operation and interval:
+                    rsc = '{}#{}_{}'.format(rsc, operation, interval)
+                else:
+                    # using '-P' option of 'crm_attribute' command
+                    cmd = "set_p"
+                    if operation:
+                        op_interval_str = '|'.join(['{}_{}'.format(operation, inter) for op, inter in failcount_res if op==operation])
+                    else:
+                        op_interval_str = '|'.join(['{}_{}'.format(op, inter) for op, inter in failcount_res])
```
gao-yan: Hrm, it works correctly only if  value 0 is specified, doesn't it? For example with a specified value 2, it'll set every matched fail-count entries to 2, which will eventually contribute `2 * n` to the resulting fail-count (sum), won't it? 

I think we still need to differentiate cases for 0 and non-zero values. And for non-zero cases, check how many fail-count entries are really matched.  Set it if only one is matched, otherwise drop an error asking user to be specific.

liangxin1300: @gao-yan 
1) I've added the transition between ms to sec and sec to ms;
2) For non-zero set to multiple entries, I've changed, drop an error for input specific (operation, interval)
Above both are mentioned in doc.
Thanks for suggestions!

crmsh/ui_resource.py
```diff
@@ -497,11 +498,67 @@ def do_constraints(self, context, resource):
...
+                    rsc = '{}#({})'.format(rsc, op_interval_str)
+                else:
+                    # value != 0
+                    context.fatal_error("Should specify (operation, interval) between {}".
+                            format([(op, ms_to_sec(inter)) for op, inter in failcount_res]))
```
gao-yan: Thanks for the change. Of course like I mentioned, it would be more friendly with a logic like: "In the case of `len(failcount_res) > 1`, for a `non-zero value`, if `only one` fail-count entry is _matched_ with the specified `operation`, even if `interval` is not specified, just set it ".

liangxin1300: Great suggestion! I've changed like this:
```
                    # value != 0
                    if operation and len([op for op, _ in failcount_res if op == operation]) == 1:
                        rsc = '{}#{}_{}'.format(rsc, operation, failcount_dict[operation])
                    else:
                        context.fatal_error("Should specify (operation, interval) between {}".
                                format([(op, ms_to_sec(inter)) for op, inter in failcount_res]))
```

crmsh/ui_resource.py
```diff
@@ -497,11 +498,70 @@ def do_constraints(self, context, resource):
...
+                    cmd = "set_p"
+                    if operation:
+                        op_interval_str = '|'.join(['{}_{}'.format(operation, inter) for op, inter in failcount_res if op==operation])
+                    else:
+                        op_interval_str = '|'.join(['{}_{}'.format(op, inter) for op, inter in failcount_res])
+                    rsc = '{}#({})'.format(rsc, op_interval_str)
+                else:
+                    # value != 0
+                    if operation and len([op for op, _ in failcount_res if op == operation]) == 1:
+                        rsc = '{}#{}_{}'.format(rsc, operation, failcount_dict[operation])
```
gao-yan: Looks good to me. Thanks for the change.

gao-yan:
Looks good to me. I like how the tests are conducted. Nice work, thanks!

PR #478, opened 2019-10-17
Node do maintenance

aleksei-burlakov:
The crmsh will change the cib.xml itself, instead of using the crm_attribute when changing the maintenance status of a node. Besides, it will inform the user if there are resources in the maintenance or is-manage mode.

PR #479, opened 2019-10-18
Dev: ui_resource: ask about ALL primitives when overriding attributes

aleksei-burlakov:
When applying a maintenance attribute, the crmsh asked the user if he wanted to delete the maintenance attribute for those primitives that had a different value. It's more logical however to ask the user if he also wants to delete the maintenance attribute from the primitives that have the same value, because this is exactly the intention - to maintain the group not each primitive separately. Even if the user would like to customize each resource individually, he can answer 'No'.

### DIFF
--- a/test/testcases/resource.exp
+++ b/test/testcases/resource.exp
@@ -712,11 +712,7 @@ INFO: Trace set, restart p0 to trace the stop operation
           <nvpair id="cg-meta_attributes-target-role" name="target-role" value="Started"/>
         </meta_attributes>
         <group id="g">
-          <primitive id="p0" class="ocf" provider="pacemaker" type="Dummy">
-            <meta_attributes id="p0-meta_attributes">
-              <nvpair id="p0-meta_attributes-target-role" name="target-role" value="Started"/>
-            </meta_attributes>
-          </primitive>
+          <primitive id="p0" class="ocf" provider="pacemaker" type="Dummy"/>
           <primitive id="p3" class="ocf" provider="pacemaker" type="Dummy"/>
         </group>
       </clone>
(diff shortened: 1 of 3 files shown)


aleksei-burlakov:
@krig , @diegoakechi , @gao-yan please review this PR. The motivation of this PR is to make the rules when applying attributes simpler. 

Basically there is one condition less when applying an attribute to a group of resources. On this step, it doesn't show much simplification, but it will when we apply attributes to a node (https://github.com/ClusterLabs/crmsh/pull/478).

aleksei-burlakov:
@krig , @diegoakechi , @gao-yan please review this PR. The motivation of this PR is to make the rules when applying attributes simpler. 

Basically there is one condition less when applying an attribute to a group of resources. On this step, it doesn't show much simplification, but it will when we apply attributes to a node (https://github.com/ClusterLabs/crmsh/pull/478).

test/testcases/resource.exp
```diff
@@ -712,11 +712,7 @@ INFO: Trace set, restart p0 to trace the stop operation
           <nvpair id="cg-meta_attributes-target-role" name="target-role" value="Started"/>
         </meta_attributes>
         <group id="g">
-          <primitive id="p0" class="ocf" provider="pacemaker" type="Dummy">
-            <meta_attributes id="p0-meta_attributes">
-              <nvpair id="p0-meta_attributes-target-role" name="target-role" value="Started"/>
-            </meta_attributes>
-          </primitive>
+          <primitive id="p0" class="ocf" provider="pacemaker" type="Dummy"/>
```
aleksei-burlakov: There is a group g containing the primitive p0. The primitive p0 has a attribute Started=true. When setting the Started=true for the group g it will delete the Started=true from the p0.

PR #482, opened 2019-11-01
Create class Parallax to simplify using parallax

liangxin1300:
#### Motivation
crmsh use parallax a lot, recently:
* #464 In qdevice/qnetd TLS certification process, need call remote command and exchange remote key files
* #483  Doing regression test on bootstrap process using python-behave, need call remote command

So we should simplify and provide a standard way to use parallax, to avoid set parallax options and handle the exceptions each time or forget to set something like ssh options

### DIFF
--- a/crmsh/parallax.py
+++ b/crmsh/parallax.py
@@ -1,0 +1,104 @@
+# Copyright (C) 2019 Xin Liang <XLiang@suse.com>
+# See COPYING for license information.
+
+
+import os
+import parallax
+
+
+class Parallax(object):
+    """
+    # Parallax SSH API
+    # call: Executes the given command on a set of hosts, collecting the output
+    # copy: Copies files from the local machine to a set of remote hosts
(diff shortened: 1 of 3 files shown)


liangxin1300:
@krig Please review it because you are the author of parallax:)

liangxin1300:
@diegoakechi @arbulu89 , we should drive this PR to merge first since it is the prerequisite for #464 and #481

krig:
Looks good to me, I don't have any objections! I wouldn't mind adding the class to the parallax library directly either, if you want to do that.

liangxin1300:
@arbulu89 I've moved it into a new file, please review:)

liangxin1300:
> PD: I guess the new parallax UT are being executed in the travis, right? I cannot find the individual tests result...

@arbulu89 
Yes, parallax UT already been executed. The unittest result would not be shown until errors happen, in normal status, it just print `...` in console

PR #483, opened 2019-11-05
Doing functional test on bootstrap process using python-behave

liangxin1300:
#### Motivation
* Most of bugs for crmsh, came from bootstrap and hb_report, those two features which are also changed frequently, are lack of functional testing.
* New feature for supporting Qdevice also need functional testing. See #464 
* Existing regression test based on shell is not suitable to implement above two requirements.
* Python-behave is behaviour-driven development, Python style.
Integrated with TravisCI and docker, functional testing for above requirements are easy to implement and maintain.

#### Plan
* This PR is using behave to test basic functionalities of bootstrap;
  Drive this PR to merge, as a baseline.
* Apply python-behave on key features of crmsh, like qdevice and hb_report, which are lack of functional testing;
* Apply python-behave on new coming PRs of crmsh for functional testing;
* To replace existing functional test codes written by shell

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
 jobs:
   include:
     - stage: test
+      name: "original regression test"
       before_script:
         - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
         - chmod +x ./cc-test-reporter
@@ -33,6 +34,18 @@ jobs:
       after_script:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
   
+    - name: "functional test for bootstrap - init, join and remove"
+      before_install:
+        - $TRAVIS_BUILD_DIR/test/docker_scripts.sh bootstrap before_install
+      script:
+        - $TRAVIS_BUILD_DIR/test/docker_scripts.sh bootstrap run init_join_remove
+
+    - name: "functional test for bootstrap - options"
+      before_install:
+        - $TRAVIS_BUILD_DIR/test/docker_scripts.sh bootstrap before_install
+      script:
+        - $TRAVIS_BUILD_DIR/test/docker_scripts.sh bootstrap run options
+
     - stage: delivery
       if: type != pull_request AND branch = master
       env:
--- a/data-manifest
+++ b/data-manifest
@@ -63,7 +63,14 @@ test/containerized-regression-tests.sh
 test/crm-interface
 test/defaults
 test/descriptions
+test/docker_scripts.sh
 test/evaltest.sh
+test/features/bootstrap_init_join_remove.feature
+test/features/bootstrap_options.feature
+test/features/environment.py
+test/features/steps/__init__.py
+test/features/steps/step_bootstrap.py
+test/features/steps/utils.py
 test/history-test.tar.bz2
 test/list-undocumented-commands.py
 test/profile-history.sh
@@ -164,6 +171,7 @@ test/unittests/test_corosync.py
 test/unittests/test_gv.py
 test/unittests/test_handles.py
 test/unittests/test_objset.py
+test/unittests/test_parallax.py
 test/unittests/test_parse.py
 test/unittests/test_report.py
 test/unittests/test_scripts.py
--- a/test/docker_scripts.sh
+++ b/test/docker_scripts.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+Tumbleweed_image='liangxin1300/hatbw'
+HA_packages='pacemaker corosync corosync-qdevice csync2 python3 python3-lxml python3-python-dateutil python3-parallax libglue-devel python3-setuptools python3-tox asciidoc autoconf automake make pkgconfig which libxslt-tools mailx procps python3-nose python3-PyYAML python3-curses tar python3-behave iproute2 iputils vim bzip2'
+TEST_TYPE='bootstrap qdevice hb_report'
+
+before() {
+  docker pull ${Tumbleweed_image}
+  docker network create --subnet 10.10.10.0/24 second_net
+
+  # deploy first node hanode1
+  docker run -d --name=hanode1 --hostname hanode1 \
+             --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v "$(pwd):/app" ${Tumbleweed_image}
+  docker network connect --ip=10.10.10.2 second_net hanode1
+  docker exec -t hanode1 /bin/sh -c "echo \"10.10.10.3 hanode2\" >> /etc/hosts"
+  docker exec -t hanode1 /bin/sh -c "zypper -n in ${HA_packages}"
+
+  # deploy second node hanode2
+  docker run -d --name=hanode2 --hostname hanode2 \
+             --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v "$(pwd):/app" ${Tumbleweed_image}
+  docker network connect --ip=10.10.10.3 second_net hanode2
+  docker exec -t hanode2 /bin/sh -c "echo \"10.10.10.2 hanode1\" >> /etc/hosts"
+  docker exec -t hanode2 /bin/sh -c "zypper -n in ${HA_packages}"
+  docker exec -t hanode2 /bin/sh -c "systemctl start sshd.service"
+  docker exec -t hanode2 /bin/sh -c "cd /app; ./test/run-in-travis.sh build"
+}
+
+run() {
+  docker exec -t hanode1 /bin/sh -c "cd /app; ./test/run-in-travis.sh $1 $2"
+}
+
+usage() {
+  echo "Usage: ./test/`basename $0` <`echo ${TEST_TYPE// /|}`>"
+}
+
+# $1 could be "bootstrap", "hb_report", "qdevice" etc.
+# $2 could be "before_install" or "run"
+# $3 could be suffix of feature file
+case "$1/$2" in
+  */before_install)
+    before $1
+    ;;
+  */run)
+    run $1 $3
+    ;;
+  *)
+    if [ "$#" -eq 0 ] || ! [[ $TEST_TYPE =~ (^|[[:space:]])$1($|[[:space:]]) ]];then
+      usage
+      exit 1
+    fi
+    before $1
+    run $1
+    ;;
+esac
--- a/test/features/bootstrap_init_join_remove.feature
+++ b/test/features/bootstrap_init_join_remove.feature
@@ -1,0 +1,36 @@
+@bootstrap
+Feature: crmsh bootstrap process - init, join and remove
+
+  Test crmsh bootstrap init/join/remove process
+  Tag @clean means need to stop cluster service if the service is available
+
+  @clean
+  Scenario: Init cluster service on node "hanode1", and join on node "hanode2"
+    Given   Cluster service is "stopped" on "hanode1"
+    And     Cluster service is "stopped" on "hanode2"
+    When    Run "crm cluster init -y --no-overwrite-sshkey" on "hanode1"
+    And     Run "crm cluster join -c hanode1 -y" on "hanode2"
+    Then    Cluster service is "started" on "hanode1"
+    And     Cluster service is "started" on "hanode2"
+    And     Online nodes are "hanode1 hanode2"
+
+  @clean
+  Scenario: Remove peer node "hanode2"
+    Given   Cluster service is "stopped" on "hanode1"
+    And     Cluster service is "stopped" on "hanode2"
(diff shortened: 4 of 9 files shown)


liangxin1300:
@diegoakechi @arbulu89 , please review this, thanks!

test/docker_scripts.sh
```diff
@@ -0,0 +1,40 @@
...
+
+  # deploy second node hanode2
+  docker run -d --name=hanode2 --hostname hanode2 \
+             --net first_net --ip 10.10.10.3 \
+             --privileged -v /sys/fs/cgroup:/sys/fs/cgroup:ro -v "$(pwd):/app" ${Tumbleweed_image}
+  docker network connect --ip=20.20.20.3 second_net hanode2
+  docker exec -t hanode2 /bin/sh -c "echo \"10.10.10.2 hanode1\" >> /etc/hosts"
+  docker exec -t hanode2 /bin/sh -c "zypper -n in ${HA_packages}"
+  docker exec -t hanode2 /bin/sh -c "systemctl start sshd.service"
+  docker exec -t hanode2 /bin/sh -c "cd /app; ./test/run-in-travis.sh build"
```
arbulu89: What is this last `docker exec -t hanode2 /bin/sh -c "cd /app; ./test/run-in-travis.sh build"` line for?
And why it's executed in node 2?

.travis.yml
```diff
@@ -33,6 +34,24 @@ jobs:
       after_script:
         - ./cc-test-reporter after-build --exit-code $TRAVIS_TEST_RESULT
   
+    - name: "regression test using python-behave thread 1"
```
arbulu89: I don't really like these names. `thread 1` is not really meaningfull,

Can we add better names? What is this thread for?

test/run-in-travis.sh
```diff
@@ -21,8 +21,25 @@ regression_tests() {
...
 }
 
-unit_tests
-rc_unittest=$?
-configure
-make_install
-regression_tests && exit $rc_unittest
+bootstrap_tests() {
+	echo "** Bootstrap process tests using python-behave"
+        behave --no-logcapture --tags "@bootstrap" --tags "~@wip" /usr/share/crmsh/tests/features/bootstrap_$1.feature
```
arbulu89: I would change that, and pass as argument the howl `.feature` file name. It would be easier to expand in the future and we can have more flexibility selecting `feature` file names

test/features/bootstrap_1.feature
```diff
@@ -0,0 +1,28 @@
+@bootstrap
+Feature: HA bootstrap process 1
```
arbulu89: We need to add better names to the feature. I guess each feature has a purpose, so this should be described better if possible

data-manifest
```diff
@@ -63,7 +63,14 @@ test/containerized-regression-tests.sh
 test/crm-interface
 test/defaults
 test/descriptions
+test/docker_scripts.sh
 test/evaltest.sh
+test/features/bootstrap_1.feature
```
arbulu89: We need to have better names for the feature files

.travis.yml
```diff
@@ -33,6 +34,24 @@ jobs:
...
+      before_install:
+        - /bin/bash $TRAVIS_BUILD_DIR/test/docker_scripts.sh before_install
```
arbulu89: I guess that you have added this `thread` thing to run the tests in parallel but it doesn't look like really scalable...

test/docker_scripts.sh
liangxin1300: Line `docker exec -t hanode2 /bin/sh -c "cd /app; ./test/run-in-travis.sh build"` is for deploying crmsh on node `hanode2`.
This should be executed on `hanode2` since we have scenarios like this:
```
  Scenario: 2. Node "hanode2" join the cluster
    Given   Cluster service is "stopped" on "hanode2"
```
Means `hanode2` must have crmsh build and installed

liangxin1300:
> Hi @liangxin1300 ,
> I think that this PR needs some improvements. Here the major details:
> 
> 1. The names of the features and files are not meaningfull. Using indexes to create the files is not a good idea in my opinion. Each feature file must be create for a reason and have a better name.

Changed as `bootstrap_init_join_remove.feature` and `bootstrap_options.feature`

> 2. I don't really find any reason for having different feature files in this case as all of them are used to test the same thing, the bootstrap. I would have one feature file for `bootstrap` and other for `hb_report` for example.

Removed hb_report feature file.

> 3. The whole implementation is really done to fit in travis, but we shouldn't do that... The tests must have sense by themself if I want to run them locally for example. So having multiple feature files just to improve the test execution in travis is not a good idea. We need to split this logic. First, create the tests and them we will see how to execute them in travis in optimal sense.

I've changed in `docker_scripts.sh` and `run-in-travis.sh`.
When running on travis, to save time, testing can running in parallel, through specify the feature file suffix as parameter;
When running on locally, in a docker envrionment, run bash file `test/docker_scripts.sh`, then testing can running in sequential order.


> 4. I didn't dig much in the test cases perse, but one thing that I would emphasize is that every test must be independent, this means that running them alone must be the same than running them all together.
> 
> Here my initial thoughts. What do you think?

test/run-in-travis.sh
liangxin1300: Changed like this:
```
bootstrap_tests() {
        SUFFIX="${1:-*}"
        echo "** Bootstrap process tests using python-behave"
        behave --no-logcapture --tags "@bootstrap" --tags "~@wip" /usr/share/crmsh/tests/features/bootstrap_$SUFFIX.feature
}
```

test/docker_scripts.sh
arbulu89: And `hanode1` doesn't require crmsh? This is what got my attention, why both machines are different, when during a normal crmsh installation both must be in the same state?

test/run-in-travis.sh
```diff
@@ -21,8 +21,26 @@ regression_tests() {
...
 
-unit_tests
-rc_unittest=$?
-configure
-make_install
-regression_tests && exit $rc_unittest
+bootstrap_tests() {
+        SUFFIX="${1:-*}"
+	echo "** Bootstrap process tests using python-behave"
+        behave --no-logcapture --tags "@bootstrap" --tags "~@wip" /usr/share/crmsh/tests/features/bootstrap_$SUFFIX.feature
```
arbulu89: I don't still understand why you are adding this `bootstrap_` prefix here hardcoded. Why are you not using the whole name? I don't see any need to have `bootstrap` prefix in the test files when we are adding this as a tag.

test/run-in-travis.sh
```diff
@@ -21,8 +21,26 @@ regression_tests() {
...
+}
+
+case "$1" in
+	build)
+		configure
+		make_install
+		exit $?;;
+	bootstrap)
```
arbulu89: I wouldn't use here this `bootstrap` entry. I would name it `functional_tests` or similar. If in the future we decide to add additional test scenario it will make more sense

test/docker_scripts.sh
```diff
@@ -0,0 +1,41 @@
...
+
+run() {
+  docker exec -t hanode1 /bin/sh -c "cd /app; ./test/run-in-travis.sh bootstrap $1"
+}
+
+case "$1" in
+  before_install)
+    before
+    ;;
+  script)
```
arbulu89: Why `script`? It's not really easy to understand. 
I would suggest something like `functional_tests` or similar.
Let's try to make everything easier for other contributors

test/features/environment.py
```diff
@@ -0,0 +1,25 @@
...
+def before_tag(context, tag):
+    context.logger = logging.getLogger("test.{}".format(tag))
+    # tag @clean means need to stop cluster service
+    if tag == "clean":
+        online_nodes = get_online_nodes(context)
+        if online_nodes:
+            try:
+                parallax.parallax_call(online_nodes, 'crm cluster stop')
+            except ValueError as err:
+                context.logger.error("{}\n".format(err))
```
arbulu89: Quick question. What happens here if we have a failure? Are tests executed?
In this exception we should finish the test execution, right?

test/features/steps/step_bootstrap.py
```diff
@@ -0,0 +1,108 @@
+import re
+import logging
+from behave import given, when, then
+from crmsh import utils, bootstrap, corosync, parallax
+
+
+def run_command(context, cmd):
```
arbulu89: Why now it's not maybe needed, I would prefer to only have `steps` code in `steps` files. I would put this auxiliar methods in some other file.

test/features/steps/step_bootstrap.py
```diff
@@ -0,0 +1,108 @@
...
+    _, out = utils.get_stdout("crm_node -l")
+    for node in nodelist.split():
+        node_info = "{} member".format(node)
+        if not node_info in out:
+            rc = False
+            context.logger.error("Node \"{}\" not online\n".format(node))
+    return rc
+
+
+@given('Cluster service is "{state}" on "{addr}"')
```
arbulu89: I think its quite strange that you put all the variables in quotes. Why is that convention? 
I wouldn't use them, it reduces human readibility. But if you think that it helps to identity variable I don't really care to keep them.

test/features/bootstrap_init_join_remove.feature
```diff
@@ -0,0 +1,26 @@
+@bootstrap
+Feature: HA bootstrap process - init, join and remove
+
+  Scenario: Init cluster service on node "hanode1", and join on node "hanode2"
+    Given   Cluster service is "stopped" on "local"
```
arbulu89: Maybe we should use `hanode1` instead of `local` ?

test/features/bootstrap_options.feature
```diff
@@ -0,0 +1,44 @@
+@bootstrap
+Feature: HA bootstrap process - options
```
arbulu89: I still miss documentation here. What's the meaning of options? A short description after the name would be helpful.
We can add docstring always to improve the test purpose.

test/features/bootstrap_init_join_remove.feature
```diff
@@ -0,0 +1,26 @@
+@bootstrap
+Feature: HA bootstrap process - init, join and remove
```
arbulu89: I still don't like having tests depending in each other.
If the 1st test fails here, the rest will fail, and this doesn't mean that the things being tested in those are incorrect.

I would prefer to have totally independent tests.
Remember that the tests might be executable 1 by 1 and not only in CI

test/run-in-travis.sh
liangxin1300: I've changed it, change function name as `functional_tests`, and for now it accept 3 types feature files(bootstrap, qdevice, hb_report)

test/run-in-travis.sh
liangxin1300: changed

test/docker_scripts.sh
liangxin1300: changed like this:
```
usage() {
  echo "Usage: ./test/`basename $0` <`echo ${TEST_TYPE// /|}`>"
}

# $1 could be "bootstrap", "hb_report", "qdevice" etc.
# $2 could be "before_install" or "run"
# $3 could be suffix of feature file
case "$1/$2" in
  */before_install)
    before $1
    ;;
  */run)
    run $1 $3
    ;;
  *)
    if [ "$#" -eq 0 ] || ! [[ $TEST_TYPE =~ (^|[[:space:]])$1($|[[:space:]]) ]];then
      usage
      exit 1
    fi
    before $1
    run $1
    ;;
esac
```
I think it become more flexible

test/features/environment.py
liangxin1300: changed, set `context.failed = True` maybe work

test/features/steps/step_bootstrap.py
liangxin1300: variables without quotes like `'Cluster service is {state} on {addr}'` will case error in behave

test/features/steps/step_bootstrap.py
liangxin1300: changed
I put them into utils.py file

test/features/bootstrap_init_join_remove.feature
liangxin1300: changed

test/features/bootstrap_init_join_remove.feature
liangxin1300: Changed, every scenario with a `@clean` tag to make sure it have no dependency

test/features/bootstrap_options.feature
liangxin1300: changed

test/features/steps/step_bootstrap.py
arbulu89: mmm, when I used behave in the past it wasn't this constraint...
Anyway, we can alwayas have a look on that in the future if we need

PR #485, opened 2019-11-14
[crmsh-3.0]Fix: bootstrap: get default ip correctly when cluster join(bsc#1156319)

liangxin1300:
#### Problem
First node configured as unicast, when second node joining, missed the default IP,
got `ERROR: cluster.join: No value for ring0`

#### Solution
Using `pick_default_value` function to provide default value for function `prompt_for_string`

#### Additional
* Make unit test for function `pick_default_value`
* Add `-v` option on `nose` to make unit test results more verbose

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -225,6 +225,21 @@ def wait_for_cluster():
     status_done()
 
 
+def pick_default_value(default_list, prev_list):
+    """
+    Provide default value for function 'prompt_for_string'.
+    Make sure give different default value in multi-ring mode.
+
+    Parameters:
+    * default_list - default value list for config item
+    * prev_list    - previous value for config item in multi-ring mode
+    """
+    for value in default_list:
+        if value not in prev_list:
+            return value
+    return ""
+
+
 def start_service(service):
     """
     Start and enable systemd service
@@ -832,11 +847,6 @@ def valid_port(port, prev_value=None):
 
 
 def init_corosync_unicast():
-    def pick_default_value(vlist, ilist):
-        # give a different value for second config items
-        if len(ilist) == 1:
-            vlist.remove(ilist[0])
-        return vlist[0]
 
     if _context.yes_to_all:
         status("Configuring corosync (unicast)")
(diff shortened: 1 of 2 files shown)


liangxin1300:
@arbulu89 @diegoakechi , please review it, thanks!

test/run
```diff
@@ -1,13 +1,13 @@
 #!/bin/sh
 case `pwd` in
 	*/test/unittests)
-		PYTHONPATH=../.. nosetests -w . "$@"
+		PYTHONPATH=../.. nosetests -v -w . "$@"
```
arbulu89: No I notices that we are still using `nose`... And it's deprecated.
We should move to `pytest` in the future. What do you think? It should be super easy

crmsh/bootstrap.py
```diff
@@ -856,27 +866,18 @@ def pick_default_value(vlist, ilist):
...
-            default_networks = [_context.ip_network]
-    if not default_networks:
+    local_iplist = utils.ip_in_local(_context.ipv6)
+    len_iplist = len(local_iplist)
+    if len_iplist == 0:
         error("No network configured at {}!".format(utils.this_node()))
 
+    default_ip = [_context.ip_address] + [ip for ip in local_iplist if ip != _context.ip_address]
+
     for i in 0, 1:
```
arbulu89: Could we change this `0, 1` by `range(2)`? It looks much better for me hehe

crmsh/bootstrap.py
```diff
@@ -856,27 +866,18 @@ def pick_default_value(vlist, ilist):
...
-            default_networks = [_context.ip_network, network_list.remove(_context.ip_network)]
-        else:
-            default_networks = [_context.ip_network]
-    if not default_networks:
+    local_iplist = utils.ip_in_local(_context.ipv6)
+    len_iplist = len(local_iplist)
+    if len_iplist == 0:
         error("No network configured at {}!".format(utils.this_node()))
 
+    default_ip = [_context.ip_address] + [ip for ip in local_iplist if ip != _context.ip_address]
```
arbulu89: I don't understand this... You are avoiding the `_context.ip_address` in the loop to add it later...
If they idea is to remove duplicates better something like: `list(dict.fromkeys(mylist))`

And maybe change `default_ip` by `current_ips` as the variable is a list with the current addresses, default looks like not the meaning of this list

arbulu89: Or simply add `_context.ip_address` if it's not in the list. Otherwise is hard to undestand the meaning of this code line

crmsh/bootstrap.py
```diff
@@ -1646,13 +1642,20 @@ def update_nodeid(nodeid, node=None):
     # if unicast, we need to add our node to $corosync.conf()
     is_unicast = "nodelist" in open(corosync.conf()).read()
     if is_unicast:
+        local_iplist = utils.ip_in_local(_context.ipv6)
+        len_iplist = len(local_iplist)
+        if len_iplist == 0:
+            error("No network configured at {}!".format(utils.this_node()))
+
+        default_ip = [_context.ip_address] + [ip for ip in local_iplist if ip != _context.ip_address]
```
arbulu89: The same than before, I don't get this line purpose

test/run
liangxin1300: OK, I will remove this commit from this PR, and create a new one to master branch to use pytest, and then check whether we could backport to crmsh-3.0 branch

crmsh/bootstrap.py
liangxin1300: OK, I changed all places using `0, 1` to `range(2)`

crmsh/bootstrap.py
liangxin1300: The logic here, is put `_context.ip_address` the **first place** to pick as default value.
`_context.ip_address` came from `ip -o route show`, may have higher priority;
While `local_iplist` came from `ip addr show`, maybe include lots of addresses but not very useful(maybe unreachable, I guess)

arbulu89: Ok, now I understand, It would be nice to put this as a comment in the top of the line (just saying that this operation is used to put in the 1st place)

PR #487, opened 2019-12-11
[crmsh-3.0]Low: bootstrap: don't overwrite ssh key if already exists

liangxin1300:
Back port #476 to crmsh-3.0 branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 ---
 sudo: required
 language: python
+python: 2.7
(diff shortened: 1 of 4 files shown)


liangxin1300:
@arbulu89 Thanks! I will add this warning later

liangxin1300:
@arbulu89 I appended two more commits from #488 , give warning when ssh key doesn't exist while using --no-overwrite-sshkey

PR #489, opened 2019-12-12
Dev: bootstrap: support multi disk sbd configure

liangxin1300:
Both for command line and interactive mode.

For command line:
```
crm cluster init -s /dev/sda1 -s /dev/sdb1
```

For interactive mode:
Prompt string is 
```
Path to storage device (e.g. /dev/disk/by-id/...), or "none", use ";" as separator for multi path
```
use `;` as separator is the same way mentioned in `/etc/sysconfig/sbd`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1362,38 +1362,44 @@ Configure SBD:
             if not confirm("SBD is already configured to use %s - overwrite?" % (configured_dev)):
                 return
 
-        dev = ""
         dev_looks_sane = False
         while not dev_looks_sane:
-            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none"', r'none|\/.*', dev)
+            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none", use ";" as separator for multi path', r'none|\/.*')
             if dev == "none":
                 _context.diskless_sbd = True
                 init_sbd_diskless()
                 return
-            if not is_block_device(dev):
-                print("    That doesn't look like a block device", file=sys.stderr)
-                dev = ""
-            else:
-                warn("All data on {} will be destroyed!".format(dev))
-                if confirm('Are you sure you wish to use this device?'):
-                    dev_looks_sane = True
+            dev_list = dev.strip(';').split(';')
+            for dev_item in dev_list:
+                if not is_block_device(dev_item):
+                    print("    {} doesn't look like a block device".format(dev_item), file=sys.stderr)
+                    dev_looks_sane = False
+                    break
                 else:
-                    dev = ""
-        _context.sbd_device = dev
+                    warn("All data on {} will be destroyed!".format(dev_item))
+                    if confirm('Are you sure you wish to use this device?'):
+                        dev_looks_sane = True
+                    else:
+                        dev_looks_sane = False
+                        break
 
-    if not is_block_device(_context.sbd_device):
-        error("SBD device %s doesn't seem to exist" % (_context.sbd_device))
+        _context.sbd_device = dev_list
+
+    for dev in _context.sbd_device:
+        if not is_block_device(dev):
+            error("SBD device %s doesn't seem to exist" % (dev))
 
     # TODO: need to ensure watchdog is available
     # (actually, should work if watchdog unavailable, it'll just whine in the logs...)
     # TODO: what about timeouts for multipath devices?
     status_long('Initializing SBD...')
-    if not invoke("sbd -d %s create" % (_context.sbd_device)):
-        error("Failed to initialize SBD device %s" % (_context.sbd_device))
+    for dev in _context.sbd_device:
+        if not invoke("sbd -d %s create" % (dev)):
+            error("Failed to initialize SBD device %s" % (dev))
     status_done()
 
     utils.sysconfig_set(SYSCONFIG_SBD,
-                        SBD_DEVICE=_context.sbd_device,
+                        SBD_DEVICE=';'.join(_context.sbd_device),
                         SBD_PACEMAKER="yes",
                         SBD_STARTMODE="always",
                         SBD_DELAY_START="no",
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -243,8 +243,8 @@ Note:
         storage_group = optparse.OptionGroup(parser, "Storage configuration", "Options for configuring shared storage.")
         storage_group.add_option("-p", "--partition-device", dest="shared_device", metavar="DEVICE",
                                  help='Partition this shared storage device (only used in "storage" stage)')
-        storage_group.add_option("-s", "--sbd-device", dest="sbd_device", metavar="DEVICE",
-                                 help="Block device to use for SBD fencing")
+        storage_group.add_option("-s", "--sbd-device", dest="sbd_device", metavar="DEVICE", action="append",
+                                 help="Block device to use for SBD fencing, use \";\" as separator for multi path")
         storage_group.add_option("-o", "--ocfs2-device", dest="ocfs2_device", metavar="DEVICE",
                                  help='Block device to use for OCFS2 (only used in "vgfs" stage)')
         parser.add_option_group(storage_group)


liangxin1300:
@arbulu89 , please review it, thanks!

crmsh/bootstrap.py
```diff
@@ -1362,38 +1362,44 @@ def init_sbd():
             if not confirm("SBD is already configured to use %s - overwrite?" % (configured_dev)):
                 return
 
-        dev = ""
         dev_looks_sane = False
         while not dev_looks_sane:
```
arbulu89: Now I see, what is this `dev_looks_sane` loop used for?
The loop finishes always in the 1st iteration, there is no way that goes to 2nd round.

If this is the case, we could just remove the loop for code simplicity

crmsh/bootstrap.py
```diff
@@ -1362,38 +1362,44 @@ def init_sbd():
...
                 init_sbd_diskless()
                 return
-            if not is_block_device(dev):
-                print("    That doesn't look like a block device", file=sys.stderr)
-                dev = ""
-            else:
-                warn("All data on {} will be destroyed!".format(dev))
-                if confirm('Are you sure you wish to use this device?'):
-                    dev_looks_sane = True
+            dev_list = dev.strip(';').split(';')
```
arbulu89: Why this initial `strip`? To remove the last `;` if it's incorrectly set by the user, right?

crmsh/ui_cluster.py
```diff
@@ -243,7 +243,7 @@ def looks_like_hostnames(lst):
         storage_group = optparse.OptionGroup(parser, "Storage configuration", "Options for configuring shared storage.")
         storage_group.add_option("-p", "--partition-device", dest="shared_device", metavar="DEVICE",
                                  help='Partition this shared storage device (only used in "storage" stage)')
-        storage_group.add_option("-s", "--sbd-device", dest="sbd_device", metavar="DEVICE",
+        storage_group.add_option("-s", "--sbd-device", dest="sbd_device", metavar="DEVICE", action="append",
```
arbulu89: Could you improve the `help` output? Just add `use ";" as separator for multi path` or similar

liangxin1300:
> Besides that, could you add a new simple test in behave with multiple disks if it's possible? It would assure that everything works fine. If we don't have this option yet, maybe some new UT test. The change looks important enough to have some tests

Test sbd in behave for now, seems impossible. Because I have trouble in setup `watchdog` inside container, without `watchdog`, sbd couldn't be setup.
UT test for `init_sbd` function? That is another huge function in bootstrap:( I don't think it can be tested until we split it into smaller ones

crmsh/bootstrap.py
```diff
@@ -1362,38 +1362,44 @@ def init_sbd():
...
+            error("SBD device %s doesn't seem to exist" % (dev))
 
     # TODO: need to ensure watchdog is available
     # (actually, should work if watchdog unavailable, it'll just whine in the logs...)
     # TODO: what about timeouts for multipath devices?
     status_long('Initializing SBD...')
-    if not invoke("sbd -d %s create" % (_context.sbd_device)):
-        error("Failed to initialize SBD device %s" % (_context.sbd_device))
+    for dev in _context.sbd_device:
+        if not invoke("sbd -d %s create" % (dev)):
```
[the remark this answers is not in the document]
liangxin1300: From the debug view, I would prefer to keep it in loop

crmsh/bootstrap.py
```diff
@@ -1362,38 +1362,44 @@ def init_sbd():
             if not confirm("SBD is already configured to use %s - overwrite?" % (configured_dev)):
                 return
 
-        dev = ""
         dev_looks_sane = False
         while not dev_looks_sane:
-            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none"', r'none|\/.*', dev)
+            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none", use ";" as separator for multi path', r'none|\/.*')
```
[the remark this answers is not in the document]
liangxin1300: I think `path` here means disk path, `path` is also mentioned in `/etc/sysconfig/sbd`
so yes, multi path here means multi disk

crmsh/bootstrap.py
liangxin1300: `dev_looks_sane` is useful.
When 
1. input dev is not a block device, 
2. user answer `no` for `Are you sure you wish to use this device`

`dev_looks_sane` is False, then prompt the string to ask user input again.

crmsh/bootstrap.py
liangxin1300: Yes, just for remove ';' like `/dev/sda1;`

crmsh/ui_cluster.py
liangxin1300: OK, changed

crmsh/bootstrap.py
arbulu89: Ok, perfect! Now I get it! Thanks

crmsh/bootstrap.py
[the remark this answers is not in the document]
arbulu89: Ok, I would put `use ";" as separator for multiple disks` but if matches with other configuration files nomenclature is OK

crmsh/bootstrap.py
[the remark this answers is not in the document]
arbulu89: No problem

gao-yan:
Quite sensible logic. Nice work, thanks!

The only confusion is,  the description of this pull request says it accepts:
`crm cluster init -s /dev/sda1 -s /dev/sdb1`

But according to the code and the help text for `-s `,  it's:
`use \";\" as separator for multi path `

arbulu89:
> Quite sensible logic. Nice work, thanks!
> 
> The only confusion is, the description of this pull request says it accepts:
> `crm cluster init -s /dev/sda1 -s /dev/sdb1`
> 
> But according to the code and the help text for `-s `, it's:
> `use \";\" as separator for multi path `

Thanks @gao-yan , you are right. Actually it's my fault as I proposed to improve the help docstring, but without considering the the `;` usage only applies for devices added during the manual process (without `-y`). So, it should only say that these parameter can be used as many times as used devices (and not mention this separator thing).

gao-yan:
> > Quite sensible logic. Nice work, thanks!
> > The only confusion is, the description of this pull request says it accepts:
> > `crm cluster init -s /dev/sda1 -s /dev/sdb1`
> > But according to the code and the help text for `-s `, it's:
> > `use \";\" as separator for multi path `
> 
> Thanks @gao-yan , you are right. Actually it's my fault as I proposed to improve the help docstring, but without considering the the `;` usage only applies for devices added during the manual process (without `-y`). 

I'm actually also hesitant about which is better, multiple `-s`, or one `-s` supplied with `;` separated string to be consistent with the interactive mode ... I guess it depends on if there's any convention we'd like crmsh to keep/follow. For example what about "-N/--nodes" option?

> So, it should only say that these parameter can be used as many times as used devices (and not mention this separator thing).

BTW sbd supports up to 3 devices :-)

liangxin1300:
@gao-yan , @arbulu89 
Maybe in option mode, it's convenient that crmsh support both:
1. -s /dev/sda1 -s /dev/sdb1
2. -s "/dev/sda1;/dev/sdb1"

> BTW sbd supports up to 3 devices :-)
And crmsh should also make this limit, check like `num_devices <= 3`.

What do you think？

arbulu89:
@liangxin1300 @gao-yan What do you mean with `sbd supports up to 3 devices`?
That with the `sbd` command we can just configure 3 devices or that we can only use 3 devices maximum with `sbd` ?

gao-yan:
> @gao-yan , @arbulu89
> Maybe in option mode, it's convenient that crmsh support both:
> 
> 1. -s /dev/sda1 -s /dev/sdb1
> 2. -s "/dev/sda1;/dev/sdb1"

Even better :-) Still I'd say there probably should be only one that we recommend in the help/doc to prevent confusions, and the other one as a hidden friendly usability tolerance.

> 
> > BTW sbd supports up to 3 devices :-)
> 
> And crmsh should also make this limit, check like `num_devices <= 3`.
> 
> What do you think？

Makes sense.

gao-yan:
> @liangxin1300 @gao-yan What do you mean with `sbd supports up to 3 devices`?
> That with the `sbd` command we can just configure 3 devices or that we can only use 3 devices maximum with `sbd` ?

There's this strict upper limitation for sbd to work:

http://docserv.nue.suse.com/documents/SLE-HA_15_SP2/SLE-HA-guide/single-html/#sec-ha-storage-protect-fencing-number

PR #492, opened 2019-12-22
Improve multi sbd case usability

liangxin1300:
This PR is improvement for #489 
Changes include:
* Support both '-s device1 -s device2' and '-s 'device1;device2' for command line option
* Improve the logic of code
* Maximum number of SBD device is 3

BTW, this PR need to backport to branch crmsh-3.0 and crmsh-4.1

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1321,10 +1321,27 @@ def init_sbd():
     SBD can also run in diskless mode if no device
     is configured.
     """
-    if not _context.sbd_device and _context.diskless_sbd:
+    def get_dev_list(dev_list):
+        result_list = []
+        for dev in dev_list:
+            if ';' in dev:
+                result_list.extend(dev.strip(';').split(';'))
+            else:
+                result_list.append(dev)
+        return result_list
+
+    # non-interactive case
+    if _context.sbd_device:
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -1321,10 +1321,29 @@ def init_sbd():
...
+        return result_list
+
+    # non-interactive case
+    if _context.sbd_device:
+        _context.sbd_device = get_dev_list(_context.sbd_device)
+        if len(_context.sbd_device) > 3:
+            error("Maximum number of SBD device is 3")
+        for dev in _context.sbd_device:
+            if not is_block_device(dev):
+                error("SBD device %s doesn't seem to exist" % (dev))
```
arbulu89: I would put the same error message used in the interactive code (or update the other if you prefect this message).
`    {} doesn't look like a block device".format(dev_item)`

liangxin1300:
> @liangxin1300 Good change.
> Anyway, I think we should improve the help description of the `-s` option to say that `-s` can be used up to 3 times.
> Maybe:
> `Block device to use for SBD fencing, use \";\" as separator or -s multiple times for multi path (up to 3 devices)`

changed, thanks!

PR #496, opened 2020-01-01
dev: unittest: fix test_report.py failing in 2020

vvidic:
Use current year instead of assuming 2019.


liangxin1300:
@vvidic , Thanks for this PR! Nice job!

PR #497, opened 2020-01-03
Fix: bootstrap: set placement-strategy value as "default"(bsc#1129462)

liangxin1300:
According to [pacemaker explained](https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/2.0/html/Pacemaker_Explained/s-cluster-options.html), set the value of cluster option "placement-strategy" as "default"

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1423,7 +1423,7 @@ def init_cluster():
     status("Loading initial cluster configuration")
 
     crm_configure_load("update", """
-property cib-bootstrap-options: stonith-enabled=false placement-strategy=balanced
+property cib-bootstrap-options: stonith-enabled=false
 op_defaults op-options: timeout=600 record-pending=true
 rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,12 @@ jobs:
       script:
         - $FUNCTIONAL_TEST bootstrap run options
 
+    - name: "regression test for bootstrap bugs"
+      before_install:
+        - $FUNCTIONAL_TEST bootstrap before_install
+      script:
+        - $FUNCTIONAL_TEST bootstrap run bugs
+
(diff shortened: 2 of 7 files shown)


liangxin1300:
@gao-yan , Please review.

BTW, according to [expained for pacemaker 1.1](https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/s-cluster-options.html), I'm afraid this PR should also backport to both crmsh-4.1 and crmsh-3.0,

zzhou1:
You might need modify test/unittests/test_bugs.py too, and delete "placement-strategy=balanced" accordingly?

crmsh/bootstrap.py
```diff
@@ -1423,7 +1423,7 @@ def init_cluster():
     status("Loading initial cluster configuration")
 
     crm_configure_load("update", """
-property cib-bootstrap-options: stonith-enabled=false placement-strategy=balanced
+property cib-bootstrap-options: stonith-enabled=false placement-strategy=default
```
gao-yan: Thanks for the change, @liangxin1300! How about we default it just by dropping the placement-strategy setting from here so that the configuration is simplified?

liangxin1300: OK, changed!

liangxin1300:
> You might need modify test/unittests/test_bugs.py too, and delete "placement-strategy=balanced" accordingly?

@zzhou1 I prefer to test this result in behave: https://travis-ci.org/ClusterLabs/crmsh/jobs/632213499#L1086

zzhou1:
> > You might need modify test/unittests/test_bugs.py too, and delete "placement-strategy=balanced" accordingly?
> 
> @zzhou1 I prefer to test this result in behave: https://travis-ci.org/ClusterLabs/crmsh/jobs/632213499#L1086

Sounds good. 

However, my point is to clean it up in the test_bugs.py too. Otherwise, it is a bit messy been left behind.

liangxin1300:
@gao-yan Please review and approve this, thanks!:)

PR #500, opened 2020-01-06
[crmsh-3.0] Fix: bootstrap: set placement-strategy value as "default"(bsc#1129462)

liangxin1300:
Backport from #497


liangxin1300:
@gao-yan Please review and approve this, thanks!:)

PR #503, opened 2020-01-07
[crmsh-3.0] Dev: bootstrap: avoid endless loop for noninteractive when join

liangxin1300:
In non-interactive mode, when peer IP is not in the same network with the join node,
will lead to endless looping

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1673,8 +1673,12 @@ def join_cluster(seed_host):
                 peer_ip = corosync.get_value("nodelist.node.ring{}_addr".format(i))
                 # e.g. peer ring0_addr and local ring0_addr must int the same network
                 if peer_ip not in utils.Network(tmp):
-                    print(term.render(clidisplay.error("    shouldn't in network {}".format(tmp))))
-                    continue
+                    errmsg = "Peer IP {} is not in the same network: {}".format(peer_ip, tmp)
+                    if _context.yes_to_all:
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -1673,8 +1673,12 @@ def update_nodeid(nodeid, node=None):
                 peer_ip = corosync.get_value("nodelist.node.ring{}_addr".format(i))
                 # e.g. peer ring0_addr and local ring0_addr must int the same network
                 if peer_ip not in utils.Network(tmp):
-                    print(term.render(clidisplay.error("    shouldn't in network {}".format(tmp))))
-                    continue
+                    errmsg = "Peer IP {} is not in the same network: {}".format(peer_ip, tmp)
+                    if _context.yes_to_all:
+                        error(errmsg)
+                    else:
+                        print(term.render(clidisplay.error("    {}".format(errmsg))))
```
arbulu89: As a curiosity, what happens when this error is raised during the interactive usage? I mean, the code execution continues, but I guess eventually something will fail, right? Why not stop here the execution raising the error (and finishing the method)? I don't know if the code coming next should be executed or not. I guess that is not harmful anyway...

liangxin1300: @arbulu89 , on interactive mode, `continue` here means if this input IP is not in the same network with the `peer IP`, error message prompt, and ask user to input this IP again, until input is the right one

PR #508, opened 2020-01-10
Fix: Not allowed space value for option(bsc#1141976)

liangxin1300:
space value like 
```
crm cluster init -i "  "
crm -c " "
crm cluster init --name " "
```
is harmful, should raise `ValueError`

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -15,6 +15,7 @@ import bz2
 import fnmatch
 import gc
 import ipaddress
+import argparse
 from contextlib import contextmanager, closing
 from . import config
 from . import userdir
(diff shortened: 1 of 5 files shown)


crmsh/utils.py
```diff
@@ -2264,4 +2265,13 @@ def get_nodeid_from_name(name):
         return res.group(1)
     else:
         return None
+
+
+def check_space_option_value(options):
+    assert isinstance(options, argparse.Namespace) is True
```
arbulu89: Using `assert` in the code that is not a test is not recommend. The `assert` method is kind of special, as depending on how python is executed this methods are avoided. I would change that for a normal execption

crmsh/utils.py
```diff
@@ -2264,4 +2265,13 @@ def get_nodeid_from_name(name):
...
+
+    for opt in vars(options):
+        value = getattr(options, opt)
+        if isinstance(value, str) and len(value.strip()) == 0:
+            raise ValueError("Not allowed space value for dest \"{}\"".format(opt))
```
arbulu89: I would change to `Space value not allowed for dest \"{}\""`

Issue #474, opened 2019-09-04
Question on modifying an existing resource op monitor parameter


handruin:
Hi,

I've been reading through the doc trying to figure out if it's possible to script a change to a primitive monitor definition using the crmsh utility.  I need to edit a bunch of clusters with stonith primitives to specifically add a new key/value of **on-fail=ignore** to the **op monitor** section.  I cannot find the right command line option to add this new value specifically to the op monitor section.  Does the crmsh utility support modifying this op monitor section of a resource primitive via a single command line operation?  I know I can use the command `crm config edit` and change it manually but I need to script this.

Example primitive definition:
> primitive stonith-fence0 stonith:fence_ipmilan
params pcmk_host_list=dcdev0 ipaddr=xxx.xxx.xxx.xxx login=stonith passwd="xxxxxxxxx" lanplus=1 pcmk_monitor_timeout=3600s pcmk_monitor_retries=500
> op monitor interval=60s **on-fail=ignore**

liangxin1300:
Hi @handruin , thanks for using crmsh and thanks for your suggestions!
I've created a PR for this, see #514 
with this new PR, in above case, you could do it like:
```
crm configure set stonith-fence0.monitor.on-fail ignore
```

PR #514, opened 2020-01-19
Feature: configure: make configure.set to update operation

liangxin1300:
#### Problem
Like issue #474 mentioned, it's not convenient for modifying an existing resource op monitor parameter.
Using `edit` can modify it but it's not work if user want to script this process

#### Solution
This PR extend the functionalities of configure.set subcommand, while try to keep compatible, if we want to update/add op parameters for en existing resource, we can do it like this:
```
set vip1.monitor.on-fail ignore
set drbd.monitor.10s.interval 20s
```
usage for configure.set
```
set <path> <value>

path:: id.[op_type.][interval.]name
```

#### Test
Behave test file is [here](https://github.com/ClusterLabs/crmsh/blob/38047ca78b46164b9f65b8d448cc66e15971b972/test/features/resource_set.feature)
include validation and normal using for configure.set

### DIFF
--- a/crmsh/ui_configure.py
+++ b/crmsh/ui_configure.py
@@ -592,24 +592,50 @@ class CibConfig(command.UI):
     @command.skill_level('administrator')
     @command.completers(_id_list)
     def do_set(self, context, path, value):
-        "usage: set <path> <value>"
-        def split_path():
-            for oid in cib_factory.id_list():
-                if path.startswith(oid + "."):
-                    return oid, path[len(oid)+1:]
-            context.fatal_error("Invalid path: " + path)
-        obj_id, obj_attr = split_path()
+        """
+        usage: set <path> <value>
+
+        path:: id.[op_type.][interval.]name
+        """
+        path_errmsg = "Invalid path: \"{}\"; Valid path: \"id.[op_type.][interval.]name\"".format(path)
+        path_list = path.split('.')
+        if len(path_list) < 2 or len(path_list) > 4:
+            context.fatal_error(path_errmsg)
+
+        obj_id, *other_path_list = path_list
         rsc = cib_factory.find_object(obj_id)
         if not rsc:
-            context.fatal_error("Resource %s not found" % (obj_id))
-        nvpairs = rsc.node.xpath(".//nvpair[@name='%s']" % (obj_attr))
-        if not nvpairs:
-            context.fatal_error("Attribute not found: %s" % (path))
-        if len(nvpairs) != 1:
-            context.fatal_error("Expected 1 attribute named %s, found %s" %
-                                (obj_attr, len(nvpairs)))
+            context.fatal_error("Object {} not found".format(obj_id))
+
+        # Use case for: set id.name value
+        if len(other_path_list) == 1:
+            obj_attr = other_path_list[0]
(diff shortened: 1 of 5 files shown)


crmsh/ui_configure.py
```diff
@@ -592,24 +592,47 @@ def do_filter(self, context, filterprog, *args):
...
-            context.fatal_error("Resource %s not found" % (obj_id))
-        nvpairs = rsc.node.xpath(".//nvpair[@name='%s']" % (obj_attr))
-        if not nvpairs:
-            context.fatal_error("Attribute not found: %s" % (path))
-        if len(nvpairs) != 1:
-            context.fatal_error("Expected 1 attribute named %s, found %s" %
-                                (obj_attr, len(nvpairs)))
+            context.fatal_error("Object {} not found".format(obj_id))
+
+        if len(other_path_list) == 1:
```
arbulu89: Could we put some comment in top of the if to show which is the use case. Like `# id.name value use case`

crmsh/ui_configure.py
```diff
@@ -592,24 +592,47 @@ def do_filter(self, context, filterprog, *args):
...
+            obj_attr = other_path_list[0]
+            nvpairs = rsc.node.xpath(".//nvpair[@name='{}']".format(obj_attr))
+            if not nvpairs:
+                context.fatal_error("Attribute not found: {}".format(path))
+            if len(nvpairs) != 1:
+                context.fatal_error("Expected 1 attribute named {}, found {}".format(obj_attr, len(nvpairs)))
+            nvpairs[0].set("value", value)
+
+        if len(other_path_list) == 2:
```
arbulu89: The same about the comment. `# id.op_type.name value use case`

crmsh/ui_configure.py
```diff
@@ -592,24 +592,47 @@ def do_filter(self, context, filterprog, *args):
...
+                context.fatal_error("Operation \"{}\" not found for resource {}".format(op_type, obj_id))
+            if len(op_res) > 1:
+                context.fatal_error("Should specify interval of {}".format(op_type))
+            op_res[0].set(name, value)
+
+        if len(other_path_list) == 3:
+            op_type, iv, name = other_path_list
+            op_res = rsc.node.xpath(".//operations/op[@id='{}-{}-{}']".format(obj_id, op_type, iv))
+            if not op_res:
+                context.fatal_error("Operation \"{}\" interval \"{}\" not found for resource {}".format(op_type, iv, obj_id))
```
arbulu89: Don't we have here the case where the `len(op_res) > 1` like in the others?

crmsh/ui_configure.py
liangxin1300: Changed, thanks!

crmsh/ui_configure.py
liangxin1300: Changed, thanks!

crmsh/ui_configure.py
liangxin1300: `len(op_res) > 1` means these Ids are exactly same, that is impossible for cib, it's invalid

liangxin1300:
Thanks @arbulu89! 
I've also updated backport PR #515 
Please approve both of them

PR #516, opened 2020-01-21
For udpu, don't check join node's ip was in the same network

liangxin1300:
## Problem
While in node join process in udpu mode, we do have restriction that forcing the join node address being the same network with the init node's. Corosync doesn't have such limit, so we should loose this checking.

But, loosing this check may have side effect: for normal/non-cloud environment, corosync might be configured with **crossed** link, in which scenario each node can not see each other after joining.

Take an example for **crossed** link:
```
node1's interface: eth1, eth2
node2's interface: eth1, eth2
node1: use eth1 for corosync
node2: use eth2 for corosync
```

So we should have to check the cluster status during joining. 
**If join node itself is online while can't see init node**, bootstrap on join node will:
1. recover original corosync.conf and sync it to init node
2. stop corosync.service on join node
3. exit with an error message 

On join node, it will successfully joined in cluster until using the correct link

## Unit test
a170cb839cce7fb5539e639603058e35d6b57389

## Functional test
b42cbe8eb6bb5070362e3b1fb9dec786fef7e548

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -18,6 +18,7 @@ import random
 import re
 import time
 import readline
+import shutil
 from string import Template
 from lxml import etree
 from . import config
@@ -38,6 +39,7 @@ SYSCONFIG_SBD = "/etc/sysconfig/sbd"
 SYSCONFIG_FW = "/etc/sysconfig/SuSEfirewall2"
 SYSCONFIG_FW_CLUSTER = "/etc/sysconfig/SuSEfirewall2.d/services/cluster"
 PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
+COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 
 
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
@@ -228,11 +230,51 @@ def wait_for_cluster():
     status_long("Waiting for cluster")
     while True:
         _rc, out, _err = utils.get_stdout_stderr("crm_mon -1")
-        if "online" in out.lower():
+        if is_online(out):
             break
         status_progress()
-        sleep(5)
+        sleep(2)
     status_done()
+
+
+def get_cluster_node_hostname():
+    """
+    Get the hostname of the cluster node used during the join process if an IP address is used.
+    """
+    peer_node = None
+    if _context.cluster_node:
+        if utils.valid_ip_addr(_context.cluster_node):
+            rc, out, err = utils.get_stdout_stderr("ssh {} crm_node --name".format(_context.cluster_node))
+            if rc != 0:
+                error(err)
+            peer_node = out
+        else:
+            peer_node = _context.cluster_node
+    return peer_node
+
+
+def is_online(crm_mon_txt):
+    """
+    Check whether local node is online
+    Besides that, in join process, check whether init node is online
+    """
+    if not re.search("Online: .* {} ".format(utils.this_node()), crm_mon_txt):
+        return False
+
+    # if peer_node is None, this is in the init process
+    peer_node = get_cluster_node_hostname()
+    if peer_node is None:
+        return True
+    # In join process
+    # If the joining node is already online but can't find the init node
+    # The communication IP maybe mis-configured
+    if not re.search("Online: .* {} ".format(peer_node), crm_mon_txt):
+        shutil.copy(COROSYNC_CONF_ORIG, corosync.conf())
+        csync2_update(corosync.conf())
+        stop_service("corosync")
+        print()
+        error("Cannot see peer node \"{}\", please check the communication IP".format(peer_node))
+    return True
 
 
 def pick_default_value(default_list, prev_list):
@@ -1845,6 +1887,8 @@ def join_cluster(seed_host):
             invoke("crm cluster run '{}' {}".format(cmd, node))
         else:
             corosync.set_value("totem.nodeid", nodeid)
+
+    shutil.copy(corosync.conf(), COROSYNC_CONF_ORIG)
 
     # check if use IPv6
     ipv6_flag = False
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1909,21 +1909,6 @@ def join_cluster(seed_host):
                                               ringXaddr_res)
                 if not ringXaddr:
                     error("No value for ring{}".format(i))
-
-                # this check does not work on GCP (bsc#1106946)
-                if utils.detect_cloud() != "google-cloud-platform":
-                    _, outp = utils.get_stdout("ip addr show")
-                    tmp = re.findall(r' {}/[0-9]+ '.format(ringXaddr), outp, re.M)[0].strip()
-                    peer_ip = corosync.get_value("nodelist.node.ring{}_addr".format(i))
-                    # peer ring0_addr and local ring0_addr must be configured in the same network
-                    if not utils.ip_in_network(peer_ip, tmp):
-                        errmsg = "    Peer IP {} is not in the same network: {}".format(peer_ip, tmp)
-                        if _context.yes_to_all:
-                            error(errmsg)
-                        else:
-                            print(term.render(clidisplay.error(errmsg)))
-                            continue
-
                 ringXaddr_res.append(ringXaddr)
                 break
             if not rrp_flag:
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -811,24 +811,47 @@ class IPAlreadyConfiguredError(Exception):
     pass
 
 
-def add_node_ucast(IParray, node_id=None):
+def find_configured_ip(ip_list):
+    """
+    find if the same IP already configured
+    If so, raise IPAlreadyConfiguredError
+    """
+    with open(conf()) as f:
+        p = Parser(f.read())
+
+    # get exist ip list from corosync.conf
+    corosync_iplist = []
+    for path in set(p.all_paths()):
+        if re.search('nodelist.node.ring[0-9]*_addr', path):
+            corosync_iplist.extend(p.get_all(path))
+
+    # all_possible_ip is a ip set to check whether one of them already configured
+    all_possible_ip = set(ip_list)
+    # get local ip list
+    local_ip_list = utils.ip_in_local(utils.is_ipv6(ip_list[0]))
+    # extend all_possible_ip if ip_list contain local ip
+    # to avoid this scenarios in join node:
+    #   eth0's ip already configured in corosync.conf
+    #   eth1's ip also want to add in nodelist
+    # if this scenarios happened, raise IPAlreadyConfiguredError
+    if bool(set(ip_list) & set(local_ip_list)):
+        all_possible_ip |= set(local_ip_list)
+    configured_ip = list(all_possible_ip & set(corosync_iplist))
+    if configured_ip:
+        raise IPAlreadyConfiguredError("IP {} was already configured".format(','.join(configured_ip)))
+
+
+def add_node_ucast(ip_list, node_id=None):
+
+    find_configured_ip(ip_list)
 
     f = open(conf()).read()
     p = Parser(f)
 
-    # to check if the same IP already configured
-    exist_iplist = []
-    for path in p.all_paths():
-        if re.search('nodelist.node.ring[0-9]*_addr', path):
-            exist_iplist.extend(p.get_all(path))
-    for ip in IParray:
-        if ip in exist_iplist:
-            raise IPAlreadyConfiguredError("IP {} was already configured".format(ip))
-
     if node_id is None:
         node_id = get_free_nodeid(p)
     node_value = []
-    for i, addr in enumerate(IParray):
+    for i, addr in enumerate(ip_list):
         node_value += make_value('nodelist.node.ring{}_addr'.format(i), addr)
     node_value += make_value('nodelist.node.nodeid', str(node_id))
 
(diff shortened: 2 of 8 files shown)


crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
 
+def get_peer_hostname():
+    """
+    In join process, cluster_node might be an IP or hostname
+    Return hostname anyway
+    """
+    peer_node = None
+    if _context.cluster_node:
+        if utils.valid_ip_addr(_context.cluster_node):
+            peer_node = utils.get_stdout("ssh {} hostname".format(_context.cluster_node))[1]
```
arbulu89: What would happen here in this scenario:
- The IP address is valid, but it's not available for SSH or it doesn't even exist.

crmsh/bootstrap.py
```diff
@@ -1846,6 +1885,9 @@ def update_nodeid(nodeid, node=None):
         else:
             corosync.set_value("totem.nodeid", nodeid)
 
+    atexit.register(os.remove, COROSYNC_CONF_ORIG)
```
arbulu89: What this `register` is doing?

crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
+
+def is_online(crm_mon_txt):
+    """
+    Check whether local node is online
+    Besides that, in join process, check whether init node is online
+    """
+    if not re.search("Online: .* {} ".format(utils.this_node()), crm_mon_txt):
+        return False
+
+    peer_node = get_peer_hostname()
```
arbulu89: I would directly return True here if peer_node is False. It would make the code easier to read.
And we could add comments.
If peer_node is False, it means that this is a `init` process, right? I would add this comments there

crmsh/bootstrap.py
```diff
@@ -1909,21 +1951,6 @@ def update_nodeid(nodeid, node=None):
                                               ringXaddr_res)
                 if not ringXaddr:
                     error("No value for ring{}".format(i))
-
```
arbulu89: Is this code executed if unicast is set? I cannot really understand the whole process.
I'm just afraid if this can affect to multicast

arbulu89: In fact, why do we have things related to multicast in that method?

crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
+            peer_node = _context.cluster_node
+    return peer_node
+
+
+def is_online(crm_mon_txt):
+    """
+    Check whether local node is online
+    Besides that, in join process, check whether init node is online
+    """
+    if not re.search("Online: .* {} ".format(utils.this_node()), crm_mon_txt):
```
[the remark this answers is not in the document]
liangxin1300: `is_online` is called by `wait_for_cluster`, which is just after running `start_service("pacemaker.service")`, while `start_service` will check whether `pacemaker.service` is successfully started, if not, the whole program will return with an error.

So, generally, the `this_node` will always online, just wait for some seconds.

crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
+        return False
+
+    peer_node = get_peer_hostname()
+    # In join process
+    # If the join node has already online but can't see the init node
```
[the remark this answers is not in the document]

crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
+    # The communicate IP maybe mis-configured
+    if peer_node and not re.search("Online: .* {} ".format(peer_node), crm_mon_txt):
+        shutil.copy(COROSYNC_CONF_ORIG, corosync.conf())
+        csync2_update(corosync.conf())
+        stop_service("corosync")
+        print()
+        error("Can not see peer node \"{}\", please check the communicate IP".format(peer_node))
```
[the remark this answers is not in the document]
liangxin1300: changed

crmsh/bootstrap.py
```diff
@@ -228,13 +231,49 @@ def wait_for_cluster():
...
-        if "online" in out.lower():
+        if is_online(out):
             break
         status_progress()
-        sleep(5)
+        sleep(2)
     status_done()
 
 
+def get_peer_hostname():
```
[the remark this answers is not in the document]
liangxin1300: changed

crmsh/bootstrap.py
liangxin1300: run `os.remove(COROSYNC_CONF_ORIG)` when the program terminated

crmsh/bootstrap.py
liangxin1300: Changed, use `utils.get_stdout_stderr` instead, and check the return code

crmsh/bootstrap.py

crmsh/bootstrap.py
liangxin1300: This block of code is under flag `is_unicast`, if this flag is Flase(which means is in multicast mode), this code wouldn't be executed.
So this block of code has no related with multicast

liangxin1300:
> Hi @liangxin1300 ,
> I cannot really say if the changes are good or not yet as I don't understand many things.
> Just to clarify some doubts:
> When unicast is used, do all the nodes need to use the same network interface?
> Is this PR affecting somehow to multicast usage?

@arbulu89 , thanks for your suggestions, I have changed and updated.

Besides that, I notice that there was also another side effect for loosing this limit, like this scenarios below:
```
on joining node:
eth0's ip already successfully configured in corosync.conf

then run crm cluster stop
then run crm cluster join -c init_node -i eth2 -y
then nodelist has three node sections:
  1. init node's ip
  2. join node eth0's ip
  3. join node eth1's ip
```
Above issue is similar with bsc#1127095,
So I append commit c2919536bb1145cc3e5115e60b923706c4393c46 and update related unittest code to resolve above issue

crmsh/bootstrap.py
arbulu89: Ok, thanks!

crmsh/corosync.py
```diff
@@ -811,20 +811,48 @@ class IPAlreadyConfiguredError(Exception):
     pass
 
 
-def add_node_ucast(IParray, node_id=None):
-
-    f = open(conf()).read()
-    p = Parser(f)
-
-    # to check if the same IP already configured
+def find_configured_ip(IParray):
```
arbulu89: What is this IParray?
Is it a list with ip addresses? If this is a variable, we should change `IParray` if `iparray`. Otherwise, it looks like a constant, and I don't know if this is the case. It would make the code more readble

crmsh/corosync.py
```diff
@@ -811,20 +811,48 @@ class IPAlreadyConfiguredError(Exception):
...
+    """
+    find if the same IP already configured
+    If so, raise IPAlreadyConfiguredError
+    """
+    with open(conf()) as f:
+        p = Parser(f.read())
+
+    # get exist ip list from corosync.conf
     exist_iplist = []
```
arbulu89: I would change this variable names. `exist_iplist` by `corosync_ip_list`

crmsh/corosync.py
```diff
@@ -811,20 +811,48 @@ class IPAlreadyConfiguredError(Exception):
...
+    #   eth1's ip also want to add in nodelist
+    # if this scenarios happened, raise IPAlreadyConfiguredError
     for ip in IParray:
+        if ip in local_ip_list:
+            all_possible_ip = IParray + local_ip_list
+            break
+    else:
+        all_possible_ip = IParray
+
+    for ip in set(all_possible_ip):
```
arbulu89: I think you can change this for loops by (if both are `sets`). This checks if some value is in both lists:
```
if bool(all_possible_ip & exist_iplist):
    raise IPAlreadyConfiguredError("IP {} was already configured".format(ip))
```

crmsh/corosync.py
```diff
@@ -811,20 +811,48 @@ class IPAlreadyConfiguredError(Exception):
...
+    all_possible_ip = []
+    # get local ip list
+    local_ip_list = utils.ip_in_local(utils.is_ipv6(IParray[0]))
+    # extend all_possible_ip if IParray contain local ip
+    # to avoid this scenarios in join node:
+    #   eth0's ip already configured in corosync.conf
+    #   eth1's ip also want to add in nodelist
+    # if this scenarios happened, raise IPAlreadyConfiguredError
     for ip in IParray:
+        if ip in local_ip_list:
```
arbulu89: I don't understand much this piece of code.
Here you check if some local IP is in the `IParray` list and if just 1 is there, you add the whole list.
Why do you add the whole list? Shouldn't you add just the found IP address?

By the way, you can replace this for and if by:
```
if bool(IParray & local_ip_list):
```

crmsh/corosync.py
```diff
@@ -811,20 +811,48 @@ class IPAlreadyConfiguredError(Exception):
...
+            break
+    else:
+        all_possible_ip = IParray
+
+    for ip in set(all_possible_ip):
         if ip in exist_iplist:
             raise IPAlreadyConfiguredError("IP {} was already configured".format(ip))
 
+
+def add_node_ucast(IParray, node_id=None):
```
arbulu89: The same than before, if this is a variable and not a constant, change `IParray` by `iparray`

crmsh/corosync.py
liangxin1300: Changed as `ip_list`

crmsh/corosync.py
liangxin1300: changed

crmsh/corosync.py
liangxin1300: Changed as `ip_list`

crmsh/corosync.py
liangxin1300: Consider this scenario:
Join node eth0's ip `10.10.10.2` already configured in corosync.conf;
then join node eth1's ip `20.20.20.2` want to add in corosync.conf;
Local ip list in join node is `["10.10.10.2", "20.20.20.2"]`,
if we don't extend local ip list to `all_possible_ip`, no eception `IPAlreadyConfiguredError` will raise, `20.20.20.2` will add to corosync.conf

arbulu89: But at the end you are adding all the IP addresses in `ip_list` to `all_possible_ip`, even though only one might be in both lists. For example:
```
ip_list = [1.1.1.1, 2.2.2.2]
local_ip_list = [1.1.1.1, 3.3.3.3, 4.4.4.4]
all_possible_ip = [1.1.1.1, 2.2.2.2, 1.1.1.1, 3.3.3.3, 4.4.4.4]
```
Only `1.1.1.1` matches but you add all of the addresses. Is this what you want? Just to understand better.

If this is what you want, and to make the code easier. I would just to this:
```
all_ip_list = set(ip_list)
if (set(ip_list) & set(local_ip_list)):
    all_ip_list |= set(local_ip_list)
if bool(all_possible_ip & set(corosync_ip_list)):
    raise IPAlreadyConfiguredError("IP {} was already configured".format(ip))
```
This piece of code does:
1. If some value in ip_list is in local_ip list -> all_ip_list = ip_list + local_ip_list, else all_ip_list = ip_list
2. If some value in all_ip_list is in corosync_ip_list, raise the error

liangxin1300: Changed.
Thank you @arbulu89

crmsh/bootstrap.py
```diff
@@ -1846,6 +1889,9 @@ def update_nodeid(nodeid, node=None):
         else:
             corosync.set_value("totem.nodeid", nodeid)
 
+    atexit.register(os.remove, COROSYNC_CONF_ORIG)
+    shutil.copy(corosync.conf(), COROSYNC_CONF_ORIG)
```
gao-yan: Would it be better to use `tempfile.mktemp()` to create the temporary file, just in case this would accidentally override any existing files of users ? AFAICT, crmsh uses tempfile to create temporary things for other purposes as well.

crmsh/bootstrap.py
```diff
@@ -228,13 +231,53 @@ def wait_for_cluster():
...
 
 
+def get_cluster_node_hostname():
+    """
+    Get the hostname of the cluster node used during the join process if an IP address is used.
+    """
+    peer_node = None
+    if _context.cluster_node:
+        if utils.valid_ip_addr(_context.cluster_node):
+            rc, out, err = utils.get_stdout_stderr("ssh {} hostname".format(_context.cluster_node))
```
gao-yan: Since a system might have multiple names corresponding to different IP addresses in /etc/hosts, to make sure we are really getting the `node name` used for the cluster, we'd better run `crm_node --name` on that node but of course while cluster service is running there.

gao-yan: And of course, probably depending on how this function will be potentially used, use `uname --n` as a fallback in case `crm_node --name` fails for example if cluster service is not running there yet?

crmsh/bootstrap.py
liangxin1300: Changed. Use `tmpfiles.create` to create temporary file.
Thanks!

liangxin1300: Already changed for crmsh-4.1 and crmsh-3.0 branche PRs

crmsh/bootstrap.py
liangxin1300: Changed as `crm_node --name`, and also changed for crmsh-4.1 and crmsh-3.0 branche PRs.

I think if the cluster service is really stopped on init node while executing this piece of code, just let `error` to terminate the join process, because in that case, we should stop and check why service on init node stopped, instead of going ahead.

Thanks @gao-yan !

crmsh/corosync.py
gao-yan: Does it mean we don't mind if any IPs from ip_list are not in local_ip_list? We'd just try whatever the user is asking for? Not really a problem since there's fallback.  Just wondering :-)

PR #517, opened 2020-01-28
Dev: bootstrap: binding interface when calling "crm cluster init" internally

liangxin1300:
#### Problem
In some network scenarios, there is no default route configured.
In that environment, when specifying interface while running `crm cluster init -i nic`, running `ssh_remote` and `csync2_remote` stages will fail as `Failed to determine default network interface`, because internally, bootstrap program try to find default network interface through route

#### Solution
Binding interface when calling "crm cluster init" in `ssh_remote` and `csync2_remote` stages can avoid such issue happened

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1652,8 +1652,8 @@ def join_ssh(seed_host):
     # authorized_keys file (again, to help with the case where the
     # user has done manual initial setup without the assistance of
     # ha-cluster-init).
-    if not invoke("ssh root@%s crm cluster init ssh_remote" % (seed_host)):
-        error("Can't invoke crm cluster init ssh_remote on %s" % (seed_host))
+    if not invoke("ssh root@{} crm cluster init -i {} ssh_remote".format(seed_host, _context.nic)):
+        error("Can't invoke crm cluster init -i {} ssh_remote on {}".format(_context.nic, seed_host))
 
 
 def join_csync2(seed_host):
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -1652,8 +1652,8 @@ def join_ssh(seed_host):
     # authorized_keys file (again, to help with the case where the
     # user has done manual initial setup without the assistance of
     # ha-cluster-init).
-    if not invoke("ssh root@%s crm cluster init ssh_remote" % (seed_host)):
-        error("Can't invoke crm cluster init ssh_remote on %s" % (seed_host))
+    if not invoke("ssh root@{} crm cluster init -i {} ssh_remote".format(seed_host, _context.nic)):
```
arbulu89: Is the `_context.nic` variable always available? Otherwise, but will happen with this change?

liangxin1300: @arbulu89 Yes, If we specify nic with `-i` option, `_context.nic` will be that nic; if not specify nic, `utils.network_defaults` will try to give a default one.
In that network environment which was no default route configured, `utils.network_defaults` will return `None`, the whole bootstrap will exit with error `ERROR: cluster.init: Failed to determine default network interface`, unless we specify nic with `-i`

arbulu89: So other question, could we have new errors cases (this error you mention now) in the cases where before it was working?
I just want to know if this change will affect the cases where everything was working fine. 
I guess not right.
I mean, this change will only provide a new option to pass the `nic` parameter to these commands if there is note default interface.

liangxin1300: No affect, we have functional test to make sure everything works fine:)

liangxin1300:
Thanks for review!

PR #522, opened 2020-01-30
Low: utils: update detect_cloud pattern for aws

arbulu89:
I have found that at least the new AWS instances are not detected with the `detect_cloud` method.

This is the current output: 
```
ec2-user@xarbulu-hana01:~> sudo dmidecode -s system-version
4.2.amazon
```

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2130,7 +2130,7 @@ def detect_cloud():
     if not is_program("dmidecode"):
         return None
     rc, system_version = get_stdout("dmidecode -s system-version")
-    if re.search(r"\<.*\.amazon\>", system_version) is not None:
+    if re.search(r".*amazon.*", system_version) is not None:
         return "amazon-web-services"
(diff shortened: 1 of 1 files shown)


liangxin1300:
Hi @arbulu89 , could you please check  whether we also have this problem in gcp and azure?
Because this function `detect_cloud` is for three of cloud providers.
If so, please append related changes to this PR.
Thanks!

arbulu89:
> Hi @arbulu89 , could you please check whether we also have this problem in gcp and azure?
> Because this function `detect_cloud` is for three of cloud providers.
> If so, please append related changes to this PR.
> Thanks!

Hi @liangxin1300 ,
I have tested that in the latest SLE15 images in Azure and gcp and they work fine for them. The only failing one is AWS.

arbulu89:
@liangxin1300 Thanks for merging! I was adding UT for the method hehe. I will create an additional PR for that

PR #532, opened 2020-02-10
Fix: Avoid circular import error when import bootstrap(bsc#1163231)

liangxin1300:
As https://docs.python.org/3.5/whatsnew/3.5.html mentioned,
`Circular imports involving relative imports are now supported.`
So this PR is a backward fix for python3.4

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -11,7 +11,6 @@ import socket
 from . import utils
 from . import tmpfiles
 from . import parallax
-from . import bootstrap
 from .msg import err_buf, common_debug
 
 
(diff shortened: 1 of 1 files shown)


liangxin1300:
@arbulu89 That's a great idea!
Currently crmsh implement its own logging system using `msg.py`, it do will have benefits if replaced by python builtin `logging`.
Take your time, that will not a small PR, although I can not estimate how long it will be.
Maybe first in bootstrap module, then change other modules if we have time later.

Anyway, thanks!

Btw, we could leave this PR, when new logging system finished, I will close it

PR #533, opened 2020-02-24
Dev: testcases: update test output due to pacemaker update

liangxin1300:
Since pacemaker version update on docker image leap:15.2, there are some test cases need to adjust


liangxin1300:
Thanks!

PR #534, opened 2020-03-03
Dev: unittest: change test_add_node_ucast to make whole test reentrant

liangxin1300:
Original test `test_add_node_ucast` function operate real config file makes the test failed on second time, since the config file changed. So I use the-mock-way to make the whole unit test re-entrant.

### DIFF
--- a/test/unittests/test_corosync.py
+++ b/test/unittests/test_corosync.py
@@ -146,28 +146,68 @@ class TestCorosyncParser(unittest.TestCase):
         mock_open_file.assert_called_once_with(mock_conf.return_value)
         mock_isv6.assert_called_once_with("10.10.10.2")
         mock_ip_local.assert_called_once_with(False)
-        mock_search.assert_called_once_with("nodelist.node.ring[0-9]*_addr", "nodelist.node.ring0_addr")
+        # For some reason mock_search.assert_called_once_with does not work
+        mock_search.assert_has_calls([mock.call("nodelist.node.ring[0-9]*_addr", "nodelist.node.ring0_addr")])
 
-    def test_add_node_ucast(self):
-        from crmsh.corosync import add_node_ucast, get_values
+    @mock.patch("crmsh.corosync.make_section")
+    @mock.patch("crmsh.corosync.get_values")
+    @mock.patch("crmsh.corosync.make_value")
+    @mock.patch("crmsh.corosync.get_free_nodeid")
+    @mock.patch("crmsh.corosync.Parser")
+    @mock.patch("builtins.open", create=True)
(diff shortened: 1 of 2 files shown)


test/unittests/test_corosync.py
```diff
@@ -148,26 +148,74 @@ def test_find_configured_ip_exception(self, mock_open_file, mock_conf, mock_pars
...
-            add_node_ucast(['10.10.10.11'])
-        self.assertEqual("IP 10.10.10.11 was already configured", str(err.exception))
-        now_iplist = get_values('nodelist.node.ring0_addr')
-        self.assertEqual(len(exist_iplist) + 1, len(now_iplist))
+    @mock.patch('crmsh.corosync.make_section')
+    @mock.patch('crmsh.corosync.get_values')
+    @mock.patch('crmsh.corosync.make_value')
+    @mock.patch('crmsh.corosync.get_free_nodeid')
+    @mock.patch('crmsh.corosync.Parser')
+    @mock.patch("builtins.open")
```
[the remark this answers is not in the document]
liangxin1300: Yes, is just for python3, in the backport PR on crmsh-3.0, I use `__builtin__.open`

test/unittests/test_corosync.py
```diff
@@ -148,26 +148,68 @@ def test_find_configured_ip_exception(self, mock_open_file, mock_conf, mock_pars
         mock_ip_local.assert_called_once_with(False)
         mock_search.assert_called_once_with("nodelist.node.ring[0-9]*_addr", "nodelist.node.ring0_addr")
 
-    def test_add_node_ucast(self):
-        from crmsh.corosync import add_node_ucast, get_values
+    @mock.patch('crmsh.corosync.find_configured_ip')
+    def test_add_node_ucast_exception(self, mock_find_ip):
```
arbulu89, line 152: I wrote a comment here before, but it looks like I didn't save it...
So...
I think we don't need this unit test. It doesn't test anything relevant. It checks that the exception is re-raised. This will happen always, with any exception, so it's not a real test.
We should only test this, if the `IPAlreadyConfiguredError` was caught using `try` and re-raised later (or any other operation).
Basically, I would just remove this test case

liangxin1300: I suggest leave this test, since at least this test increase the test coverage:)

liangxin1300:
@arbulu89 Changed with adding `with` statement, and synced with other two backport PRs
Thanks!

test/unittests/test_corosync.py
arbulu89: Really? How is that possible? This test code branch doesn't run any different code compared with `test_add_node_ucast`

PR #537, opened 2020-03-05
Fix bug of remove node(bsc#1165644)

liangxin1300:
## Problem
After running `crm cluster remove peer-node`,
* On udpu mode, peer-node's address wouldn't be removed in corosync.conf
* Value of `expected_votes` and `two_node` changed, but not sync to other left cluster nodes
* `bootstrap_remove` function is huge, better to split into smaller one
* `remove_get_hostname` function is complex, I think it's not necessary to let user input once again,
just exit with error message. So I decide to delete this function
## Solution
- `On udpu mode, peer-node's address wouldn't be removed in corosync.conf`
  This is because pass the hostname to `corosync.del_node`, not ip address;
  Add `get_cluster_node_ip` function to make sure we got the right parameter value, that is to say,
  if a hostname configured, pass that hostname, if IP configured, pass that IP
- `Value of expected_votes and two_node changed, but not sync to other left cluster nodes`
  Calling `csync2_update` to sync corosync.conf after all things are done.
## Test
* Functional test
* Unit test

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -81,6 +81,7 @@ class Context(object):
         self.shared_device = None
         self.ocfs2_device = None
         self.cluster_node = None
+        self.cluster_node_ip = None
         self.force = None
         self.arbitrator = None
         self.clusters = None
@@ -96,6 +97,8 @@ class Context(object):
         self.default_ip_list = []
         self.local_ip_list = []
         self.local_network_list = []
+        self.rm_list = [SYSCONFIG_SBD, CSYNC2_CFG, corosync.conf(), CSYNC2_KEY,
+                COROSYNC_AUTH, "/var/lib/heartbeat/crm/*", "/var/lib/pacemaker/cib/*"]
 
     @classmethod
     def set_context(cls, options):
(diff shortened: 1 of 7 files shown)


crmsh/bootstrap.py
```diff
@@ -2159,105 +2160,42 @@ def start_qdevice_on_join_node(seed_host):
...
-
-    # check whether corosync has been started on local node
-    if not utils.service_is_active("corosync.service"):
-        error("Cluster is not active - can't execute removing action")
-
-    remove_get_hostname(seed_host)
-
-
-def remove_get_hostname(seed_host):
+def get_cluster_node_ip():
```
arbulu89, line 2163: This function name (`get_cluster_node_ip`) is pretty strange. You mention `get`, but the method is not used to get anything. It actually sets `_context.cluster_node_ip` value. We should change the method name to avoid confusions. Maybe `set_cluster_node_ip` would be better (just the fist thing coming to my mind)

crmsh/bootstrap.py
```diff
@@ -2159,105 +2160,42 @@ def start_qdevice_on_join_node(seed_host):
...
-    _context.connect_name = seed_host
-    _context.host_status = 0
-
-    # if node is localhost, assume connected and hostname = cluster name
-    if seed_host == utils.this_node():
-        _context.cluster_node = seed_host
-        _context.host_status = 1
+    node = _context.cluster_node
+    addr_list = corosync.get_values('nodelist.node.ring0_addr')
+    if node in addr_list:
```
arbulu89, line 2174: Why don't you just assign node to `cluster_node`? This way you don't need to do more checks in the future code (like `del_target = _context.cluster_node_ip or node`

crmsh/bootstrap.py
```diff
@@ -2466,61 +2400,69 @@ def bootstrap_remove(context):
...
+        rc = utils.ext_cmd_nosudo("ssh{} -o StrictHostKeyChecking=no {} '{}'".format("" if yes_to_all else " -t", othernode, cmd))
+        if rc != 0:
+            error("Failed to remove this node from {}".format(othernode))
+    else:
+        # stop cluster
+        if not stop_service("corosync"):
+            error("Stopping corosync failed")
+        # remove all trace of cluster from this node
+        # delete configuration files from the node to be removed
+        toremove = [SYSCONFIG_SBD, CSYNC2_CFG, corosync.conf(), CSYNC2_KEY,
```
arbulu89, line 2462: Maybe we could put these list as a constant. I have already seen it in other place.

arbulu89 (approved):
@liangxin1300 The code looks good.
I have done some minor suggestions. Feel free to implement them (or not hehe, I'm not sure about all of them).
Besides that, and with the new tests, I would say we are fine to go.

crmsh/bootstrap.py
liangxin1300: OK, Changed this:)

crmsh/bootstrap.py
liangxin1300: OK, changed:)

PR #542, opened 2020-03-15
Resolve csync2 conflicts(bsc#1166684)

liangxin1300:
Based on https://github.com/LINBIT/csync2/blob/master/doc/csync2.adoc#resolving-a-conflict
Csync2 option -f should be used **after** conflicts happen, not **before**

This PR based on the truth that:
We trust each steps change to configure file **on the bootstrap side**, so we use -f option to win all possible conficts

#### Functional test
```
When    Run "echo "# This is test for bsc#1166684" >> /etc/corosync/corosync.conf" on "hanode1"
When    Run "scp /etc/corosync/corosync.conf root@hanode2:/etc/corosync" on "hanode1"
```
These two lines added after two-nodes cluster started.
Means change the corosync.conf manually, so csync2 db not aware the change.
Then next functional test try to setup qdevice on the running cluster.
Command `csync -rxv /etc/corosync/corosync.conf` will failed.
Now use '-f' option to force to win for the conflicts.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -784,9 +784,17 @@ include /etc/pacemaker/authkey;
 
 
 def csync2_update(path):
-    invoke("csync2 -rm %s" % (path))
-    invoke("csync2 -rf %s" % (path))
-    invoke("csync2 -rxv %s" % (path))
+    '''
+    Sync path to all peers
+
+    If there was a conflict, use '-f' to force this side to win
+    '''
+    invoke("csync2 -rm {}".format(path))
+    if invoke("csync2 -rxv {}".format(path)):
+        return
+    invoke("csync2 -rf {}".format(path))
+    if not invoke("csync2 -rxv {}".format(path)):
+        warn("{} was not synced".format(path))
 
 
 def init_csync2_remote():
@@ -1740,7 +1748,7 @@ def join_csync2(seed_host):
     # they haven't gone to all nodes in the cluster, which means a
     # subseqent join of another node can fail its sync of corosync.conf
     # when it updates expected_votes.  Grrr...
-    if not invoke('ssh -o StrictHostKeyChecking=no root@%s "csync2 -mr / ; csync2 -fr / ; csync2 -xv"' % (seed_host)):
+    if not invoke('ssh -o StrictHostKeyChecking=no root@{} "csync2 -rm /; csync2 -rxv || csync2 -rf / && csync2 -rxv"'.format(seed_host)):
         print("")
         warn("csync2 run failed - some files may not be sync'd")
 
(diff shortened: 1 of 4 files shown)


crmsh/bootstrap.py
```diff
@@ -784,9 +784,17 @@ def init_csync2():
...
+    Sync path to all peers
+
+    If there was a conflict, use '-f' to force this side to win
+    '''
+    invoke("csync2 -rm {}".format(path))
+    if invoke("csync2 -rxv {}".format(path)):
+        return
+    invoke("csync2 -rf {}".format(path))
+    if not invoke("csync2 -rxv {}".format(path)):
+        warn("{} was not sync".format(path))
```
arbulu89: `was not synced` better

crmsh/bootstrap.py
```diff
@@ -1740,7 +1748,7 @@ def join_csync2(seed_host):
     # they haven't gone to all nodes in the cluster, which means a
     # subseqent join of another node can fail its sync of corosync.conf
     # when it updates expected_votes.  Grrr...
-    if not invoke('ssh -o StrictHostKeyChecking=no root@%s "csync2 -mr / ; csync2 -fr / ; csync2 -xv"' % (seed_host)):
+    if not invoke('ssh -o StrictHostKeyChecking=no root@{} "csync2 -mr /; csync2 -xv || csync2 -fr / && csync2 -xv"'.format(seed_host)):
```
arbulu89: Just to keep the commands more aligned I would replace
- `-mr` by `-rm`
- `xv` by `rxv` (is there any reason why we don't put `r` here ?)
- `-fr` by `-rf`.
This way is easier to identify that the used commands are the same than in the other usages

test/features/qdevice_setup_remove.feature
```diff
@@ -32,6 +32,8 @@ Feature: corosync qdevice/qnetd setup/remove process
     Then    Cluster service is "started" on "hanode2"
     And     Online nodes are "hanode1 hanode2"
     And     Service "corosync-qdevice" is "stopped" on "hanode2"
+    When    Run "echo "# This is test for bsc#1166684" >> /etc/corosync/corosync.conf" on "hanode1"
```
arbulu89: Update `This is test...` by `This is a test..`

crmsh/bootstrap.py
liangxin1300: Changed and sync to other two branches

crmsh/bootstrap.py
liangxin1300: Changed and sync to other two branches

test/features/qdevice_setup_remove.feature
liangxin1300: Changed and sync to other two branches

liangxin1300:
@arbulu89 , Thanks for suggestions! I think now it can be approved, and other two backports PR
Thank you!

PR #545, opened 2020-03-18
Fix: bootstrap: Change condition to add stonith-sbd resource

liangxin1300:
`stonith-sbd` resource should not be added unless sbd is configured and sbd device be specified by '-s' option

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -660,7 +660,7 @@ def init_cluster_local():
 
     # for cluster join, diskless_sbd flag is set in join_cluster() if
     # sbd is running on seed host
-    if configured_sbd_device() or _context.diskless_sbd:
+    if (configured_sbd_device() and _context.sbd_device) or _context.diskless_sbd:
         invoke("systemctl enable sbd.service")
     else:
         invoke("systemctl disable sbd.service")
@@ -1496,7 +1496,7 @@ op_defaults op-options: timeout=600 record-pending=true
 rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
 
-    if configured_sbd_device():
+    if configured_sbd_device() and _context.sbd_device:
         if not invoke("crm configure primitive stonith-sbd stonith:external/sbd pcmk_delay_max=30s"):
             error("Can't create stonith-sbd primitive")
         if not invoke("crm configure property stonith-enabled=true"):


liangxin1300:
Hi @arbulu89 ,
> @liangxin1300 It makes sense. If we don't define sbd we don't need this resource.
> Anyway this change will affect current usage. I guess there is not much risk removing this default behaviour, right?

No, this PR will not affect current usage.
Actually I've fired a new bug, bsc#1166967,
This PR will fix this bug's scenarios:
```
Setup a cluster with sbd device using crmsh bootstrap "-s" option, we got a
"stonith-sbd" resource running.

Then stop the cluster.

This time run "crm cluster init -y", without specify sbd device, still got a
"stonith-sbd" resource running.
```

> By the way, I guess that SBD is configured during this bootstrap process, right? So, I was wondering when we would have the scenario where `configured_sbd_device` returns something without setting `sbd` in the bootstrap operation

There might be another issue should be resolved: sbd service will be started, only `/etc/sysconfig/sbd` has configured `SBD_DEVICE`, even not using bootstrap option '-s'

liangxin1300:
@arbulu89 I've updated the code, change the condition to enable sbd service as 
```
    if (configured_sbd_device() and _context.sbd_device) or _context.diskless_sbd:
        invoke("systemctl enable sbd.service")
    else:
        invoke("systemctl disable sbd.service")
```
In this way, sbd service will not be started with cluster service, even the `SBD_DEVICE` was configured in `/etc/sysconfig/sbd`, only if without specify `-s` option while bootstrap init

liangxin1300:
@arbulu89 @gao-yan 
It seems a bit of tricky here.
For scenarios of bsc#1166967, SBD resource still be configured even not specify with '-s' option while re-configuring on init node;
And on join node, sbd service still started only if found sbd configured;
Then I created PR https://github.com/ClusterLabs/crmsh/pull/545/files to fix above issue
but forget that join node doesn't know whether init node has specified -s option, so the value of _context.sbd_device always None on join node

For now I have three ideas:
1. Comment out or remove SBD_DEVICE line in sysconfig/sbd when -s option not provided on init node
2. on join node, ssh to init node, check the status of sbd service, if it's running and have configured, enable sbd service
3. on init node, recording all contents of Context into a file and on join node, fetch that file. So join node will know exactly which option has been specified

arbulu89:
@liangxin1300 Can you open a github issue for this? It will be easier to follow up instead of merged PR. You can just copy/paste the content

PR #550, opened 2020-03-21
Bootstrap network improvement

liangxin1300:
Changes:
- Enable specify second interface using `-i` option while init and join process
- Add `validate_option` in `Context` class to valid options(reusable by init and join process)
- Add class `utils.IP` to get some properties of IP address
- Add class `utils.Interface` to get information from one interface
- Add class `utils.InterfacesInfo` to collect interfaces information on local node
- Refactor `bootstrap.init_network` function, get all needed network information through `utils.InterfacesInfo`
- Refactor multiple `bootstrap.valid_*` functions, add class `bootstrap.Validation` to validate values from interactive inputs
- Delete redundant codes and functions

TODO:
- [x] More test for interactive mode
- [ ] More test in cloud environment
- [x] UT and functional test, append ipv6 scenarios
- [x] delete more codes which will not be used anymore, like `utils.network_defaults`

RPM:
https://build.opensuse.org/package/show/home:XinLiang:branches:network:ha-clustering:Factory/crmsh

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -25,9 +25,6 @@ from . import term
 from .msg import common_warn, common_info, common_debug, common_err, err_buf
 
 
-mcast_regrex = r'2(?:2[4-9]|3\d)(?:\.(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d?|0)){3}'
-
-
 def to_ascii(input_str):
     """Convert the bytes string to a ASCII string
     Usefull to remove accent (diacritics)"""
@@ -91,94 +88,6 @@ gethomedir = userdir.gethomedir
 def this_node():
     'returns name of this node (hostname)'
     return os.uname()[1]
-
-
-def network_defaults(interface=None):
-    """
-    returns (interface, ip-address, network, prefix-length)
-    """
-    def valfor(l, key):
-        for i in range(0, len(l) - 1):
-            if l[i] == key:
-                return l[i + 1]
-        return None
-    _, outp = get_stdout("/sbin/ip -o route show")
-    info = [None, None, None, None]
-    if interface is not None:
-        info[0] = interface
-    for l in outp.splitlines():
-        sp = l.split()
-        if info[0] is None and len(sp) >= 5 and sp[0] == 'default' and sp[1] == 'via':
-            info[0] = sp[4]
-        if info[0] is not None and valfor(sp, 'dev') == info[0] and sp[0] != 'default':
-            src = valfor(sp, 'src')
-            if sp[0].find('/') >= 0:
-                nw, length = sp[0].split('/')
-                info[1], info[2], info[3] = src, nw, length
-                break
-            # we are reading /32 route entry
-            else:
-                info[1], info[2], info[3] = src, sp[0], 32
-    if info[0] is None:
-        raise ValueError("Failed to determine default network interface")
-    if info[1] is None:
-        ips = ip_in_local()
-        if len(ips) > 0:
-            info[1] = ips[0]
-    return tuple(info)
-
-
-def network_all(with_mask=False):
-    """
-    returns all networks on local node
-    """
-    all_networks = []
-    _, outp = get_stdout("/sbin/ip -o route show")
-    for l in outp.split('\n'):
-        if re.search(r'\.0/[0-9]+ ', l):
-            if with_mask:
-                all_networks.append(l.split()[0])
-            else:
-                all_networks.append(l.split('/')[0])
-    return all_networks
-
-
-def network_v6_all():
-    _, outp = get_stdout("/sbin/ip -6 -o addr show")
-    dict_ = {}
-    for line in outp.split('\n'):
-        if re.search(r' ::1/| [Ff][Ee]80:', line):
-            # skip local address and link-local address
-            continue
-        dict_[line.split()[1]] = []
-    for line in outp.split('\n'):
-        if re.search(r' ::1/| [Ff][Ee]80:', line):
-            # skip local address and link-local address
-            continue
-        dict_[line.split()[1]].append(line.split()[3])
-    return dict_
-
-
-def ip_in_local(IPv6=False):
-    # short-circuit ip list handling for
-    # cloud providers in order to use
-    # metadata service instead
-    ips = iplist_for_cloud()
-    if len(ips) > 0:
-        return ips
-
-    if not IPv6:
-        regex = r' [0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}/[0-9]+ '
-    else:
-        regex = r' [0-9a-fA-F]{1,4}:.*:[0-9a-fA-F]{1,4}/[0-9]+ '
-
-    ip_local = []
-    _, outp = get_stdout("/sbin/ip addr show")
-    tmp = re.findall(regex, outp, re.M)
-    if tmp:
-        tmp = map(lambda x: x.split('/')[0].strip(), tmp)
-        ip_local += filter(lambda x: x != "127.0.0.1" and not re.search(r'^[Ff][Ee]80:', x), tmp)
-    return ip_local
 
 
 _cib_shadow = 'CIB_shadow'
@@ -2050,45 +1959,11 @@ def obscure(obscure_list):
         _obscured_nvpairs = prev
 
 
-def valid_ip_addr(addr, version=4):
-    import socket
-    try:
-        if version == 4:
-            socket.inet_pton(socket.AF_INET, addr)
-        elif version == 6:
-            socket.inet_pton(socket.AF_INET6, addr)
-        else:
-            return False
-    except socket.error:
-        return False
-    return True
-
-
-def get_ipv6_network(addr_with_mask):
-    return str(ipaddress.ip_interface(addr_with_mask).network.network_address)
-
-
 def gen_nodeid_from_ipv6(addr):
     return int(ipaddress.ip_address(addr)) % 1000000000
 
 
-def ip_in_network(addr, net):
-    return ipaddress.ip_address(addr) in ipaddress.ip_interface(net).network
-
-
-class IP:
-    def __init__(self, addr):
-        self.addr = ipaddress.ip_address(addr)
-
-    def version(self):
-        return self.addr.version
-
-
-def is_ipv6(addr):
-    return IP(addr).version() == 6
-
-
-# Set by detect_cloud() or iplist_for_cloud()
+# Set by detect_cloud()
 # to avoid multiple requests
 _ip_for_cloud = None
 
@@ -2158,24 +2033,6 @@ def detect_cloud():
             _ip_for_cloud = result
             return "google-cloud-platform"
     return None
-
-
-def iplist_for_cloud():
-    """
-    Return list of viable IPs for the
-    current cloud provider (if any)
-    """
-    global _ip_for_cloud
-    provider = detect_cloud()
-    if _ip_for_cloud is not None:
-        return [_ip_for_cloud]
-    if provider == "amazon-web-services":
-        uri = "http://169.254.169.254/latest/meta-data/local-ipv4"
-        result = _cloud_metadata_request(uri)
-        if result:
-            _ip_for_cloud = result
-            return [_ip_for_cloud]
-    return []
 
 
 def debug_timestamp():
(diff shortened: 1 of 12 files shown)


crmsh/utils.py
```diff
@@ -2287,4 +2253,58 @@ def interface_choice():
...
+    nic_ip_dict = {}
+    ip_list = []
+    ip_with_mask_list = []
+
+    cmd = "ip -{} -o addr show".format(6 if ipv6 else 4)
+    rc, out, err = get_stdout_stderr(cmd)
+    if rc != 0:
+        raise ValueError(err)
+
+    for line in out.split('\n'):
```
arbulu89, line 2275: You could just use `splitlines()`

crmsh/utils.py
```diff
@@ -2287,4 +2253,58 @@ def interface_choice():
...
+        _, nic, _, ip_with_mask, *_ = line.split()
```
arbulu89, line 2276: Even though most probably it works, this line looks pretty weak. I would prefer to have a regular expression here. Anyway, if the tests pass with this solution I'm fine. Maybe add just a comment before this line with a an example like:
`2: eth0    inet 10.162.31.141/18 brd 10.162.63.255 scope global eth0\       valid_lft forever preferred_lft forever`

crmsh/utils.py
```diff
@@ -2287,4 +2253,58 @@ def interface_choice():
...
+        if nic == "lo":
+            continue
+        ip, _ = ip_with_mask.split('/')
+        ip_list.append(ip)
+        ip_with_mask_list.append(ip_with_mask)
+        # get first address on each nic
+        if nic not in nic_ip_dict:
+            nic_ip_dict[nic] = ip
+    return nic_ip_dict, ip_list, ip_with_mask_list
```
arbulu89, line 2285: I don't like this return. It returns a dictionary and 2 list, with really redundant data. Having such a complex return brings a lot of issues in the future and makes the code really difficult to maintain
I would prefer to have a unique dictionary, or even better a class.
Option 1:
```
{
    "eth0" : {
        "ip": "1.2.3.4".
        "mask": 24
    },
# More network interfaces
}
```

Option 2: Create a really simple class that has a list of interfaces, and each interface has the nic, ip address and the mask.

crmsh/utils.py
```diff
@@ -2287,4 +2253,58 @@ def interface_choice():
...
+    Return None if route not configured
+    """
+    cmd = "ip -o route show"
+    rc, out, err = get_stdout_stderr(cmd)
+    if rc != 0:
+        raise ValueError(err)
+    res = re.search(r'^default via .* dev (.*?) proto', out)
+    if res:
+        return res.group(1)
+    else:
```
arbulu89, line 2308: In this case, is better to raise an exception. Not having a default is an exception. Better to treat this way instead of returning `None`

liangxin1300:
> @liangxin1300 Some comments with possible improvements.
> Besides that, I don't understand why we need a 2nd network interface. Could you explain the use case?

@arbulu89 , corosync2 support at most 2 interfaces
Currently, if user want to use crmsh to configure 2 interfaces, '-M' should be specified, but it's not flexible enough, '-M' always fetch first two interfaces.
Assuming each node has multi interfaces, user can use '-i' option to configure specific interface they want be used by corosync

crmsh/utils.py
liangxin1300: Well, `splitlines()` looks good, it can strip last '\n', while `split('\n')` will not

crmsh/utils.py
liangxin1300: I've added a comment here

crmsh/utils.py
liangxin1300: I prefer to raise an exception and treat that exception as a warning, not an error, since configure could be continued

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
...
+
+    @property
+    def network_list(self):
+        return [interface.network for interface in self.interface_list]
+
+
+class Interface(object):
+    def __init__(self, ip_with_mask):
+        self.ip = ip_with_mask.split('/')[0]
+        self.ip_with_mask = ip_with_mask
```
arbulu89, line 2326: Why do you store the ip with mask? Store the mast and that's it, right? You can combine them in anytime as you have both there

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
     # should consider interface format like "ethx@xxx"
     interface_list = re.findall(r'(?:[0-9]+:) (.*?)(?=: |@.*?: )', out)
     return [nic for nic in interface_list if nic != "lo"]
+
+
+class InterfaceInfo(object):
```
arbulu89, line 2258: I would change to `InterfacesInfo` as you store there all of them

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
...
+
+
+class Interface(object):
+    def __init__(self, ip_with_mask):
+        self.ip = ip_with_mask.split('/')[0]
+        self.ip_with_mask = ip_with_mask
+        self.network = get_network(ip_with_mask)
+
+
+def get_network(ip_with_mask):
```
arbulu89, line 2330: Why don't you put this `get_network` inside `Interface`. It makes sense, right?
This way we don't need to provide any parameter as the `ip_with_mask` can be created with the `self` attributes. You could provider `ip_with_mask` as getter if you want

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
...
+    IPv6 support
+    """
+    return str(ipaddress.ip_interface(ip_with_mask).network.network_address)
+
+
+class NoDefaultRouteError(Exception):
+    pass
+
+
+def get_default_nic_from_route():
```
arbulu89, line 2342: I would add this method to the `InterfacesInfo` as classmethod as it's related to the interfaces data

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
     # should consider interface format like "ethx@xxx"
     interface_list = re.findall(r'(?:[0-9]+:) (.*?)(?=: |@.*?: )', out)
     return [nic for nic in interface_list if nic != "lo"]
+
+
+class InterfaceInfo(object):
+    def __init__(self, ipv6=False, second_hb=False):
+        self.ipv6 = ipv6
+        self.nic_info_dict = {}
+        self.second_hb = second_hb
```
arbulu89, line 2262: Maybe better `second_heartbeat`, `hb` is not really meaningfull

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
...
+            _ip_list.append(self.nic_info_dict[nic][0].ip)
+        if self.second_hb and len(specify_nic_list) == 1:
+            for nic in self.all_nic:
+                if nic not in specify_nic_list:
+                    _ip_list.append(self.nic_info_dict[nic][0].ip)
+                    break
+        return _ip_list
+
+    @property
+    def all_nic(self):
```
arbulu89, line 2300: Maybe better `nic_list` to be aligned with the other method names

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
     # should consider interface format like "ethx@xxx"
     interface_list = re.findall(r'(?:[0-9]+:) (.*?)(?=: |@.*?: )', out)
     return [nic for nic in interface_list if nic != "lo"]
+
+
+class InterfaceInfo(object):
+    def __init__(self, ipv6=False, second_hb=False):
+        self.ipv6 = ipv6
+        self.nic_info_dict = {}
```
arbulu89, line 2261: I think it's better to use `nic_info_dict` like protected variable. It looks like an internal data structure

crmsh/utils.py
```diff
@@ -2287,4 +2253,104 @@ def interface_choice():
...
+                continue
+            # one nic might configured multi IP addresses
+            if nic not in self.nic_info_dict:
+                self.nic_info_dict[nic] = []
+            self.nic_info_dict[nic].append(Interface(ip_with_mask))
+
+        if not self.nic_info_dict:
+            raise ValueError("No address configured")
+        if self.second_hb and len(self.nic_info_dict) == 1:
+            raise ValueError("Cannot configure second heartbeat line, since only one address available")
```
arbulu89, line 2284: `...only one address **is** available..`

crmsh/utils.py
liangxin1300: Changed

crmsh/utils.py
liangxin1300: Changed

crmsh/utils.py
liangxin1300: OK, changed!

crmsh/utils.py
liangxin1300: changed

crmsh/utils.py
liangxin1300: changed

crmsh/utils.py
liangxin1300: changed

crmsh/utils.py
liangxin1300: changed

crmsh/utils.py
```diff
@@ -2287,4 +2249,109 @@ def interface_choice():
...
+    @property
+    def network_list(self):
+        return [interface.network for interface in self.interface_list]
+
+
+class Interface(object):
+    def __init__(self, ip_with_mask):
+        self.ip, self.mask = ip_with_mask.split('/')
+
+    def get_network(self):
```
arbulu89, line 2347: I think we can remove this method `get_network` method, and just set `self.nework` in the `init` method. Having `property` type method calling a method is a bit strange. In fact, it looks like this value is static.

Anyway, current solution works too

crmsh/utils.py
```diff
@@ -2287,4 +2249,109 @@ def interface_choice():
...
     interface_list = re.findall(r'(?:[0-9]+:) (.*?)(?=: |@.*?: )', out)
     return [nic for nic in interface_list if nic != "lo"]
+
+
+class InterfacesInfo(object):
+    def __init__(self, ipv6=False, second_heartbeat=False, specify_nic_list=[]):
+        self.ipv6 = ipv6
+        self._nic_info_dict = {}
+        self.second_heartbeat = second_heartbeat
+        self.get_interface_info()
```
arbulu89, line 2259: I'm not a big fun of having methods in the `__init__`, but this is just a personal preference hehe
They make unit testing a bit more complicated.
Anyway, it's Ok if you like this way.

crmsh/utils.py
```diff
@@ -2287,4 +2249,109 @@ def interface_choice():
     # should consider interface format like "ethx@xxx"
     interface_list = re.findall(r'(?:[0-9]+:) (.*?)(?=: |@.*?: )', out)
     return [nic for nic in interface_list if nic != "lo"]
+
+
+class InterfacesInfo(object):
+    def __init__(self, ipv6=False, second_heartbeat=False, specify_nic_list=[]):
```
arbulu89, line 2255: `specify_nic_list` name is a bit strange. We should find something better. Maybe `custom_nic_list` or simply `nic_list`

crmsh/utils.py
```diff
@@ -2287,4 +2249,109 @@ def interface_choice():
...
+    def get_default_nic_list_from_route(self):
+        cmd = "ip -o route show"
+        rc, out, err = get_stdout_stderr(cmd)
+        if rc != 0:
+            raise ValueError(err)
+        res = re.search(r'^default via .* dev (.*?) ', out)
+        if res:
+            return [res.group(1)]
+        else:
+            common_warn("No default route configured")
```
arbulu89, line 2306: Maybe better `No default route configured. Using the first found nic`

liangxin1300: changed

crmsh/utils.py
liangxin1300: Changed to `custom_nic_list`

crmsh/utils.py
liangxin1300: Changed, no function be called in __init__

crmsh/utils.py
```diff
@@ -2292,4 +2149,255 @@ def interface_choice():
...
+
+    @classmethod
+    def is_ipv6(cls, addr):
+        """
+        Check whether the address is IPV6 address
+        """
+        return cls(addr).version == 6
+
+    @classmethod
+    def is_valid_IP(cls, addr):
```
arbulu89, line 2195: Why `IP` in capital letters here? In other parts of the code we use `ip`

crmsh/utils.py
```diff
@@ -2292,4 +2149,255 @@ def interface_choice():
...
+        cls_inst = cls(is_ipv6)
+        cls_inst.get_interfaces_info()
+        return cls_inst.ip_list
+
+    @property
+    def network_list(self):
+        """
+        Get network list
+        """
+        return list(set([interface.network for interface in self.interface_list]))
```
arbulu89, line 2346: Why do you do `list` and `set` operations? Just to understand what they are for

crmsh/utils.py
```diff
@@ -2292,4 +2149,255 @@ def interface_choice():
...
+        cmd = "ip -o route show"
+        rc, out, err = get_stdout_stderr(cmd)
+        if rc != 0:
+            raise ValueError(err)
+        res = re.search(r'^default via .* dev (.*?) ', out)
+        if res:
+            self._default_nic_list = [res.group(1)]
+        else:
+            common_warn("No default route configured. Using the first found nic")
+            self._default_nic_list = [self.nic_list[0]]
```
arbulu89, line 2371: This will only work if `get_interfaces_info` is executed before, right?
So we might have 2 different returns depending on how we use this method, not so nice.
Should be run `get_interfaces_info` here just in case?

crmsh/utils.py
```diff
@@ -2292,4 +2149,255 @@ def interface_choice():
...
+        Init function
+
+        On init process,
+        "ipv6" is provided by -I option
+        "second_heartbeat" is provided by -M option
+        "custom_nic_list" is provided by -i option
+        """
+        self.ip_version = 6 if ipv6 else 4
+        self.second_heartbeat = second_heartbeat
+        self._default_nic_list = custom_nic_list
```
arbulu89, line 2278: I don't understand much why you set the `custom_nic_list` to `default_nic_list`. Why don't you just keep `self._nic_list` or similar. What's the meaning of `default` and `custom` here?

liangxin1300:
> @liangxin1300 I think the code is good!
> I have found some questions because I don't understand some functionalities, but besides that I see that you have created a lot of tests, so it looks pretty safe to say that the changes are good.
> 
> The unique thing I could be a bit afraid of is that most of the methods in the class `InterfacesInfo` depends on having `get_interfaces_info` executed, otherwise the things will be different. I guess you will run this command always, but maybe it would be good to write in the class docstring that this is the method that populates the data, or even just execute in in the `__init__` or create a new classmethod, that returns and instance with this method executed, something like: `interface_info = InterfaceInfo.get_data(...)```
> 
> **Edit:** I see that you already have a similar code for that
@arbulu89 Thanks for pointing this, I've added a docstring in this method
```
IMPORTANT: This is the method that populates the data, should always be called after initialize
```
And for those scenarios which need to use a method simply, I've created two `classmethod`: `get_local_ip_list` and `ip_in_network`, do you think that is enough?

crmsh/utils.py
liangxin1300: OK, changed

crmsh/utils.py
liangxin1300: To deduplicate the possible same network value

crmsh/utils.py
liangxin1300: Thanks, changed

PR #551, opened 2020-03-22
Simplify bootstrap context

liangxin1300:
Simplify context of bootstrap, based on below reasons:
1. Current bootstrap.Context assign attributes multi times
2. Function like bootstrap_* has too many parameters
3. Not easy to maintain if we want to add new options

Other changes:
1. Since the change in `bootstrap.py` involved geo cluster related functions, I setup functional test for geo cluster setup using behave, to make sure that process still working.
2. opensuse/leap:15.2 currently not stable, switch to opensuse/tumbleweed as docker image for master branch


crmsh/ui_cluster.py
```diff
@@ -353,14 +345,13 @@ def do_join(self, context, *args):
...
-        bootstrap.bootstrap_join(
-            cluster_node=options.cluster_node,
-            ui_context=context,
-            nic=options.nic,
-            quiet=options.quiet,
-            yes_to_all=options.yes_to_all,
-            watchdog=options.watchdog,
-            stage=stage)
+        join_context = bootstrap.Context()
+        for opt in vars(options):
```
arbulu89: We could put this many time duplicated lines in a method, right?

crmsh/ui_cluster.py
```diff
@@ -283,27 +284,18 @@ def looks_like_hostnames(lst):
...
-            no_overwrite_sshkey=options.no_overwrite_sshkey,
-            unicast=options.unicast,
-            second_hb=options.second_hb,
-            ipv6=options.ipv6,
-            watchdog=options.watchdog,
-            qdevice=qdevice,
-            stage=stage,
-            args=args)
+        for opt in vars(options):
+            if opt.startswith("qdevice"):
```
arbulu89: Do you really need to do this `opt.startswith("qdevice")` check?
Why don't you just update the order of the code, and manage qdevice after this loop?

arbulu89: This way you could have the same functionality you have in join and remove

crmsh/ui_cluster.py
```diff
@@ -524,7 +512,15 @@ def do_geo_init(self, context, *args):
                 ticketlist = [t for t in re.split('[ ,;]+', options.tickets)]
             except ValueError:
                 parser.error("Invalid ticket list")
-        bootstrap.bootstrap_init_geo(options.quiet, options.yes_to_all, options.arbitrator, clustermap, ticketlist, ui_context=context)
+
+        geo_context = bootstrap.Context()
+        for opt in vars(options):
```
arbulu89: Again, code duplicated

crmsh/ui_cluster.py
liangxin1300: No, no matter what's the order of this block of code, since it was a loop of `options`, qdevice option will be included always

liangxin1300: Solution might be, move this block of `corosync.Qdevice` into function `bootstrap.bootstrap_init`

liangxin1300:
> @liangxin1300 First of all, I cannot figure it out, what is happening with all of the parameters you are removing from this methods? Could you elaborate a bit more?

Although I could not tell exactly, it's a feature of python class, the result is, through `setattr` method, all options defined in argparse has become attributes of `class Context` 

> Second, I see the possibility to remove a lot of duplicated code

I've create function `set_bootstrap_context_attr` to remove duplicated codes, and `init_qdevice` in class `Context` to handle qdevice options

crmsh/bootstrap.py
```diff
@@ -2271,6 +2233,11 @@ def check_option():
             except ValueError as err:
                 error(err)
 
+    global _context
+    _context = context
+    _context.init_qdevice()
+
+    stage = _context.stage
```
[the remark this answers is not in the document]
liangxin1300: Just for compatibility for original code.
Take above block of code as example, variables `stage` appears many times in that function, it originally as a function parameter, if I don't do that, I must replace all `stage` as `_context.stage`
@arbulu89 what do you think?

crmsh/bootstrap.py
```diff
@@ -50,27 +50,31 @@ class Context(object):
...
     to every bootstrap method.
     """
-    def __init__(self, quiet, yes_to_all, nic=None, ip_address=None, ip_network=None):
-        self.quiet = quiet
-        self.yes_to_all = yes_to_all
-        self.nic = nic
-        self.ip_address = ip_address
-        self.ip_network = ip_network
-        self.cluster_name = None
+    def __init__(self):
```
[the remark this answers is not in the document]

crmsh/ui_cluster.py
```diff
@@ -64,6 +64,13 @@ def get_cluster_name():
     return cluster_name
 
 
+def set_bootstrap_context_attr(options):
```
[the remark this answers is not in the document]
liangxin1300: changed, thanks!

PR #563, opened 2020-04-03
Validate interface using argparse choices

liangxin1300:
To make sure we got the right interface name when init/join

Additionally, use choices to validate -t option

### DIFF
--- a/test/unittests/test_utils.py
+++ b/test/unittests/test_utils.py
@@ -571,3 +571,35 @@ def test_detect_cloud_gcp_rc_error(mock_metadata, mock_get_stdout, mock_is_progr
     ])
     assert mock_metadata.call_count == 0
     assert utils._ip_for_cloud is None
+
+
+@mock.patch("crmsh.utils.get_stdout")
+def test_interface_choice(mock_get_stdout):
+    ip_a_output = """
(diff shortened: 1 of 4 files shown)


test/unittests/test_utils.py
```diff
@@ -571,3 +571,13 @@ def test_detect_cloud_gcp_rc_error(mock_metadata, mock_get_stdout, mock_is_progr
     ])
     assert mock_metadata.call_count == 0
     assert utils._ip_for_cloud is None
+
+
+@mock.patch("re.findall")
```
arbulu89, line 576: I wouldn't mock the `findall` method. It masks the main functionality of the method.
In this case, I would just set the output of `get_stdout` with a real `ip a` output.
Otherwise, the UT is almost useless

test/features/steps/const.py
```diff
@@ -62,7 +62,7 @@
                         destructive, especially during the "storage" stage.
                         The /root/.ssh/id_rsa key will be overwritten unless
                         the option "--no-overwrite-sshkey" is used)
-  -t TEMPLATE, --template TEMPLATE
+  -t {ocfs2}, --template {ocfs2}
```
arbulu89, line 65: Why this change? I think that template was a better option. It's true that by now we only have `ocfs2. What if we add some new option in the future, the name won't have sense anymore

liangxin1300:
Thanks @arbulu89

PR #567, opened 2020-04-09
Low: bootstrap: exit with proper error messages when ssh return failed

liangxin1300:
When doing some ssh/scp jobs, ssh connect will failed if not input password for a long time.
In that way, process should exit, and with proper error messages


liangxin1300:
Thanks @gao-yan !

PR #571, opened 2020-04-17
Fix: bootstrap: no overwrite ssh key if detect passwordless(bsc#1169581)

liangxin1300:
Steps to reproduce,
I have two nodes(node1 and node2)
1. each node generate its ssh key
2. on node1, run cat .ssh/id_rsa.pub | ssh root@node2 'cat >> .ssh/authorized_keys'
3. on node2, run cat .ssh/id_rsa.pub | ssh root@node1 'cat >> .ssh/authorized_keys'
4. now node1 and node2 can login each other without password
5. on node1, run crm cluster init -y --no-overwrite-sshkey
6. on node2, run crm cluster join -c node1 -y, since now on node2, all the key were overwrite by crmsh, join process still require password input

--no-overwrite-sshkey can just avoid overwrite existing keys on init node

#### Reason
The unique ssh key and authorized_keys will be overwritten since `-y` option was specified while joining.
#### Solution
Before overwrite join node's keys with init node's keys, test whether join node can login to init node without password, if yes, even `-y` option specified, join node's key will not be overwritten

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,12 @@ jobs:
       script:
         - $FUNCTIONAL_TEST bootstrap run bugs
 
+    - name: "regression test for bootstrap bugs (second set)"
+      before_install:
+        - $FUNCTIONAL_TEST bootstrap before_install
+      script:
+        - $FUNCTIONAL_TEST bootstrap run bugs2
+
     - name: "functional test for bootstrap - init, join and remove"
       before_install:
         - $FUNCTIONAL_TEST bootstrap before_install
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -752,6 +752,7 @@ def init_ssh():
                 not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
             return
         rmfile("/root/.ssh/id_rsa")
+        rmfile("/root/.ssh/authorized_keys")
     status("Generating SSH key")
     invoke("ssh-keygen -q -f /root/.ssh/id_rsa -C 'Cluster Internal' -N ''")
     append("/root/.ssh/id_rsa.pub", "/root/.ssh/authorized_keys")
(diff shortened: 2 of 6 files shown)


crmsh/bootstrap.py
```diff
@@ -752,6 +752,7 @@ def init_ssh():
                 not confirm("/root/.ssh/id_rsa already exists - overwrite?"):
             return
         rmfile("/root/.ssh/id_rsa")
+        rmfile("/root/.ssh/authorized_keys")
```
liangxin1300, line 755: This step seems destructive when confirm to overwrite.
But if not rm authorized_keys, the file might still include peer node's public key, then when peer node joining, peer node detect that no need password to login to init node, so peer node will not use init node's new generated keys

.travis.yml
```diff
@@ -44,6 +44,12 @@ jobs:
       script:
         - $FUNCTIONAL_TEST bootstrap run bugs
 
+    - name: "regression test for bootstrap bugs (second set)"
```
arbulu89, line 47: Why have you added a new test execution? Why not just use the first `regression test for bootstrap bugs`?

crmsh/bootstrap.py
arbulu89: I don't like this. You are basing what to do in order to make your solution work. If this file creates issues in your new approach, I don't think the solution is to remove this file, we need to think about other solution first.
We already started wrong `crmsh` using `/root/.ssh` keys, let's avoid interfiering more in the machines environment.
I don't that this kind of actions will bring issues in the future.

crmsh/bootstrap.py
```diff
@@ -1701,6 +1697,19 @@ def init():
     init_network()
 
 
+def append_to_file(source_file, target_file):
```
arbulu89, line 1700: This looks like a method that should go in `utils.py`, right?

crmsh/bootstrap.py
```diff
@@ -1725,11 +1740,17 @@ def join_ssh(seed_host):
             continue
 
         if os.path.exists(os.path.join("/root/.ssh", key)):
-            if not confirm("/root/.ssh/%s already exists - overwrite?" % (key)):
+            # not yes or not True  ==> False, overwrite
+            # not yes or not False ==> True, continue
+            # not no or not True ==> True, continue
+            # not no or not False  ==> True, continue
+            if not confirm("/root/.ssh/%s already exists - overwrite?" % (key)) or not need_pw:
```
arbulu89, line 1747: I don't understand this `confirm` question. overwrite what? The authorized_keys content?

.travis.yml
liangxin1300: To deploy unique ssh key on each node.
Since from behave(running from one node), it's hard to generate unique ssh key on each node without disturb ssh connection

crmsh/bootstrap.py
MalloZup: removing a 3rd party file should not be the case. I agree on this

crmsh/bootstrap.py
liangxin1300: Overwrite the ssh key if confirmed `yes`, so we need add `need_pw` flag here, to avoid overwrite if already passwordless

PR #575, opened 2020-04-23
Fix: bootstrap: crmsh use its own specific ssh key(bsc#1169581)

liangxin1300:
#### Motivation
Using default ssh key of system might have potential side effects and not flexible
#### Major changes:
* bootstrap.init_ssh/join_ssh only maintain **crmsh key: /root/.ssh/id_rsa.crmsh**
* ssh/scp use -i option to use the crmsh key
* For crmsh.parallax.parallax_call/parallax_copy/parallax_slurp, whether to use crmsh key is optional(default yes)
* Remove --no-overwrite-sshkey option
* Remove ssh_remote stage
* Remove ssh_merge stage
* Related change for hb_report
* Related changes for UT and functional test

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -29,6 +29,7 @@ from . import corosync
 from . import tmpfiles
 from . import clidisplay
 from . import term
+from .constants import SSH_KEY_CRMSH, SSH_WITH_KEY, SCP_WITH_KEY, SSH_KEY_CRMSH_TAG
 
 
 LOG_FILE = "/var/log/ha-cluster-bootstrap.log"
@@ -42,7 +43,7 @@ PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 
 
-INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
+INIT_STAGES = ("ssh", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
 
 
 class Context(object):
@@ -60,7 +61,6 @@ class Context(object):
         self.cluster_name = None
         self.diskless_sbd = None
         self.watchdog = None
-        self.no_overwrite_sshkey = None
         self.nic = None
         self.unicast = None
         self.admin_ip = None
@@ -276,7 +276,7 @@ def get_cluster_node_hostname():
     peer_node = None
     if _context.cluster_node:
         if utils.valid_ip_addr(_context.cluster_node):
-            rc, out, err = utils.get_stdout_stderr("ssh {} crm_node --name".format(_context.cluster_node))
+            rc, out, err = utils.get_stdout_stderr("{} {} crm_node --name".format(SSH_WITH_KEY, _context.cluster_node))
             if rc != 0:
                 error(err)
             peer_node = out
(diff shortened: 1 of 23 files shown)


crmsh/ui_cluster.py
```diff
@@ -637,6 +634,8 @@ def do_run(self, context, cmd, *nodes):
         opts = parallax.Options()
         opts.ssh_options = ['StrictHostKeyChecking=no']
         opts.askpass = utils.check_ssh_passwd_need(hosts)
+        if hasattr(opts, 'ssh_key'):
+            opts.ssh_key = SSH_KEY_CRMSH
```
liangxin1300, line 638: For now, there do exist some piece of code in crmsh using parallax directly, instead of using crmsh.parallax
That seems duplicated.
But I suggest we replace them with crmsh.parallax in other PR, not in this one, since validation those functional work was time cost

crmsh/parallax.py
```diff
@@ -14,10 +15,11 @@ class Parallax(object):
     # slurp: Copies files from a set of remote hosts to local folders
     """
     def __init__(self, nodes, cmd=None, localdir=None, filename=None,
-                 src=None, dst=None, askpass=False, ssh_options=None):
+                 src=None, dst=None, askpass=False, ssh_options=None, use_crmsh_key=None):
```
liangxin1300, line 18: For now, this option `use_crmsh_key` is only used by behave, which will set the value `False`

hb_report/utillib.py
```diff
@@ -1532,8 +1533,8 @@ def start_slave_collector(node, arg_str):
             cmd += " {}".format(str(item))
         _, out = crmutils.get_stdout(cmd)
     else:
-        cmd = r'ssh {} {} "{} hb_report __slave"'.\
-              format(constants.SSH_OPTS, node,
+        cmd = r'ssh -i {} {} {} "{} hb_report __slave"'.\
+              format(SSH_KEY_CRMSH, constants.SSH_OPTS, node,
```
arbulu89, line 1537: You could use `SSH_WITH_KEY` I guess

hb_report/utillib.py
```diff
@@ -1632,7 +1633,7 @@ def tail(n, indata):
 
 
 def test_ssh_conn(addr):
-    cmd = r"ssh %s -T -o Batchmode=yes %s true" % (constants.SSH_OPTS, addr)
+    cmd = r"ssh -i %s %s -T -o Batchmode=yes %s true" % (SSH_KEY_CRMSH, constants.SSH_OPTS, addr)
```
arbulu89, line 1636: Use `SSH_WITH_KEY` and `format` to align with other code

crmsh/bootstrap.py
```diff
@@ -743,36 +743,16 @@ def mkdirs_owned(dirs, mode=0o777, uid=-1, gid=-1):
...
-    authkeys_data = authkeys.read()
-    for key in ("id_rsa", "id_dsa", "id_ecdsa", "id_ed25519"):
-        fn = os.path.join("/root/.ssh", key)
-        if not os.path.exists(fn):
-            continue
-        keydata = open(fn + ".pub").read()
-        if keydata not in authkeys_data:
-            append(fn + ".pub", authorized_keys_file)
+        rmfile(SSH_KEY_CRMSH)
+        invoke("sed -i '/crmsh-generated/d' /root/.ssh/authorized_keys")
```
arbulu89, line 753: Add this `crmsh-generated` to a constant. I have seen it in many places. It will be better if we decide to change the value for any reason.

crmsh/parallax.py
arbulu89: @liangxin1300 I don't really like the idea. Why do you need this behaviour? If it's because we need the latest version of parallax, don't worry, we will merge it before this, and we should specify that this version needs the latest paralallax version in the `setup.py` of `spec` file

arbulu89: By the way, we could avoid this if we add the `ssh_key` option as `kwargs` in the last paralllax change. What do you think about this?

arbulu89:
@gao-yan , @diegoakechi ,
What do you think about the new design? I think that it's much better than the old one. Now, we are not going to touch any already existing key. It will change the behaviour for some users that might expect that, but I think it's an improvement in general. Other benefits:
- A lot of code related to ssh keys removed
- We can get rid off `no-overwrite-sshkeys` parameter.
Have a look please.

crmsh/parallax.py
liangxin1300: For now, behave in crmsh testing always running on one node `hanode1`,
while still need to run command on remote node using parallax like:
`When    Run "crm cluster join -c hanode1 -y" on "hanode2"`,
here parallax should use default ssh key provided by my docker image, not crmsh ssh key

crmsh/bootstrap.py
```diff
@@ -1775,9 +1727,8 @@ def join_csync2(seed_host):
     # invoke scp root@seed_host:/etc/hosts $tmp_conf \
     #   || error "Can't retrieve /etc/hosts from seed_host"
     # install_tmp $tmp_conf /etc/hosts
-
-    if not invoke("scp root@%s:'/etc/csync2/{csync2.cfg,key_hagroup}' /etc/csync2" % (seed_host)):
-        error("Can't retrieve csync2 config from %s" % (seed_host))
+    if not invoke("%s root@%s:'/etc/csync2/{csync2.cfg,key_hagroup}' /etc/csync2" % (SCP_WITH_KEY, seed_host)):
```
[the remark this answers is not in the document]
liangxin1300: I did tired, but string format will take this `{csync2.cfg,key_hagroup}` as a value to replace and raise `KeyError: 'csync2'`.
use `\{csync2.cfg,key_hagroup\}` have the same error

crmsh/scripts.py
```diff
@@ -1082,7 +1085,7 @@ def _check_control_persist():
     Checks if ControlPersist is available. If so,
     we'll use it to make things faster.
     '''
-    cmd = 'ssh -o ControlPersist'.split()
+    cmd = 'ssh -i {} -o ControlPersist'.format(SSH_KEY_CRMSH).split()
```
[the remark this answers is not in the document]
liangxin1300: OK, changed

hb_report/utillib.py
liangxin1300: OK, changed

hb_report/utillib.py
liangxin1300: changed

crmsh/bootstrap.py
liangxin1300: changed as `constants.SSH_KEY_CRMSH_TAG`

diegoakechi:
@liangxin1300 @arbulu89 LGTM. It is not invasive anymore. As it brings a behavior change, please, include a meaning description on the change log.

crmsh/bootstrap.py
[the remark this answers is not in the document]
arbulu89: For that you need to put double `{{`. It would be: 
```
"{} root@{}:'/etc/csync2/{{csync2.cfg,key_hagroup}}' /etc/csync2" .format(SCP_WITH_KEY, seed_host)
```

crmsh/parallax.py
arbulu89: I don't get it. I'm afraid this code is hiding some behavioural change. I want to test what happens in the real code, not something done for testing. This is a bad practice in my opinion.
If `parallax` needs to use the default key for joining, just implement this, not this `test mode` code.
To be honest, I don't want this kind of code.

test/features/steps/utils.py
```diff
@@ -25,7 +25,7 @@ def run_command_local_or_remote(context, cmd, addr, err_record=False):
         return out
     else:
         try:
-            results = parallax.parallax_call([addr], cmd)
+            results = parallax.parallax_call([addr], cmd, use_crmsh_key=False)
```
arbulu89, line 28: You shouldn't be using this `parallax.parallax_call` here. You are using the code under test to run tests. 
If you need to use parallax for any reason, import parallax directly and implement the test code.

crmsh/bootstrap.py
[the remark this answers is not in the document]
liangxin1300: Thanks! Already changed!

test/features/steps/utils.py
liangxin1300: Thanks! changed as using parallax directly in behave functional testing

PR #576, opened 2020-04-23
Low: bootstrap: change ha-cluster-bootstrap log path

aleksei-burlakov:
Let's store the bootstrap logs in the more specific location.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -69,6 +69,7 @@ all-local:
 python_prefix = --prefix=$(prefix)
 
 install-exec-local:
+	$(INSTALL) -d -m 770 $(DESTDIR)/${localstatedir}/log/crmsh
 	-mkdir -p $(DESTDIR)$(pkgpythondir)
 	$(PYTHON) $(srcdir)/setup.py install \
 		--root $(DESTDIR)/// \
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -21,6 +21,7 @@ import readline
 import shutil
 from string import Template
 from lxml import etree
+from pathlib import Path
 from . import config
 from . import utils
 from . import xmlutil
@@ -31,7 +32,7 @@ from . import clidisplay
 from . import term
 
 
-LOG_FILE = "/var/log/ha-cluster-bootstrap.log"
+LOG_FILE = "/var/log/crmsh/ha-cluster-bootstrap.log"
 CSYNC2_KEY = "/etc/csync2/key_hagroup"
 CSYNC2_CFG = "/etc/csync2/csync2.cfg"
 COROSYNC_AUTH = "/etc/corosync/authkey"
@@ -145,6 +146,7 @@ def log_file_fallback():
 def log(*args):
     global LOG_FILE
     try:
+        Path(os.path.dirname(LOG_FILE)).mkdir(parents=True, exist_ok=True)
         with open(LOG_FILE, "ab") as logfile:
             text = " ".join([utils.to_ascii(arg) for arg in args]) + "\n"
             logfile.write(text.encode('ascii', 'backslashreplace'))
--- a/crmsh/config.py
+++ b/crmsh/config.py
@@ -277,7 +277,7 @@ DEFAULTS = {
         'from_time': opt_string('-12H'),
         'compress': opt_boolean('yes'),
         'speed_up': opt_boolean('no'),
-        'collect_extra_logs': opt_string('/var/log/messages /var/log/pacemaker/pacemaker.log /var/log/pacemaker.log /var/log/ha-cluster-bootstrap.log'),
+        'collect_extra_logs': opt_string('/var/log/messages /var/log/pacemaker/pacemaker.log /var/log/pacemaker.log /var/log/crmsh/ha-cluster-bootstrap.log'),
         'remove_exist_dest': opt_boolean('no'),
         'single_node': opt_boolean('no')
     }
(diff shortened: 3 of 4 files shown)


crmsh/bootstrap.py
```diff
@@ -141,10 +142,10 @@ def log_file_fallback():
     """
     return os.path.join(utils.get_tempdir(), "ha-cluster-bootstrap.log")
 
-
 def log(*args):
     global LOG_FILE
     try:
+        Path(os.path.dirname(LOG_FILE)).mkdir(parents=True, exist_ok=True)
```
gao-yan, line 148: Such a directory should usually be created by Makefile.am or at the `%install` stage in rpm spec file, and listed as one of the %files there, so that it's maintained by the rpm package.

Makefile.am
```diff
@@ -49,6 +49,7 @@ endif
 
 # Shared data files
 install-data-hook:
+	mkdir -p /var/log/crmsh/; \
```
gao-yan, line 52: This should belong to the `install-exec-local` part, and should be done the conventional way like:
`$(INSTALL) -d  -m 770 $(DESTDIR)/${localstatedir}/log/crmsh`

The macros are important for it to be created properly at the right place depending on build/installation environments, distributions and so on.

This should be also listed in the `%files` in crm.spec.in file as well like:

`%dir %attr (770, %{uname}, %{gname}) %{_var}/log/crmsh`

And please test it and make sure it works both for rpmbuild and manual build/install.

It might still be worth it to check if it exists at runtime and create it if it doesn't.

aleksei-burlakov:
Please don't merge it yet.

crmsh/config.py
```diff
@@ -277,7 +277,7 @@ def get(self, value):
         'from_time': opt_string('-12H'),
         'compress': opt_boolean('yes'),
         'speed_up': opt_boolean('no'),
-        'collect_extra_logs': opt_string('/var/log/messages /var/log/pacemaker/pacemaker.log /var/log/pacemaker.log /var/log/ha-cluster-bootstrap.log'),
+        'collect_extra_logs': opt_string('/var/log/messages /var/log/pacemaker/pacemaker.log /var/log/pacemaker.log /var/log/crmsh/ha-cluster-bootstrap.log'),
```
gao-yan, line 280: IIUC, this should let `crm report` correctly collect this log file?

aleksei-burlakov: Not sure. I saw no difference in the hb report.

gao-yan: The point is, as long as it's collected into a hb_report tar ball :-)

Makefile.am
gao-yan: > It might still be worth it to check if it exists at runtime and create it if it doesn't.

What do you think of this? :-) That's what you partially did in the original commit of this pull request.

aleksei-burlakov: I have added a condition, that the file is only installed if it doesn't exist.

gao-yan: By "runtime", I meant when crmsh is running, like what you did for in the original commit.

aleksei-burlakov: :)

Makefile.am
```diff
@@ -58,6 +58,7 @@ install-data-hook:
 hanoarchdir = $(datadir)/@PACKAGE@/hb_report
 hanoarch_DATA = hb_report/constants.py hb_report/utillib.py
 hanoarch_SCRIPTS = hb_report/hb_report
+var_log_crmsh = $(DESTDIR)/${localstatedir}/log/crmsh
```
gao-yan, line 61: We don't need to define it here. Be aware the prefixes and suffixes of these variables are meaningful and imply certain processing done by automake/make.

Makefile.am
```diff
@@ -69,6 +70,7 @@ all-local:
 python_prefix = --prefix=$(prefix)
 
 install-exec-local:
+	test -d $(var_log_crmsh) || $(INSTALL) -d -m 770 $(var_log_crmsh)
```
gao-yan, line 73: The previous solution was good enough. We don't need to do the test. We can just install the directory.

Makefile.am
aleksei-burlakov: I added the very first code. The Path function will create the /var/log/crmsh folder if it doesn't exist.

crmsh/config.py
aleksei-burlakov: I think, it is simply ignored. (see https://github.com/ClusterLabs/crmsh/blob/master/hb_report/hb_report.in#L64)

aleksei-burlakov: @liangxin1300 , could you please approve or comment on this change?

gao-yan: > I think, it is simply ignored. (see https://github.com/ClusterLabs/crmsh/blob/master/hb_report/hb_report.in#L64)

Simply ignored or collected? We definitely want the latter :-)

aleksei-burlakov: > > I think, it is simply ignored. (see https://github.com/ClusterLabs/crmsh/blob/master/hb_report/hb_report.in#L64)
> 
> Simply ignored or collected? We definitely want the latter :-)

The second. It exists in the constants.EXTRA_LOGS, but the hb_report doesn't care about the value.

gao-yan: @liangxin1300, it sounds like `crm report` no longer collects bootstrap log file even without @aleksei-burlakov's change in there?

liangxin1300: @gao-yan, I've tested @aleksei-burlakov's patch, `hb_report` will collect bootstrap log file which located at `/var/log/crmsh`

PR #584, opened 2020-05-06
[12sp2] scripts: fix yaml loader warning

liangxin1300:
unittests.test_scripts.test_load_legacy ... /usr/lib/python3/dist-packages/crmsh/scripts.py:431: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@
 sudo: false
 language: python
 python:
-  - 2.6
   - 2.7
 
 install:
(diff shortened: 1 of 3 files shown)


clanig:
I can confirm that this patch fixes issues with the HAWK compatibility when using wizards.

The pipeline fails because a necessary archive to run the pipeline is not available anymore.

MalloZup:
I guess we can safely merge this

PR #585, opened 2020-05-10
SBD management improve

liangxin1300:
#### Motivation
We are facing several issues about sbd in crmsh lately.
It's might not a good way to add more codes without structure, and might cause more issues with small piece of code here and there.
It's better to manage sbd as a whole part.
#### Solution
Create a class `SBDManager`, and a set of methods to manage sbd configuration and services
#### Advantages
Treats sbd as a whole part
Specific methods are easy to find, understand and test
#### Changes

- Create class `SBDManager`
- `SBDManager.sbd_init` to get sbd device, initialize and write configure file
- `SBDManager.manage_sbd_service` to enable or disable sbd service
- `SBDManager.configure_sbd_resource` to configure stonith-sbd resource and stonith-enabled property
- `SBDManager.join_sbd` to check and verify sbd device on join process
- Make `-s`(for sbd device) option and `-S`(for diskless sbd) option mutual
- Unit test for `SBDManager`
- Functional test `bootstrap_sbd.feature` which can only running on local environment(since it seems impossible to configure watchdog inside container)

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -60,7 +60,6 @@ class Context(object):
         self.yes_to_all = None
         self.template = None
         self.cluster_name = None
-        self.diskless_sbd = None
         self.watchdog = None
         self.nic = None
         self.unicast = None
@@ -76,7 +75,6 @@ class Context(object):
         self.qdevice_heuristics = None
         self.qdevice_heuristics_mode = None
         self.shared_device = None
-        self.sbd_device = None
         self.ocfs2_device = None
         self.cluster_node = None
         self.force = None
@@ -85,6 +83,9 @@ class Context(object):
         self.tickets = None
         self.ip_address = None
         self.ip_network = None
+        self.sbd_manager = None
+        self.sbd_devices = None
+        self.diskless_sbd = None
 
     @classmethod
     def set_context(cls, options):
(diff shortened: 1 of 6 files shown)


crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+    """
+    Class SBDManager is for sbd configuration and sbd service management
+    """
+    sbd_config_dict = {
+            "SBD_PACEMAKER": "yes",
+            "SBD_STARTMODE": "always",
+            "SBD_DELAY_START": "no"
+            }
+
+    def __init__(self, sbd_device=None, diskless_sbd=False):
```
arbulu89, line 122: Maybe better to call `sbd_devices`, as the value can be a list

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+                error("Can't create stonith-sbd primitive")
+            if not invoke("crm configure property stonith-enabled=true"):
+                error("Can't enable STONITH for SBD")
+        elif self.diskless_sbd:
+            if not invoke("crm configure property stonith-enabled=true stonith-watchdog-timeout=5s"):
+                error("Can't enable STONITH for diskless SBD")
+
+    def _get_sbd_device(self):
+        if self.sbd_device:
+            self.sbd_device = self._parse_sbd_device()
```
arbulu89, line 184: I don't know if it's a good idea to change the value of `sbd_device`, it might create issues. We should maybe create other parameter to store the values after parsing. What do you think?

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+            self.sbd_config_dict["SBD_DEVICE"] = ';'.join(self.sbd_device)
+        utils.sysconfig_set(SYSCONFIG_SBD, **self.sbd_config_dict)
+        csync2_update(SYSCONFIG_SBD)
+
+    def _get_sbd_device_interactive(self):
+        if _context.yes_to_all:
+            warn("Not configuring SBD (%s left untouched)." % (SYSCONFIG_SBD))
+            return
+
+        status("""
```
arbulu89, line 210: Put this string in a constant for this Class, it will be better

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+            warn("Not configuring SBD - STONITH will be disabled.")
+            return
+
+        configured_dev = self.configured_sbd_device()
+        if configured_dev and not confirm("SBD is already configured to use {} - overwrite?".format(configured_dev)):
+            return
+
+        dev_looks_sane = False
+        while not dev_looks_sane:
+            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none", use ";" as separator for multi path', r'none|\/.*')
```
arbulu89, line 231: Maybe we should add an improved message like `... or "none" for diskless sbd, use ";" ...` . What do you think?

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+            if dev == "none":
+                self.diskless_sbd = True
+                return
+            dev_list = dev.strip(';').split(';')
```
arbulu89, line 235: I think we can somehow use `_verify_sbd_device` method here, the code is exactly the same. Not just copy/pasting, what with some minor changes it might be done

crmsh/bootstrap.py
```diff
@@ -1375,136 +1536,14 @@ def sysconfig_comment_out(scfile, key):
...
-    utils.sysconfig_set(SYSCONFIG_SBD,
-                        SBD_PACEMAKER="yes",
-                        SBD_STARTMODE="always",
-                        SBD_DELAY_START="no",
-                        SBD_WATCHDOG_DEV=detect_watchdog_device())
-    csync2_update(SYSCONFIG_SBD)
-    status_done()
-
-
 def init_sbd():
```
arbulu89, line 1539: I think we can remove this method, right?

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
 
+    def init_sbd_manager(self):
+        self.sbd_manager = SBDManager(self.sbd_device, self.diskless_sbd)
+
+
+class SBDManager(object):
+    """
+    Class SBDManager is for sbd configuration and sbd service management
+    """
+    sbd_config_dict = {
```
arbulu89, line 116: This dict doesn't look like follows PEP8 indentation rules

crmsh/bootstrap.py
```diff
@@ -2016,10 +2047,8 @@ def update_nodeid(nodeid, node=None):
         csync2_update(corosync.conf())
         invoke("ssh -o StrictHostKeyChecking=no root@{} corosync-cfgtool -R".format(seed_host))
 
-    # if no SBD devices are configured,
-    # check the existing cluster if the sbd service is enabled
-    if not configured_sbd_device() and invoke("ssh -o StrictHostKeyChecking=no root@{} systemctl is-enabled sbd.service".format(seed_host)):
-        _context.diskless_sbd = True
+    if invoke("ssh -o StrictHostKeyChecking=no root@{} systemctl is-enabled sbd.service".format(seed_host)):
```
arbulu89, line 2050: Why don't you put this piece of code in the `SBDManager`? It totally belongs to sbd

crmsh/bootstrap.py
arbulu89, line 116: This variable is not so properly used. You declare here as constant or class method, but you change it for each instance. That's not a good idea.  If you are using for a particular object instance class (what you do right now), just declare this method like the others, and not as class method

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+        self._check_environment()
+        self._get_sbd_device()
+        if not self.sbd_device and not self.diskless_sbd:
+            return
+        self._initialize_sbd()
+        self._update_configuration()
+        # If process work through here, consider it's ready for enable service
+        self._sbd_service_flag = True
+
+    def configured_sbd_device(self):
```
arbulu89, line 159: I would change the name of this method to `get_sbd_configuration`

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
                 cmds=self.qdevice_heuristics,
                 mode=self.qdevice_heuristics_mode)
 
+    def init_sbd_manager(self):
+        self.sbd_manager = SBDManager(self.sbd_device, self.diskless_sbd)
+
+
+class SBDManager(object):
+    """
+    Class SBDManager is for sbd configuration and sbd service management
```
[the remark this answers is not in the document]
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
```diff
@@ -104,6 +105,179 @@ def init_qdevice(self):
...
+                        dev_looks_sane = True
+                    else:
+                        dev_looks_sane = False
+                        break
+
+        self.sbd_device = dev_list
+
+    def _check_environment(self):
+        if not check_watchdog():
+            error("Watchdog device must be configured if want to use SBD!")
```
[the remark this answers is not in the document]
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: Currently, we can't do that. `init_xxx` was a format for `stage`, used by `globals()["init_" + stage]()`

crmsh/bootstrap.py
liangxin1300: I'm not very clear here @arbulu89 , could you give me an example? Thanks!

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: OK, Changed

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: @arbulu89 , I put `sbd_config_dict` into function `_update_configuration` as a internal variables, what do you think of this change?

crmsh/bootstrap.py
liangxin1300: I put this into `detect_remote_sbd_service_status` method

crmsh/bootstrap.py
liangxin1300: Changed, reuse `_verify_sbd_device`

crmsh/bootstrap.py
liangxin1300: Trying to change this piece of code like below:
```
def _get_sbd_device(self):
        """
        Get sbd device from options or interactive mode
        """
        dev_list = []
        if self.sbd_devices:
            dev_list = self._parse_sbd_device()
            self._verify_sbd_device(dev_list)
        elif not self.diskless_sbd:
            dev_list = self._get_sbd_device_interactive()
        self.sbd_devices = dev_list
```

liangxin1300:
@arbulu89 Thanks a lot! Your suggestions are really helpful for me!

> 
> * Add a docstring to the methods please (to all of them). It's not a big effort but makes easier to understand the purpose of the method to newcomers (like me hehe)

Changed

> * I think the methods inside `SBDManager` are strangely ordered. Put in the top methods like `_check_environment` and `_parse_sbd_device` as they are used by `_sbd_init` (I would put them before that method). Otherwise we need to be scrolling up and down all the time.

I've adjusted the order of methods.

> * I have seen many methods that don't use anything of the instance (`self` parameters). I would convert those methods to `Classmethod` or `staticmethod`. It makes sense

I've converted those methods to `staticmethod`.
Thank you!

> 
> PD: With these changes the code looks much cleaner. Eventually, we should do the same for all the methods in the init and join processes, to have a neat code

crmsh/bootstrap.py
arbulu89: What I don't like is that the value of `self.sbd_devices` changes in format during the usage of this instance. You might have `sbd1;sbd2;sbd3` at the beginning and suddenly you have `['sbd', 'sbd2', 'sbd3']`. So it's a bit confusing. Usually doing this changes it's a bit risky, as you use the variable different in each situation.
That's why I suggested to have 2 variables like `self.sbd_devices_input` where you have always what the user pases, and `self._sbd_devices` after the transformation.
What do you think?

crmsh/bootstrap.py
arbulu89: Ok, I see. Thanks for the clarification

crmsh/bootstrap.py
```diff
@@ -104,6 +105,201 @@ def init_qdevice(self):
...
 
+    def init_sbd_manager(self):
+        self.sbd_manager = SBDManager(self.sbd_devices, self.diskless_sbd)
+
+
+class SBDManager(object):
+    """
+    Class to manage sbd configuration and services
+    """
+    sbd_status_description = """
```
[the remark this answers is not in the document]
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: OK, changed as `self.sbd_devices_input` and `self._sbd_devices`

crmsh/bootstrap.py
```diff
@@ -104,6 +105,201 @@ def init_qdevice(self):
...
+
+        dev_list = []
+        dev_looks_sane = False
+        while not dev_looks_sane:
+            dev = prompt_for_string('Path to storage device (e.g. /dev/disk/by-id/...), or "none" for diskless sbd, use ";" as separator for multi path', r'none|\/.*')
+            if dev == "none":
+                self.diskless_sbd = True
+                return
+            dev_list = dev.strip(';').split(';')
+            self._verify_sbd_device(dev_list)
```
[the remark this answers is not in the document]
liangxin1300: Changed as it can continue again with an error message

crmsh/bootstrap.py
```diff
@@ -105,6 +106,225 @@ def init_qdevice(self):
...
+        Get sbd device on interactive mode
+        """
+        if _context.yes_to_all:
+            warn("Not configuring SBD (%s left untouched)." % (SYSCONFIG_SBD))
+            return
+
+        status(self.SBD_STATUS_DESCRIPTION)
+
+        if not confirm("Do you wish to use SBD?"):
+            warn("Not configuring SBD - STONITH will be disabled.")
```
gao-yan, line 187: I don't think we are going to turn off `stonith-enabled` cluster property in this case, right? To avoid confusions, we could  probably say something like  "Not configuring SBD - Fencing won't work without fencing resources being configured ".

crmsh/bootstrap.py
```diff
@@ -105,6 +106,225 @@ def init_qdevice(self):
...
+            if not invoke("crm configure property stonith-enabled=true stonith-watchdog-timeout=5s"):
+                error("Can't enable STONITH for diskless SBD")
+
+    def join_sbd(self, peer_host):
+        """
+        Function join_sbd running on join process only
+        On joining process, check whether peer node has enabled sbd.service
+        If so, check prerequisites of SBD and verify sbd device on join node
+        """
+        if not os.path.exists(SYSCONFIG_SBD):
```
gao-yan, line 317: Just wondering,  does it mean /etc/sysconfig/sbd is supposed to have already been synchronized to this joining node if there's any on the first node?

PR #587, opened 2020-05-12
Low: crmsh: parallax slurp doesn't create an extra dir

aleksei-burlakov:
DON'T MERGE.

@liangxin1300 , @arbulu89 please have a look.

This PR is made in favor of one of the qdevice task that is called: Remove directory left by parallax.slurp

It doesn't remove the folder it simply copies files without creating an extra directory.

I can see that the parallax.parallax_slurp in the crmsh/corosync.py is used for only one node, so I thought about the solution in this PR.

### DIFF
--- a/crmsh/parallax.py
+++ b/crmsh/parallax.py
@@ -90,6 +90,32 @@ def parallax_slurp(nodes, localdir, filename, askpass=False, ssh_options=None):
     return p.slurp()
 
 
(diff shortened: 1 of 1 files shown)


aleksei-burlakov:
@arbulu89 there is indeed a line missing. I ammended the PR. Yes, it's aimed to copy directly to the localdir instead of the localdir/node_ip.

PR #591, opened 2020-05-22
Give suggestions when init node doesn't have crmsh ssh key

liangxin1300:
When init node doesn't have crmsh ssh key `/root/.ssh/id_rsa.crmsh`(might be from an old version crmsh), join process will fail to fetch that key, on this time, crmsh give suggestions about how to handle this scenarios

### DIFF
--- a/crmsh/constants.py
+++ b/crmsh/constants.py
@@ -486,4 +486,11 @@ SSH_KEY_CRMSH = "/root/.ssh/id_rsa.crmsh"
 SSH_KEY_CRMSH_TAG = "crmsh-generated"
 SSH_WITH_KEY = "ssh -i {}".format(SSH_KEY_CRMSH)
 SCP_WITH_KEY = "scp -i {}".format(SSH_KEY_CRMSH)
+KEY_NOT_EXIST_HINTS = """{key} couldn't be fetch from {peer}.
+This happens if the cluster was created with an older version of crmsh or the keys have been removed.
+Please, run the next commands on {peer} with admin rights:
+  ssh-keygen -q -f /root/.ssh/id_rsa.crmsh -C 'crmsh-generated' -N ''
+  cat /root/.ssh/id_rsa.crmsh.pub  >> /root/.ssh/authorized_keys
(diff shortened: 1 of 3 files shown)


crmsh/constants.py
```diff
@@ -486,4 +486,11 @@
 SSH_KEY_CRMSH_TAG = "crmsh-generated"
 SSH_WITH_KEY = "ssh -i {}".format(SSH_KEY_CRMSH)
 SCP_WITH_KEY = "scp -i {}".format(SSH_KEY_CRMSH)
+KEY_NOT_EXIST_HINTS = """{key} couldn't be fetch from {peer}.
+This happens if the cluster was created with an older version of crmsh or the keys have been removed.
+Please, run the next commands on {peer}:
```
arbulu89, line 491: Maybe we need to specify that these commands must be executed with admin rights (for us is obvious, but who knows). Maybe something like:
```
Please, run the next commands on {peer} with admin rights:
```

liangxin1300:
> Thanks @liangxin1300
> One question about this change. How is the user usage here? This is what I assume.
> 
> 1. The user has a cluster created with an old crmsh, so the id_rsa.crmsh is not there
> 2. The user runs the `join` process with a new crmsh.
> 3. If any other ssh keys is not configure, this process will ask for a password
> 4. The user sets the password, and as the `id_rsa.crmsh` doesn't exist, it raises this error, right?
> 
> So, now the user runs the command asked in the error message, and the `join` process again. In this case, the password will be asked again, but this time, the key will be retrieved and now on everything goes fine.
> 

Yes, that is exactly the same scenarios!

> From me side it's ok. What do you think @diegoakechi , @gao-yan ? I don't see any better solution than this

crmsh/constants.py
liangxin1300: Changed. Thanks! @arbulu89

zzhou1:
Thx! This approach is better than documentation whhich could be blind to the end users. The future ideal code refactoring could only ask once for the same password, and eventually removes this manual intervention.

PR #594, opened 2020-05-26
Test: bootstrap_bugs: don't overwrite the ssh key when init-ing

aleksei-burlakov:
By specifying only the -y key in the "crm cluster init -y" the
ssh key may be overwritten and this will make the test fail.
The --no-overwrite-sshkey flag will prevent it from this.


aleksei-burlakov:
The --no-overwrite-sshkey is not more in the master branch.

PR #595, opened 2020-05-27
Low: cibconfig: Avoid adding the ID attribute to select_* nodes

petrpavlu:
Alert-specific XML nodes `select_fencing`, `select_nodes` and `select_resources` should not have any ID attributes added. The attribute is not valid in this context according to the alert schema.

```
# crm configure alert snmp_alert "/var/lib/pacemaker/alert_snmp.sh" select fencing nodes resources to "192.168.0.1"
ERROR: Relax-NG validity error : Extra element alerts in interleave
       alerts (7): Element configuration failed to validate content
   error: main: CIB did not pass DTD/schema validation
Errors found during check: config not valid
Do you still want to commit (y/n)?
```

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -687,7 +687,11 @@ def fix_node_ids(node, oldnode):
         'alerts': 'alert',
         }
 
-    idless = set(['operations', 'fencing-topology', 'network', 'docker', 'rkt', 'storage', 'select', 'select_attributes'])
+    idless = set([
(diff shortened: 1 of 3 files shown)


liangxin1300:
@petrpavlu Could you please backport this commit into `crmsh-4.1` branch when you have time?
Thank you:)

petrpavlu:
Thanks @liangxin1300 for the review and merge. Cherry-pick to 4.1 is now at #598.

PR #597, opened 2020-06-08
Revert PR#575 -- Fix: bootstrap: crmsh use its own specific ssh key(bsc#1169581)

liangxin1300:
Consider #575 will cause compatibility problem with old version of crmsh, we decided to revert it.


arbulu89:
Thanks @liangxin1300 ,
We will try to implement this feature in the future with a better planning!

PR #598, opened 2020-06-08
[crmsh-4.1] Low: cibconfig: Avoid adding the ID attribute to select_* nodes

petrpavlu:
Backport of #595.


liangxin1300:
Thank you! @petrpavlu

PR #607, opened 2020-06-28
[crmsh-4.1] bootstrap process will not overwrite existing ssh keys

liangxin1300:
backport #603


liangxin1300:
> Thanks @liangxin1300 . As far as I see the changes are exactly the same as in the master branch, right?

Yes

PR #608, opened 2020-07-02
Qdevice config process improve

liangxin1300:
The work of this pull request was based on #540 and #464
Changes include:
- More reasonable naming for variables
- More function docstrings
- Move function to more reasonable location, like:
  * change `bootstrap.add_nodelist_from_cmaptool` to `corosync.add_nodelist_from_cmaptool`
- Create functions to integrate similar functions inside one, like:
  * create `QDevice.certificate_process_on_init` and `QDevice.certificate_process_on_join` methods to integrate certification related fuctions
- Change big function to small one, more easier for unit test, like:
  * `bootstrap.init_qdevice`
- Refactor functions:
  * Make `utils.service_is_enabled`, `utils.service_is_active` and `utils.package_is_installed` can be executed on remote node
- Create `utils.cluster_run_cmd` function to avoid using `crm cluster run` directly in code
- Simplify container deploy bash script `docker_scripts.sh` for functional test
- Complete unit test and functional test
- Append functional test for qdevice running on multi node and running on ipv6

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -106,6 +106,18 @@ def query_qnetd_status():
         print(utils.to_ascii(qnetd_result_stdout))
 
 
+def add_nodelist_from_cmaptool():
+    for nodeid, iplist in utils.get_nodeinfo_from_cmaptool().items():
+        try:
+            add_node_ucast(iplist, nodeid)
+        except IPAlreadyConfiguredError:
+            continue
+
+
+def is_unicast():
+    return get_value("totem.transport") == "udpu"
+
+
 _tCOMMENT = 0
 _tBEGIN = 1
 _tEND = 2
@@ -130,7 +142,9 @@ class Token(object):
 
 
 class QDevice(object):
-    '''
+    """
+    Class to manage qdevice configuration and services
+
     Whole certification process:
     For init
     Step 1:  init_db_on_qnetd
@@ -150,7 +164,7 @@ class QDevice(object):
     Step 2:  init_db_on_local
     Step 3:  fetch_p12_from_cluster
     Step 4:  import_p12_on_local
-    '''
+    """
     qnetd_service = "corosync-qnetd.service"
     qnetd_cacert_filename = "qnetd-cacert.crt"
     qdevice_crq_filename = "qdevice-net-node.crq"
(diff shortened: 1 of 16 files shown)


crmsh/corosync.py
```diff
@@ -167,56 +184,101 @@ def __init__(self, ip, port=5403, algo="ffsplit", tie_breaker="lowest",
...
+        Validate qdevice related options
+        """
+        if not utils.package_is_installed("corosync-qdevice"):
             raise ValueError("Package \"corosync-qdevice\" not installed on this node")
-        if self.ip == utils.this_node() or self.ip in interfaces_inst.ip_list:
+
+        try:
+            # socket.getaddrinfo works for both ipv4 and ipv6 address
+            res = socket.getaddrinfo(self.qnetd_addr, None)
+            self.qnetd_ip = res[0][-1][0]
```
arbulu89, line 262: Could you put an example of this `res` object? To know what `[0][-1][0]` means

crmsh/corosync.py
```diff
@@ -167,56 +184,101 @@ def __init__(self, ip, port=5403, algo="ffsplit", tie_breaker="lowest",
         self.askpass = False
         self.cmds = cmds
         self.mode = mode
+        self.qnetd_ip = None
```
arbulu89, line 187: What is the difference between `qnetd_addr` and `qnetd_ip`. We need to document what all of these variables are, some of them are not obvious enough to let them undocumented.

In fact, as far as I see, `self.qnetd_ip` is used only in 1 method. Maybe we could use it as a local variable. Os am I missing something?

arbulu89 (approved):
@liangxin1300 Great job!

I cannot find things that look wrong, in general the code looks fine, and better documented and tested.
The unique way of knowing if something is broken is running tests, so if everything works, I think we can go ahead and merge.

I have pointed out some code smells that we could recheck.

crmsh/corosync.py
liangxin1300: I will put more comments here

crmsh/corosync.py
liangxin1300: Thanks
I changed `self.qnetd_ip` as local variable

PR #610, opened 2020-07-07
Low: hb_report: Fix collecting of binary data

liangxin1300:
When creating a resulting archive in a slave collector ('hb_report
__slave'), the code tried to decode output produced by the tar utility
as a UTF-8 string (crmutils.get_stdout_stderr() -> to_ascii()). This
corrupted any binary files in the archive, such as gzipped pengine
files.

The patch fixes the problem by obtaining the output from the tar utility
as a bytes object and writing it in the repr() form on stdout of the
slave collector. Similarly, the receiving side is updated to accept this
format.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -725,7 +725,7 @@ def pipe_cmd_nosudo(cmd):
     return rc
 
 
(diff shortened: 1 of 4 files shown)


liangxin1300:
Thanks @arbulu89 !

PR #612, opened 2020-07-22
Low: ui_corosync: copy ssh key to qnetd while detect need password

liangxin1300:
While setup qdevice on init node, init node will copy its ssh key to qnetd node to setup passwordless ssh login with qnetd node.
But after merging PR #603, each cluster node has its own ssh key, so it's possible that query qnetd status using `crm corosync status qnetd` on other cluster nodes will failed.
The solution is to copy the ssh key to qnetd node if detecting need password while the first time querying qnetd status on the other node

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -36,6 +36,73 @@ def quorumtool(*args):
     return utils.get_stdout(['corosync-quorumtool'] + list(args), shell=False)
 
 
+def query_status(status_type):
+    """
+    Query status of corosync
+
+    Possible types could be ring/quorum/qnetd
+    """
+    if status_type == "ring":
+        query_ring_status()
+    elif status_type == "quorum":
+        query_quorum_status()
+    elif status_type == "qnetd":
+        query_qnetd_status()
+    else:
+        raise ValueError("Wrong type \"{}\" to query status".format(status_type))
+
+
+def query_ring_status():
+    """
+    Query corosync ring status
+    """
+    rc, out, err = utils.get_stdout_stderr("corosync-cfgtool -s")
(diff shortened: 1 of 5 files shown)


liangxin1300:
> @liangxin1300 Actually, I'm wondering if this is the correct place of authorizing the key. The key should be authorized in the moment where the qdevice node is connected to the qnetd server. Is this that moment?

If setup qdevice via command `crm cluster init --qnetd-hostname`, the ssh key will be authorized just between the init node and the qnetd node, not other nodes in the cluster.

If user query qnetd status via `crm corosync status qnetd` on the other node, I think this is the right place to authorized the key

arbulu89:
> > @liangxin1300 Actually, I'm wondering if this is the correct place of authorizing the key. The key should be authorized in the moment where the qdevice node is connected to the qnetd server. Is this that moment?
> 
> If setup qdevice via command `crm cluster init --qnetd-hostname`, the ssh key will be authorized just between the init node and the qnetd node, not other nodes in the cluster.
> 
> If user query qnetd status via `crm corosync status qnetd` on the other node, I think this is the right place to authorized the key

@liangxin1300 Ok, let's go forward with this solution. After chatting, we agreed that the unique thing would be to change where all of the `status` operations are coded, that should belong to `corosync.py`. Besides that, the implementation is good.

liangxin1300:
> Excellent Xin! This looks much better than before!
> Just one doubt. Why do you use `get_stdout_stderr` instead of `corosync.quorumtool('-s')[1]`? To get the stderr output as well?

Yes, I think what if quorumtool command crashed itself, so I can catch the error

PR #613, opened 2020-07-28
Low: corosync: handle the return code of corosync-quorumtool correctly

liangxin1300:
Since the return code 2 of command `corosync-quorumtool -s` means no problem appeared but node is not quorate, we should adopt to handle that return code when quorum lost.
Additionally, behave test in `qdevice_usercase.feature` has changed to use command `crm corosync status quorum` to make sure this scenarios been tested while quorum lost

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -66,12 +66,15 @@ def query_ring_status():
 def query_quorum_status():
     """
     Query corosync quorum status
+
     """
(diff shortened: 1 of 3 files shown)


liangxin1300:
> @liangxin1300 This change is not involving any change in our unit tests?

No, because in the unit test for test `query_quorum_status` function, I only test two return code 1 and 0, should I update return code 2 in the UT?

PR #617, opened 2020-08-12
Adjust sbd configure process in stage

liangxin1300:
#### To reproduce the problem
```
crm cluster init ssh -y
crm cluster init csync2 -y
crm cluster init corosync -y
crm cluster init sbd -s /dev/sdb1 -y (or diskless mode, crm cluster init sbd -S -y)
crm cluster init cluster -y
```
Then sbd.service would not start

#### Root cause
Configure sbd and enable sbd.service are in different steps, now combine them together when calling `sbd_init`

#### Other changes
* Dev: unittest: adjust unit test cases for sbd config process changes
* Dev: behave: adjust functional test for sbd config process changes
* Fix: bootstrap: revert ssh_merge function for compatibility(bsc#1175057)

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -166,7 +166,6 @@ Configure SBD:
         """
         self.sbd_devices_input = sbd_devices
         self.diskless_sbd = diskless_sbd
-        self._sbd_service_flag = False
         self._sbd_devices = None
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


crmsh/bootstrap.py
```diff
@@ -1876,6 +1865,48 @@ def join_csync2(seed_host):
     status_done()
 
 
+def join_ssh_merge(_cluster_node):
```
arbulu89, line 1868: I guess that this change recovers some older code, but the name of the method doesn't look really accurate.
Let's skip this issue anyway XD

arbulu89 (approved):
@liangxin1300 Thank you for the changes.
My unique concern is that we are fixing 2 different issues in the same pull request. It's true that they are reported in the same ticket though.
Anyway, I think it's good enough.
Thank you for adding more test cases, let's see if we can continue adding more coverage!

PR #635, opened 2020-09-22
fix on_fail should be on-fail

MalloZup:
see #bsc1176569 @liangxin1300 cc


liangxin1300 (approved):
Thanks @MalloZup 
It looks good to me

PR #639, opened 2020-10-03
Fix: print error message when invoke has error output(bsc#1177023)

liangxin1300:


MalloZup (approved):

zzhou1 (changes requested):
No rush to merge the code.

1) Again, not intuitive enough in my view, especially if I force my self to be a not-so-skillful-sysadmin. Even though bumping the **internal error** message is more informative slightly. Well, internal error often means **suprise**, and is not appropriate in this **normal** situation in Public Cloud.

2) Kind reminder to evaluate the side effect in this generl code path. I wonder if stderr is badly presented in other error situations. 

I think I incline to deal the error only against "Configuring SSH passwordless". To be more user friendly, 

a) in such "normal" situation of Public Cloud, I naturlly expect a "normal" message to match the context overall

b) Even a message to inform user to check "/var/log/ha-cluster-bootstrap.log" is better than dumping the internal error message, in my view.

zzhou1:
> 2\. Kind reminder to evaluate the side effect in this generl code path. I wonder if stderr is badly presented in other error situations.

I ran into one of the example: csync2 ERROR might generate quite amount of stderr message, and looks not approriate in the user's terminal session.

PR #557, opened 2020-03-28
Refactor crm report

liangxin1300:
# Overview
Since customer complain the performance and to fix existing bugs, I'm pleased to create this PR, in general, changes include:
* More **faster**
* More **stronger** (100% unit test coverage; new added functional test)
* -b option
* Leveled debug messages
* Existing bugs fix
* Updated help and man page

# RPM link
If you have interests, here are rpm links:
**[For 12sp4-15sp1](https://build.opensuse.org/package/show/home:XinLiang/crmsh)**
**[For Tumbleweed](https://build.opensuse.org/package/show/home:XinLiang:branches:network:ha-clustering:Factory/crmsh)**
Any suggestions and feedback are welcome!

# Changes
## **More faster**
- Original
```
time hb_report 
real	0m11.038s
user	0m4.669s
sys	0m1.095s
```
- New
```
time hb_report 
real	0m4.551s
user	0m2.662s
sys	0m0.533s
```

#### Unit test: **100%** test coverage (original percentage was very low)

```
/app/hb_report/__init__.py                                   0      0   100%
/app/hb_report/collect.py                                  343      0   100%
/app/hb_report/const.py                                     34      0   100%
/app/hb_report/core.py                                     588      0   100%
/app/hb_report/utils.py                                    248      0   100%
```
#### Functional test(**new added**): 

  **[hb_report_normal](https://github.com/liangxin1300/crmsh/blob/20200202_Refactor_hb_report/test/features/hb_report_normal.feature)**, include scenarios:
    -  Run hb_report on empty environment
    -  Verify log file filter by time span
    -  Verify hb_report options

  **[hb_report_bugs](https://github.com/liangxin1300/crmsh/blob/20200202_Refactor_hb_report/test/features/hb_report_bugs.feature)**
    - Bug regression, include bsc#1148874, 1148873, 1067456, 1077553, 1106052, 1135696, 1130715, 1163581

## **-b option**
How long time in the past, before now ([1-9][0-9]*[YmdHM])
    Examples:
```
  # collect from 1 mounth ago
  hb_report -b 1m
  # collect from 12 days ago
  hb_report -b 12d
  # collect from 75 hours ago
  hb_report -b 75H
  # collect from 10 minutes ago
  hb_report -b 10M
```
## **Leveled debug messages**
`-v`: show basic debug messages indicate what files have been collected and where they are
`-vv`: besides above, show internal debug messages
```
# show basic debug messages
hb_report -v
[...]

PR #641, opened 2020-10-12
Fix: hb_report: collect corosync.log if it defined in config file(bsc#1148874)

liangxin1300:
## Problem
The hb_report result didn't include corosync.log

## Solution
Collect corosync log if it was defined in **logging.logfile** in corosync.conf and in the time span

## Functional test
``` Shell
  @clean
  Scenario: Collect corosync.log(bsc#1148874)
    When    Run "sed -i 's/\(\s*logfile:\s*\).*/\1\/var\/log\/cluster\/corosync.log/' /etc/corosync/corosync.conf" on "hanode1"
    And     Run "hb_report report" on "hanode1"
    And     Run "tar jxf report.tar.bz2" on "hanode1"
    Then    File "corosync.log" not in "report.tar.bz2"
    When    Run "rm -rf report.tar.gz report" on "hanode1"

    When    Run "sed -i 's/\(\s*to_logfile:\s*\).*/\1yes/' /etc/corosync/corosync.conf" on "hanode1"
    And     Run "crm cluster stop" on "hanode1"
    And     Run "crm cluster start" on "hanode1"
    And     Run "hb_report report" on "hanode1"
    And     Run "tar jxf report.tar.bz2" on "hanode1"
    Then    File "corosync.log" in "report.tar.bz2"
    When    Run "rm -rf report.tar.gz report" on "hanode1"
```

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -27,6 +27,7 @@ import constants
 import crmsh.config
 from crmsh import msg as crmmsg
 from crmsh import utils as crmutils
+from crmsh import corosync
 
 
 class Tempfile(object):
@@ -312,7 +313,13 @@ def collect_info():
     for p in process_list:
         p.join()
 
-    for l in constants.EXTRA_LOGS.split():
+    logfile_list = []
+    corosync_log = corosync.get_value('logging.logfile')
+    if corosync_log:
+        logfile_list.append(corosync_log)
+    logfile_list += constants.EXTRA_LOGS.split()
+
+    for l in logfile_list:
         if not os.path.isfile(l):
             continue
         if l == constants.HA_LOG and l != constants.HALOG_F:
--- a/test/features/hb_report_bugs.feature
+++ b/test/features/hb_report_bugs.feature
@@ -43,6 +43,22 @@ Feature: hb_report functional test for verifying bugs
     When    Run "rm -rf report1.tar.gz report1" on "hanode1"
 
   @clean
+  Scenario: Collect corosync.log(bsc#1148874)
+    When    Run "sed -i 's/\(\s*logfile:\s*\).*/\1\/var\/log\/cluster\/corosync.log/' /etc/corosync/corosync.conf" on "hanode1"
+    And     Run "hb_report report" on "hanode1"
+    And     Run "tar jxf report.tar.bz2" on "hanode1"
+    Then    File "corosync.log" not in "report.tar.bz2"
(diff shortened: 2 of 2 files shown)


nick-wang:
@liangxin1300 Is that possible to dynamically check and add `logging.logfile` file to `constants.EXTRA_LOGS`?
Since besides `collect_info()`, i can see some other place use `constants.EXTRA_LOGS`. For example, find/get/check log. I am not sure whether these spots need `logging.logfile` file as well.

liangxin1300:
> @liangxin1300 Is that possible to dynamically check and add `logging.logfile` file to `constants.EXTRA_LOGS`?
> Since besides `collect_info()`, i can see some other place use `constants.EXTRA_LOGS`. For example, find/get/check log. I am not sure whether these spots need `logging.logfile` file as well.

Thanks nick,
I would say I don't want to make more efforts for hb_report below 15sp3, except fix bugs
Things(I mean the code structures) will be totally different in https://github.com/ClusterLabs/crmsh/pull/557, which will be backpoted to old branches eventually

nick-wang:
> I would say I don't want to make more efforts for hb_report below 15sp3, except fix bugs
> Things(I mean the code structures) will be totally different in #557, which will be backpoted to old branches eventually

Ok, if you say so as a temporarily fix. I am fine with the change, cause it may be incomplete but won't lead any issue.

nick-wang (approved):
The code to collect_info() is OK. Though dynamically change `constants.EXTRA_LOGS` is better, since it is also used in other places like get/check/find logs.

Per Xin, it is just a temporarily fix. The code will be completely change after merged, so i am fine with the change.

liangxin1300:
> The code to collect_info() is OK. Though dynamically change `constants.EXTRA_LOGS` is better, since it is also used in other places like get/check/find logs.
> 
> Per Xin, it is just a temporarily fix. The code will be completely change after merged, so i am fine with the change.

Thanks for your suggestions!

hb_report/utillib.py
```diff
@@ -312,7 +313,13 @@ def collect_info():
     for p in process_list:
         p.join()
 
-    for l in constants.EXTRA_LOGS.split():
+    logfile_list = []
+    corosync_log = corosync.get_value('logging.logfile')
+    if corosync_log and os.path.isfile(corosync_log):
```
nick-wang, line 318: os.path.isfile(corosync_log) is useless, since it will be check again immediately in line 323

liangxin1300: Thanks for pointing this!

PR #642, opened 2020-10-13
[crmsh-4.1] Fix bug of remove node(bsc#1165644) 

liangxin1300:
backport #537

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -76,6 +76,7 @@ class Context(object):
         self.shared_device = None
         self.ocfs2_device = None
         self.cluster_node = None
+        self.cluster_node_ip = None
         self.force = None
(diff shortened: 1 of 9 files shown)


arbulu89 (approved):
Thanks @liangxin1300 
This branch already has `qdevice` implementation, right?
Otherwise some code doesn't make sense (when you do some operations with the algorithms)

liangxin1300:
> Thanks @liangxin1300
> This branch already has `qdevice` implementation, right?
> Otherwise some code doesn't make sense (when you do some operations with the algorithms)

@arbulu89  Branch crmsh-4.1 has some basic qdevice configure options as a long time, not fully implemented and not recommended to use this branch to configure qdevice(also for SLE12SP4 and SP5, corosync's verson not support qdevice)

PR #644, opened 2020-10-14
Show help message from command line if that subcommand use argparse

liangxin1300:
## Problem

1. For some subcommands which were using argparse, help messages exist both in crm.8.adoc and code, it's easy to cause inconsistent since code in argparse might changed frequently
2. Run `crm cluster --help` will pop a warning: `WARNING: This command '--help' is deprecated, please use 'help'`, this is also inconsistent for the command syntax

## Solution
**For problem 1**: In `crm.8.adoc`, mark those subcommands using argparse as `From Code`, then calling `-h/--help`, display that output, make sure only one place has help messages
**For problem 2**: Add additional condition based on https://github.com/ClusterLabs/crmsh/pull/558, to check whether the command name was -h or --help, don't consider it as alias

## TEST rpm
https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_testing/crmsh

### DIFF
--- a/crmsh/help.py
+++ b/crmsh/help.py
@@ -27,7 +27,7 @@ Help for the level itself is like this:
 
 import os
 import re
-from .utils import page_string
+from .utils import page_string, get_stdout_stderr
 from .msg import common_err
 from . import config
 from . import clidisplay
@@ -100,6 +100,9 @@ class HelpEntry(object):
             prefix = helpfilter("(Redirected from `%s` to `%s`)\n" % self.alias_for)
 
         page_string(short_help + '\n' + prefix + long_help)
+
+    def set_long_help(self, long_help):
+        self.long = long_help
 
     def __str__(self):
         if self.long:
@@ -300,9 +303,15 @@ def _load_help():
 
     def parse_header(line):
         'returns a new entry'
-        entry = {'type': '', 'name': '', 'short': '', 'long': ''}
+        entry = {'type': '', 'name': '', 'short': '', 'long': '', "from_cli": False}
         line = line[2:-3]  # strip [[ and ]]\n
         info, short_help = line.split(',', 1)
+        # TODO see https://github.com/ClusterLabs/crmsh/pull/644
+        # This solution has shortcome to delete the content of adoc,
+        # which lose the static man page archive
+        if "From Code" in short_help:
+            short_help, _ = short_help.split(',')
+            entry['from_cli'] = True
         entry['short'] = short_help.strip()
         info = info.split('_')
         if info[0] == 'topics':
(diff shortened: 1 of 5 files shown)


nick-wang:
@liangxin1300 I think the logic of add/check "From code" do add the unnecessary complexity of the change. 
Since you need to parser the man page to check the "From code", which is you manually added in code. I would suggest manage a tuple list for the parameters that need ArgParser "from cli". Then use "crm {} {} --help" to check if it exist in the tuple list.

arbulu89:
@liangxin1300 Why don't we centralize the `help` messages in a unique place? `adoc` preferably. It looks like we are going to add a patch because we have a bad design

liangxin1300:
> @liangxin1300 Why don't we centralize the `help` messages in a unique place? `adoc` preferably. It looks like we are going to add a patch because we have a bad design

Hi @arbulu89 ,
I had such idea previously, like:

For bsc#1175982, to keep content consistent between -h/--help of argparse and help of crmsh subcommand,
As far as I know, I think we should make extra efforts to do that
(let me take `crm cluster help init` as an example)
1. `crm cluster help init` will parse /usr/share/crmsh/crm.8.adoc, to load/extract and display `cluster init` section between the whole adoc file
2. crm.8.adoc was huge, I want to split it into some small ones(which content will from argparse), by using include syntax of Asciidoc, like, `include::sub_adoc/crm_cluster_init.adoc[]`, while `crm_cluster_init.adoc` was generated from `crm cluster init -h`(argparse)
3. There was no suitable tools to convert help output of argparse to .adoc file, https://github.com/klali/help2adoc might be an alternative one, but I should do some adjustments in perl to make sure the output looks pretty(it's painful process for me, I would like to rewrite a python version `help2adoc` for reuse)
4. When `help2adoc` was ready, run command like `help2adoc -e "crm cluster init -h"  > sub_adoc/crm_cluster_init.adoc`
5. Improve crmsh/help.py, to make sure crm cluster help init could parse `include::` syntax from asciidoc

To automatic the above process, we should push `help2adoc` to Factory, and use it in crmsh.spec, in package install stage, to generate manpage and sub_adoc/*.adoc files

liangxin1300:
@zzhou1 I suggest we merged this, if find better or more flexible solution, then we could improve

zzhou1:
I'm okay as the incomplete solution at the moment, which is much better than the inconsistent mess. I can see another shortcome to delete the content of adoc, which lose the static man page archive. I suggest to add TODO comments in the code to refer back to this PR for the further change in the future when we have the capcity, before we completely lose track of this context.

crmsh/help.py
```diff
@@ -300,9 +303,15 @@ def _load_help():
 
     def parse_header(line):
         'returns a new entry'
-        entry = {'type': '', 'name': '', 'short': '', 'long': ''}
+        entry = {'type': '', 'name': '', 'short': '', 'long': '', "from_cli": False}
         line = line[2:-3]  # strip [[ and ]]\n
         info, short_help = line.split(',', 1)
+        # TODO see https://github.com/ClusterLabs/crmsh/pull/644
+        # This solution has shortcome to delete the content of adoc,
+        # which lose the static man page archive
```
liangxin1300, line 311: Add `TODO` note here

zzhou1 (approved):

PR #645, opened 2020-10-16
Refactor service management codes

liangxin1300:
## Problem
1. Some service management related functions were here and there, and some of them were redundant
2. It will more convenient to make all of these functions running remotely or locally
## Solution
1. Create class `ServerManager` which has a set of service management functions
2. Create shortcuts for the classmethod to make sure the API wouldn't change

This PR would be small enough to backport to other branches

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1779,45 +1779,6 @@ def service_info(name):
     return None
 
 
-def start_service(service, enable=False, start=True):
-    """
-    Start (and optionally enable) systemd service
-    TODO: support other init systems
-    """
-    if enable:
-        _rc, _out, _err = get_stdout_stderr("systemctl -q enable {}".format(service))
-    if start:
-        rc, _out, _err = get_stdout_stderr("systemctl -q is-active {}".format(service))
-        if rc != 0:
-            rc, _out, err = get_stdout_stderr("systemctl -q start {}".format(service))
-            if rc != 0:
-                raise IOError("systemd failed to start {}: {}".format(service, err))
-
-
-def stop_service(service, disable=False, stop=True):
-    """
-    Stop (and optionally disable) systemd service
-    TODO: support other init systems
-    """
-    rc, err = 0, ""
-    if stop:
-        rc, _out, err = get_stdout_stderr("systemctl -q stop {}".format(service))
-    if disable:
-        _rc, _out, _err = get_stdout_stderr("systemctl -q disable {}".format(service))
-    if rc != 0:
-        raise IOError("systemd failed to stop {}: {}".format(service, err))
-
-
-def enable_service(service, start=False):
-    "Enable (and optionally start) system service"
-    start_service(service, enable=True, start=start)
-
-
-def disable_service(service, stop=False):
-    "Enable (and optionally stop) system service"
-    stop_service(service, disable=True, stop=stop)
-
-
 def running_on(resource):
     "returns list of node names where the given resource is running"
     rsc_locate = "crm_resource --resource '%s' --locate"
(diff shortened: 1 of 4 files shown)


crmsh/utils.py
```diff
@@ -2512,34 +2473,141 @@ def check_file_content_included(source_file, target_file):
...
+            rc, output, err = get_stdout_stderr(cmd)
+        if rc != 0 and err:
+            raise ValueError("Run \"{}\" error: {}".format(cmd, err))
+        return rc == 0, output
+
+    def is_available(self):
+        return self.service_name in self._do_action("list-unit-files")[1]
+
+    def is_enabled(self):
+        return self._do_action("is-enabled")[0]
```
arbulu89, line 2510: I guess we should use the strings from `ACTION_LIST`, right?
Instead of duplicating here

crmsh/utils.py
```diff
@@ -2512,34 +2473,141 @@ def check_file_content_included(source_file, target_file):
...
+    @classmethod
+    def disable_service(cls, name, remote_addr=None):
+        """
+        Disable service
+        """
+        inst = cls(name, remote_addr)
+        inst.disable()
+
+    @classmethod
+    def enable_start_service(cls, name, remote_addr=None):
```
arbulu89, line 2584: This name looks pretty strange.
I would only have `start_service` and add a parameter named `enabled=False`, or something like that. And if this parameters comes, just run the `enable` method.
What do you think?

crmsh/utils.py
```diff
@@ -2512,34 +2473,141 @@ def check_file_content_included(source_file, target_file):
...
+    def disable_stop_service(cls, name, remote_addr=None):
+        """
+        Disable and stop service
+        """
+        inst = cls(name, remote_addr)
+        inst.disable()
+        inst.stop()
+
+
+service_is_available = ServiceManager.service_is_available
```
arbulu89, line 2602: This looks a bit strange, but well. It's correct anyway

arbulu89 (changes requested):
@liangxin1300 In general the changes are good.
I suggest to use the service action names from the already created constant, and maybe re-do the the `enable_start_service` and `disable_stop_service`

crmsh/utils.py
liangxin1300: OK, changed

crmsh/utils.py
liangxin1300: Ok, changed as `enable` parameter in `start_service` function, and `disable` parameter in `stop_service` function

liangxin1300:
@arbulu89 Already changed! Thanks!

crmsh/utils.py
```diff
@@ -2512,34 +2473,125 @@ def check_file_content_included(source_file, target_file):
...
+            prompt_msg = "Run \"{}\" on {}".format(cmd, self.remote_addr)
+            rc, output, err = run_cmd_on_remote(cmd, self.remote_addr, prompt_msg)
+        else:
+            rc, output, err = get_stdout_stderr(cmd)
+        if rc != 0 and err:
+            raise ValueError("Run \"{}\" error: {}".format(cmd, err))
+        return rc == 0, output
+
+    def is_available(self):
+        return self.service_name in self._do_action(self.ACTION_LIST[-1])[1]
```
arbulu89, line 2507: This looks pretty ugly (`self.ACTION_LIST[-1]`). I would change the `ACTION_LIST` from being a list to be a map or named list. The code would be neater.

arbulu89 (approved):
@liangxin1300 Approved.
I still see room from improvement changing how the `ACTION_LIST` works.

PR #646, opened 2020-10-16
Convert score to kind for rsc_order configure

liangxin1300:
## Problem

1. In pacemaker 2.x, `score` syntax in rsc_order configure was deprecated
2. crmsh give error while the return message of crm_verify was a warning

## Solution
For problem 1, convert score to `kind`, that is, convert `INFINITY` and score greater than "0" to `Mandatory`, else, convert as `Optional`
For problem 2, give a warning if return message of crm_verify was a warning

## Test
For problem 1:
```Shell
When    Run "crm configure primitive d1 Dummy op monitor interval=10s" on "hanode1"
And     Run "crm configure primitive d2 Dummy op monitor interval=10s" on "hanode1"
And     Run "crm configure order o1 100: d1 d2" on "hanode1"
When    Run "crm configure show" on "hanode1"
Then    Expected "order o1 Mandatory: d1 d2" in stdout
```

### DIFF
--- a/crmsh/parse.py
+++ b/crmsh/parse.py
@@ -581,12 +581,15 @@ class BaseParser(object):
         else:
             self.err("Unknown date operation '%s', please upgrade crmsh" % (op))
 
-    def validate_score(self, score, noattr=False):
+    def validate_score(self, score, noattr=False, to_kind=False):
         if not noattr and score in olist(constants.score_types):
             return ["score", constants.score_types[score.lower()]]
         elif re.match("^[+-]?(inf(inity)?|INF(INITY)?|[0-9]+)$", score):
             score = re.sub("inf(inity)?|INF(INITY)?", "INFINITY", score)
-            return ["score", score]
+            if to_kind:
+                return ["kind", score_to_kind(score)]
+            else:
+                return ["score", score]
         if noattr:
             # orders have the special kind attribute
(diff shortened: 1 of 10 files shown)


crmsh/parse.py
```diff
@@ -1742,4 +1745,13 @@ def parse(s, comments=None):
...
         return False
 
+
+def score_to_kind(score):
+    """
+    Convert score to kind for rsc_order
+    """
+    kind = "Optional"
+    if score == "INFINITY" or re.search("^[1-9][0-9]*$", score):
+        kind = "Mandatory"
```
gao-yan, line 1755: Pacemaker treats a score `0` as `Optional`,  and technically actually anything else as `Mandatory`.

gao-yan (commented):
Generally looks good to me. Thanks for the nice work, @liangxin1300 !

crmsh/parse.py
gao-yan: @liangxin1300 What do you think of the above comment?  Do you think it makes sense to make the logic here simpler and consistent with pacemaker:
https://github.com/ClusterLabs/pacemaker/blob/master/lib/pacemaker/pcmk_sched_constraints.c#L158-L166
?

gao-yan (approved):
Thanks for the change.

PR #648, opened 2020-10-20
Stop hawk service when removing node

liangxin1300:
## Problem
crmsh should stop hawk.service when removing node

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -44,7 +44,7 @@ COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 RSA_PRIVATE_KEY = "/root/.ssh/id_rsa"
 RSA_PUBLIC_KEY = "/root/.ssh/id_rsa.pub"
 AUTHORIZED_KEYS_FILE = "/root/.ssh/authorized_keys"
-
+SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
 
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
 
@@ -2152,6 +2152,16 @@ def set_cluster_node_ip():
             break
 
 
+def stop_services(stop_list, remote_addr=None):
+    """
+    Stop cluster related service
+    """
+    for service in stop_list:
+        if utils.service_is_available(service, remote_addr=remote_addr):
+            status("Stopping the {}".format(service))
+            utils.stop_service(service, disable=True, remote_addr=remote_addr)
+
+
 def remove_node_from_cluster():
     """
     Remove node from running cluster and the corosync / pacemaker configuration.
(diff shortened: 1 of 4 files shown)


crmsh/bootstrap.py
```diff
@@ -2152,15 +2152,24 @@ def set_cluster_node_ip():
             break
 
 
+def stop_cluster_related_service(remote_addr=None):
```
MalloZup, line 2155: def stop_cluster_related_service(services, remote_addr=None):
    """
    Disable and stop cluster related service
    """
    for service in services:
        if utils.service_is_available(service, remote_addr=remote_addr):
            status("Stopping the {}".format(service))
            utils.stop_service(service, disable=True, remote_addr=remote_addr)

crmsh/bootstrap.py
```diff
@@ -2424,7 +2433,7 @@ def remove_self():
             error("Failed to remove this node from {}".format(othernode))
     else:
         # disable and stop cluster
-        utils.stop_service("corosync", disable=True)
+        stop_cluster_related_service()
```
MalloZup, line 2436: services = ["corosync-qdevice.service", "corosync.service", "hawk.service"]:
stop_cluster_related_service(services)

MalloZup (changes requested):
looks good
I would separate the Input from a function to make it more stateless and generic.


So in the example the function should not known about the services but they should be an arg

liangxin1300:
> looks good
> I would separate the Input from a function to make it more stateless and generic.
> 
> So in the example the function should not known about the services but they should be an arg

Thanks @MalloZup ! Already changed!

crmsh/bootstrap.py
```diff
@@ -2152,15 +2152,24 @@ def set_cluster_node_ip():
             break
 
 
+def stop_cluster_related_service(stop_list, remote_addr=None):
```
arbulu89, line 2155: As this works now as a generic method, maybe it should be renamed to `stop_services` and receive `service` as argument.

crmsh/bootstrap.py
```diff
@@ -44,7 +44,7 @@
 RSA_PRIVATE_KEY = "/root/.ssh/id_rsa"
 RSA_PUBLIC_KEY = "/root/.ssh/id_rsa.pub"
 AUTHORIZED_KEYS_FILE = "/root/.ssh/authorized_keys"
-
+SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
```
arbulu89, line 47: I hope this won't bring some issue during the backports as `corosync-qdevice.service` doesn't exist in some branches

arbulu89 (approved):
@liangxin1300 It looks good. I proposed some renaming of methods, but up to you if you like them or not.

crmsh/bootstrap.py
MalloZup: yep I agree with naming

crmsh/bootstrap.py
liangxin1300: I will notice to remove `corosync-qdevice.service` in other branches

MalloZup: we should then check if the service exists on the function name and don't errror. good catch

crmsh/bootstrap.py
liangxin1300: OK, already change to `stop_services`

crmsh/bootstrap.py
arbulu89: Maybe it's not even needed. The method already checks if the service is available. What happens if it's not? It fails? Or it just continues? If it continues, we would be fine

MalloZup: yep agree if it continues is no problem . is just we shouldn't raise exceptions

PR #657, opened 2020-10-22
Integrate preflight_check into crmsh

liangxin1300:
## Background
This PR was for integrating [cluster_preflight_check](https://github.com/liangxin1300/cluster_preflight_check) into crmsh

## Benefits:

- Consistency
- Easier to memorize for our customers
- Integrated under one root ("crm cluster")
- Less maintain job for  python-cluster-preflight-check package
- Easy to make tests under the existing crmsh test framework 

## Scope in this PR

- [ ] Integrate into crmsh (where and which sub command name)
- [x] No functionalities change
- [x] Reasonable refactors (reuse crmsh functions; using meaningful classes instead of functions)
- [ ] Good unit test coverage

#### Major flow
```
main.run
    setup_basic_context
    parse_argument
    setup_logging

    check.check  (-e/--env-check and -c/--cluster-check option, task.TaskCheck)
    kill_process (--kill-sbd/--kill-corosync/--kill-pacemakerd, task.TaskKill)
    fence_node   (--fence-node, task.TaskFence)
    split_brain  (--split-brain-iptables, task.TaskSplitBrain)
``` 
#### Task* class
- Prerequisite checking
- Logging
- Results/report dumping
- Run command
- Wait/Monitor result in thread

## RPM
[only for Tumbleweed and 15sp2](https://build.opensuse.org/package/show/home:XinLiang:integrate_preflight_check_into_crmsh/crmsh)

## How to run

`/usr/share/crmsh/preflight_check/ha-cluster-preflight-check -h`

## Next steps

- Back port to other branches
- More checking items for `-e`(env checking) and `-c`(cluster related checking) options

> check nic's mtu used by corosync, should large than totem.netmtu
> check rrp mode is passive or active
> check whether using non-exist IP

- Unify logging system with crmsh(#614 ) and hb_report
- More preflight scenarios

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -58,7 +58,11 @@ install-data-hook:
 hanoarchdir = $(datadir)/@PACKAGE@/hb_report
 hanoarch_DATA = hb_report/constants.py hb_report/utillib.py
 hanoarch_SCRIPTS = hb_report/hb_report
-EXTRA_DIST = $(hanoarch_DATA)
+preflightnoarchdir = $(datadir)/@PACKAGE@/preflight_check
+preflightnoarch_DATA = preflight_check/check.py preflight_check/explain.py preflight_check/__init__.py \
+		       preflight_check/config.py preflight_check/main.py preflight_check/utils.py preflight_check/task.py
+preflightnoarch_SCRIPTS = preflight_check/ha-cluster-preflight-check
+EXTRA_DIST = $(hanoarch_DATA) $(preflightnoarch_DATA)
 
 # Python module installation
 all-local:
--- a/crmsh/completers.py
+++ b/crmsh/completers.py
@@ -49,22 +49,21 @@ def resources(args=None):
     if cib_el is None:
         return []
     nodes = xmlutil.get_interesting_nodes(cib_el, [])
-    res = [x.get("id") for x in nodes if xmlutil.is_resource(x)]
+    rsc_id_list = [x.get("id") for x in nodes if xmlutil.is_resource(x)]
     if args and args[0] in ['promote', 'demote']:
-        return list(filter(xmlutil.RscState().is_ms, res))
+        return [item for item in rsc_id_list if xmlutil.RscState().is_ms(item)]
     if args and args[0] == "started":
-        return list(filter(xmlutil.RscState().is_running, res))
+        return [item for item in rsc_id_list if xmlutil.RscState().is_running(item)]
     if args and args[0] == "stopped":
-        for item in filter(xmlutil.RscState().is_running, res):
-            res.remove(item)
-    return res
+        return [item for item in rsc_id_list if not xmlutil.RscState().is_running(item)]
+    return rsc_id_list
 
 
-def resources_started(args):
+def resources_started(args=None):
     return resources(["started"])
 
 
-def resources_stopped(args):
+def resources_stopped(args=None):
     return resources(["stopped"])
 
 
--- a/preflight_check/check.py
+++ b/preflight_check/check.py
@@ -1,0 +1,236 @@
+import re
+from . import utils
+from . import task
+from crmsh import utils as crmshutils
+from crmsh import bootstrap as crmshboot
+from crmsh import completers
+
+
+def check(context):
+    """
+    Check environment and cluster state if related options are enabled
+    """
+    if context.env_check:
+        check_environment()
+    if context.cluster_check:
+        check_cluster()
+    print()
+
+
+def check_environment():
+    """
+    A set of functions to check environment
+    """
+    print("\n============ Checking environment ============")
+    check_my_hostname_resolves()
+    check_time_service()
+    check_firewall()
+
+
+def check_my_hostname_resolves():
+    """
+    check if the hostname is resolvable
+    """
+    task_inst = task.TaskCheck("Checking hostname resolvable")
+    with task_inst.run():
+        if not crmshboot.my_hostname_resolves():
+            task_inst.error('''Hostname "{}" is unresolvable.
+  Please add an entry to /etc/hosts or configure DNS.'''.format(utils.this_node()))
+
+
+def check_time_service():
+    """
+    Check time service
+    """
+    task_inst = task.TaskCheck("Checking time service")
+    with task_inst.run():
+        timekeepers = ('chronyd.service', 'ntp.service', 'ntpd.service')
+        timekeeper = None
+        for tk in timekeepers:
+            if crmshutils.service_is_available(tk):
+                timekeeper = tk
+                break
+        else:
+            task_inst.warn("No NTP service found.")
+            return
+
+        task_inst.info("{} is available".format(timekeeper))
+        if crmshutils.service_is_enabled(timekeeper):
+            task_inst.info("{} is enabled".format(timekeeper))
+        else:
+            task_inst.warn("{} is disabled".format(timekeeper))
+        if crmshutils.service_is_active(timekeeper):
+            task_inst.info("{} is active".format(timekeeper))
+        else:
+            task_inst.warn("{} is not active".format(timekeeper))
+
+
+def check_port_open(task, firewall_type):
+    """
+    Check whether corosync port is blocked by iptables
+    """
+    ports = utils.corosync_port_list()
+    if not ports:
+        task.error("Can not get corosync's port")
+        return
+
+    if firewall_type == "firewalld":
+        rc, out, err = crmshutils.get_stdout_stderr('firewall-cmd --list-port')
+        if rc != 0:
+            task.error(err)
+            return
+        for p in ports:
+            if re.search(' {}/udp'.format(p), out):
+                task.info("UDP port {} is opened in firewalld".format(p))
+            else:
+                task.error("UDP port {} should open in firewalld".format(p))
+    elif firewall_type == "SuSEfirewall2":
+        #TODO
+        pass
+
+
+def check_firewall():
+    """
+    Check the firewall status
+    """
+    task_inst = task.TaskCheck("Checking firewall")
+    with task_inst.run():
+        for item in ("firewalld", "SuSEfirewall2"):
+            if crmshutils.package_is_installed(item):
+                task_inst.info("{}.service is available".format(item))
+                if crmshutils.service_is_active(item):
+                    task_inst.info("{}.service is active".format(item))
+                    check_port_open(task_inst, item)
+                else:
+                    task_inst.warn("{}.service is not active".format(item))
+                break
+        else:
+           task_inst.warn("Failed to detect firewall")
+
+
+def check_cluster():
+    """
+    A set of functions to check cluster state
+    """
+    print("\n============ Checking cluster state ============")
+    if not check_cluster_service():
+        return
+    check_fencing()
+    check_nodes()
+    check_resources()
+
+
+def check_cluster_service(quiet=False):
+    """
+    Check service status of pacemaker/corosync
+    """
+    task_inst = task.TaskCheck("Checking cluster service", quiet=quiet)
+    with task_inst.run():
+        if crmshutils.service_is_enabled("pacemaker"):
+            task_inst.info("pacemaker.service is enabled")
+        else:
+            task_inst.warn("pacemaker.service is disabled")
+
+        if crmshutils.service_is_enabled("corosync"):
+            task_inst.warn("corosync.service is enabled")
+
+        for s in ("corosync", "pacemaker"):
+            if crmshutils.service_is_active(s):
+                task_inst.info("{}.service is running".format(s))
+            else:
+                task_inst.error("{}.service is not running!".format(s))
+        return task_inst.passed
+
+
+def check_fencing():
+    """
+    Check STONITH/Fence:
+      Whether stonith is enabled
+      Whether stonith resource is configured and running
+    """
+    task_inst = task.TaskCheck("Checking STONITH/Fence")
+    with task_inst.run():
+        if not utils.FenceInfo().fence_enabled:
+            task_inst.warn("stonith is disabled")
+            return
+
+        task_inst.info("stonith is enabled")
+        rc, outp, _ = crmshutils.get_stdout_stderr("crm_mon -r1 | grep '(stonith:.*):'")
+        if rc != 0:
+            task_inst.warn("No stonith resource configured!")
+            return
+
+        res = re.search(r'([^\s]+)\s+\(stonith:(.*)\):\s+(\w+)', outp)
+        res_name, res_agent, res_state = res.groups()
+        common_msg = "stonith resource {}({})".format(res_name, res_agent)
+        state_msg = "{} is {}".format(common_msg, res_state)
+
+        task_inst.info("{} is configured".format(common_msg))
+        if res_state == "Started":
+            task_inst.info(state_msg)
+        else:
+            task_inst.warn(state_msg)
+
+        if re.search(r'sbd$', res_agent):
+            if crmshutils.service_is_active("sbd"):
+                task_inst.info("sbd service is running")
+            else:
+                task_inst.warn("sbd service is not running!")
+
+
+def check_nodes():
+    """
+    Check nodes info:
+      Current DC
+      Quorum status
+      Online/OFFLINE/UNCLEAN nodes
+    """
+    task_inst = task.TaskCheck("Checking nodes")
+    with task_inst.run():
+        rc, outp, errp = crmshutils.get_stdout_stderr("crm_mon -1")
+        if rc != 0:
+            task_inst.error("run \"crm_mon -1\" error: {}".format(errp))
+            return
+        # check DC
+        res = re.search(r'Current DC: (.*) \(', outp)
+        if res:
+            task_inst.info("DC node: {}".format(res.group(1)))
+
+        # check quorum
+        if re.search(r'partition with quorum', outp):
+            task_inst.info("Cluster have quorum")
+        else:
+            task_inst.warn("Cluster lost quorum!")
+
+        # check Online nodes
+        res = re.search(r'Online:\s+(\[.*\])', outp)
+        if res:
+            task_inst.info("Online nodes: {}".format(res.group(1)))
+
+        # check OFFLINE nodes
+        res = re.search(r'OFFLINE:\s+(\[.*\])', outp)
+        if res:
+            task_inst.warn("OFFLINE nodes: {}".format(res.group(1)))
+
+        # check UNCLEAN nodes
+        res = re.findall(r'Node (.*): UNCLEAN', outp)
+        for item in res:
+            task_inst.warn('Node {} is UNCLEAN!'.format(item))
+
+
+def check_resources():
+    """
+    Check items of Started/Stopped/FAILED resources
+    """
+    task_inst = task.TaskCheck("Checking resources")
+    with task_inst.run():
+        started_list = completers.resources_started()
+        stopped_list = completers.resources_stopped()
+        # TODO need suitable method to get failed resources list
+        failed_list = []
+        if started_list:
+            task_inst.info("Started resources: {}".format(','.join(started_list)))
+        if stopped_list:
+            task_inst.info("Stopped resources: {}".format(','.join(stopped_list)))
+        if failed_list:
+            task_inst.warn("Failed resources: {}".format(','.join(failed_list)))
--- a/preflight_check/config.py
+++ b/preflight_check/config.py
@@ -1,0 +1,7 @@
+FENCE_TIMEOUT = 60
+FENCE_NODE = "crm_attribute -t status -N '{}' -n terminate -v true"
+BLOCK_IP = '''iptables -{action} INPUT -s {peer_ip} -j DROP;
+              iptables -{action} OUTPUT -d {peer_ip} -j DROP'''
+REMOVE_PORT = "firewall-cmd --zone=public --remove-port={port}/udp"
+ADD_PORT = "firewall-cmd --zone=public --add-port={port}/udp"
+FENCE_HISTORY = "stonith_admin -h {node}"
--- a/preflight_check/main.py
+++ b/preflight_check/main.py
@@ -1,0 +1,250 @@
+import os
+import sys
+import argparse
+import logging
+import logging.config
+from argparse import RawTextHelpFormatter
+
+from . import check
+from . import config
+from . import utils
+from . import task
+from crmsh import utils as crmshutils
+
+
+logger = logging.getLogger('cpc')
+
+
+class Context(object):
+    """
+    Class to store context attributes
+    """
+    def __init__(self):
+        """
+        Initialize attributes
+        """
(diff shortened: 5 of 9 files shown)


nick-wang:
I just begin to use/review the preflight check, so it may take time. List the finding here for reference:
1. I think better to move binary `/usr/share/crmsh/preflight_check/ha-cluster-preflight-check` from `/usr/share/...` to `/usr/sbin/ha-cluster-preflight-check` together with `crm` binary.
2. `ha-cluster-preflight-check -e` will check watchdog and show *ERROR* like:
```
[2020/12/17 16:16:46]INFO: Checking watchdog [Fail]
  WARNING: Watchdog device must be configured if want to use SBD!
```
I think may only show this ERROR when SBD is configured. But it is fine to leave as it is.
3. Firewall is not a must and even it is not installed by default. It is quite normal that no firewall in Cloud environment (Network Security Rules used instead). So i think better to use `WARNING` instead of `ERROR` for this.
```
[2020/12/17 16:16:46]INFO: Checking firewall [Fail]
  ERROR: Failed to detect firewall
```

liangxin1300:
> 2. `ha-cluster-preflight-check -e` will check watchdog and show _ERROR_ like:
> 
> ```
> [2020/12/17 16:16:46]INFO: Checking watchdog [Fail]
>   WARNING: Watchdog device must be configured if want to use SBD!
> ```
> 
> I think may only show this ERROR when SBD is configured. But it is fine to leave as it is.
> 

Right now, this logic was:
```Python
def check_watchdog():
    """
    Verify watchdog device
    """
    task_inst = task.TaskCheck("Checking watchdog")
    with task_inst.run():
        if not crmshutils.service_is_active("sbd.service") and not crmshboot.check_watchdog():
            task_inst.warn("Watchdog device must be configured if want to use SBD!")
```
Watchdog wouldn't be check if sbd.service is active

preflight_check/config.py
```diff
@@ -0,0 +1,7 @@
+FENCE_TIMEOUT = 60
+FENCE_NODE = "crm_attribute -t status -N '{}' -n terminate -v true"
+BLOCK_IP = '''iptables -{action} INPUT -s {peer_ip} -j DROP;
+              iptables -{action} OUTPUT -d {peer_ip} -j DROP'''
+REMOVE_PORT = "firewall-cmd --zone=public --remove-port={port}/udp"
+ADD_PORT = "firewall-cmd --zone=public --add-port={port}/udp"
+FENCE_HISTORY = "stonith_admin --history={node}"
```
nick-wang, line 7: Should use `stonith_admin -h {node}` instead, otherwise won't able to detect the fence message.

Error message: (Actually node2 is fenced as expect)
```
[2020/12/21 17:40:53]INFO: Trying to fence node "SLE15-sp2-gm-nwang-crmshreview-node2"
[2020/12/21 17:40:53]INFO: Waiting 60s for node "SLE15-sp2-gm-nwang-crmshreview-node2" reboot...
[2020/12/21 17:40:54]INFO: Node "SLE15-sp2-gm-nwang-crmshreview-node2" will be fenced by "SLE15-sp2-gm-nwang-crmshreview-node1"!
[2020/12/21 17:41:53]ERROR: Target fence node "SLE15-sp2-gm-nwang-crmshreview-node2" still alive
```
Test environment: SLE15SP2
```
# stonith_admin --history=SLE15-sp2-gm-nwang-crmshreview-node2
reboot of SLE15-sp2-gm-nwang-crmshreview-node2 successful: delegate=SLE15-sp2-gm-nwang-crmshreview-node1, client=pacemaker-controld.1187, origin=SLE15-sp2-gm-nwang-crmshreview-node2, last-successful='2020-12-17 16:50:48 +08:00'
1 event found
# stonith_admin -h SLE15-sp2-gm-nwang-crmshreview-node2
Node SLE15-sp2-gm-nwang-crmshreview-node2 last fenced at: Mon Dec 21 18:01:58 2020
```

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
+import re
+from . import utils
+from . import task
+from crmsh import utils as crmshutils
+from crmsh import bootstrap as crmshboot
+
+
+def check(context):
+    """
+    Check environment and cluster state if related options enabled
```
nick-wang, line 10: s/options enabled/options are enabled/

preflight_check/main.py
```diff
@@ -0,0 +1,250 @@
...
+    parse_argument(context)
+    if not utils.is_root():
+        logging.fatal("{} can only be executed as user root!".format(context.process_name))
+        sys.exit(1)
+    if not os.path.exists(context.var_dir):
+        os.makedirs(context.var_dir, exist_ok=True)
+    setup_logging(context)
+
+    try:
+        check.check(context)
```
nick-wang, line 242: Personally, i would perfer to check each flag of `context`, then call the responding function. 
Instead of always call the function with `context`, do nothing if the `flag` not set.

But it is not a bug that must be fixed.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    check_time_service()
+    check_watchdog()
+    check_firewall()
+
+
+def check_my_hostname_resolves():
+    """
+    Check hostname resolvable
+    """
+    task_inst = task.TaskCheck("Checking hostname resolvable")
```
nick-wang, line 34: Together with the `__doc__`, check if the hostname is resolvable.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    with task_inst.run():
+        timekeepers = ('chronyd.service', 'ntp.service', 'ntpd.service')
+        timekeeper = None
+        for tk in timekeepers:
+            if crmshutils.service_is_available(tk):
+                timekeeper = tk
+                break
+
+        if timekeeper is None:
+            task_inst.warn("No NTP service found.")
```
nick-wang, line 55: Better to move this as `else` branch of `for tk in timekeepers:`. Then no need extra indent for later code.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+                task_inst.warn("{} is not active".format(timekeeper))
+
+
+def check_watchdog():
+    """
+    Verify watchdog device
+    """
+    task_inst = task.TaskCheck("Checking watchdog")
+    with task_inst.run():
+        if not crmshutils.service_is_active("sbd.service") and not crmshboot.check_watchdog():
```
nick-wang, line 75: Maybe only check service_is_enabled is better, since `sbd` is not a must for many cases.
Also don't need to check the condition of `sbd.service` active, since it will watchdog always exist if "sbd.service" actived

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    ports = utils.corosync_port_list()
+    if not ports:
+        task.error("Can not get corosync's port")
+        return
+
+    if firewall_type == "firewalld":
+        rc, out, err = crmshutils.get_stdout_stderr('firewall-cmd --list-port')
+        if rc != 0:
+            task.error(err)
```
nick-wang, line 91: Need return if `rc != 0`.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+        for p in ports:
+            if re.search(' {}/udp'.format(p), out):
+                task.info("UDP port {} is opened in firewalld".format(p))
+            else:
+                task.error("UDP port {} should open in firewalld".format(p))
+    if firewall_type == "SuSEfirewall2":
```
nick-wang, line 97: elif is better, or need return after condition "firewalld"

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    print("\n============ Checking environment ============")
+    check_my_hostname_resolves()
+    check_time_service()
+    check_watchdog()
+    check_firewall()
+
+
+def check_my_hostname_resolves():
+    """
+    Check hostname resolvable
```
nick-wang, line 32: check if the hostname is resolvable.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+
+            if crmshutils.service_is_active(timekeeper):
+                task_inst.info("{} is active".format(timekeeper))
+            else:
+                task_inst.warn("{} is not active".format(timekeeper))
+
+
+def check_watchdog():
+    """
+    Verify watchdog device
```
nick-wang, line 71: Check if watchdog device is exist.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+            else:
+                task.error("UDP port {} should open in firewalld".format(p))
+    if firewall_type == "SuSEfirewall2":
+        #TODO
+        pass
+
+
+def check_firewall():
+    """
+    Check firewall
```
nick-wang, line 104: Check the firewall status

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    Check service status of pacemaker/corosync
+    """
+    task_inst = task.TaskCheck("Checking cluster service", quiet=quiet)
+    with task_inst.run():
+        for s in ("pacemaker", ):
+            if crmshutils.service_is_enabled(s):
+                task_inst.info("{} is enabled".format(s))
+            else:
+                task_inst.warn("{} is disabled".format(s))
+
```
nick-wang, line 144: We also need to check "corosync" should be disabled.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+      Whether stonith resource is configured and running
+    """
+    task_inst = task.TaskCheck("Checking STONITH/Fence")
+    with task_inst.run():
+        if not utils.FenceInfo().fence_enabled:
+            task_inst.warn("stonith is disabled")
+            return
+
+        task_inst.info("stonith is enabled")
+        rc, outp, _ = crmshutils.get_stdout_stderr("crm_mon -r1 | grep '(stonith:.*):'")
```
nick-wang, line 166: Please move `"crm_mon -r1 | grep '(stonith:.*):'"` to config.py as a const.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+def check_nodes():
+    """
+    Check nodes info:
+      Current DC
+      Quorum status
+      Online/OFFLINE/UNCLEAN nodes
+    """
+    task_inst = task.TaskCheck("Checking nodes")
+    with task_inst.run():
+        rc, outp, errp = crmshutils.get_stdout_stderr("crm_mon -1")
```
nick-wang, line 198: Please move this to config.py with a const.

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+            res = re.search(r'OFFLINE:\s+(\[.*\])', outp)
+            if res:
+                task_inst.warn("OFFLINE nodes: {}".format(res.group(1)))
+
+            # check UNCLEAN nodes
+            res = re.findall(r'Node (.*): UNCLEAN', outp)
+            for item in res:
+                task_inst.warn('Node {} is UNCLEAN!'.format(item))
+        else:
+            task_inst.error("run \"crm_mon -1\" error: {}".format(errp))
```
nick-wang, line 226: Use the const above. 
Better to return first when `rc != 0`, so that no extra indent needed for the later `if`

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+    task_inst = task.TaskCheck("Checking resources")
+    with task_inst.run():
+        started_list = []
+        stopped_list = []
+        failed_list = []
+        cmd = "crm_mon -r1"
+
+        rc, outp, errp = crmshutils.get_stdout_stderr(cmd)
+        if rc == 0:
+            res = re.findall(r'([^\s]+)\s+\(.*\):\s+(\w+) ', outp)
```
nick-wang, line 242: This won't show the clone resource, for example:
```
  * libvirt_stonith	(stonith:external/libvirt):	 Started SLE15-sp2-gm-nwang-crmshreview-node1
  * Clone Set: ms_1-multi-0 [res-1-multi-0] (promotable):
    * Masters: [ SLE15-sp2-gm-nwang-crmshreview-node2 ]
    * Slaves: [ SLE15-sp2-gm-nwang-crmshreview-node1 ]
```

preflight_check/check.py
```diff
@@ -0,0 +1,254 @@
...
+            res = re.findall(r'Node (.*): UNCLEAN', outp)
+            for item in res:
+                task_inst.warn('Node {} is UNCLEAN!'.format(item))
+        else:
+            task_inst.error("run \"crm_mon -1\" error: {}".format(errp))
+
+
+def check_resources():
+    """
+    Check number of Started/Stopped/FAILED resources
```
nick-wang, line 231: The output is the items instead of number of each list.

preflight_check/main.py
```diff
@@ -0,0 +1,250 @@
...
+                'level': 'DEBUG'
+                }
+            }
+        }
+
+
+def kill_process(context):
+    """
+    Testcase: kill cluster related processes
+    --kill-sbd:           restarted or fenced
```
nick-wang, line 91: Better to remove `:` to align others

preflight_check/task.py
```diff
@@ -0,0 +1,484 @@
...
+        self.task_pre_check()
+        rc, pid = utils.get_process_status(self.target_kill)
+        if not rc:
+            raise TaskError("Process {} is not running!".format(self.target_kill))
+
+    def run(self):
+        while True:
+            rc, pid = utils.get_process_status(self.target_kill)
+            if rc:
+                self.info("Process {}({}) is running...".format(self.target_kill, pid))
```
nick-wang, line 352: Is this necessary? Since have already run `pre_check` for the process status.

liangxin1300: Might be overlapped, but it has different concern, `pre_check` is to check the cluster is running, here is to check target process is running

preflight_check/main.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed.
Instead of using un-readable regexp, here just calling `crmsh.completers.resources_started` and `crmsh.completers.resources_stopped` to fetch the resources list

preflight_check/check.py
liangxin1300: Good suggestions! Thanks!

preflight_check/check.py
liangxin1300: OK. Changed

preflight_check/check.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

Makefile.am
```diff
@@ -58,7 +58,11 @@ install-data-hook:
 hanoarchdir = $(datadir)/@PACKAGE@/hb_report
 hanoarch_DATA = hb_report/constants.py hb_report/utillib.py
 hanoarch_SCRIPTS = hb_report/hb_report
-EXTRA_DIST = $(hanoarch_DATA)
+preflightnoarchdir = $(datadir)/@PACKAGE@/preflight_check
```
nick-wang, line 61: Strictly speaking, binary should not put in `datadir`.

crmsh/completers.py
```diff
@@ -49,22 +49,19 @@ def resources(args=None):
...
     if args and args[0] == "stopped":
-        for item in filter(xmlutil.RscState().is_running, res):
-            res.remove(item)
-    return res
+        return [item for item in rsc_id_list if not xmlutil.RscState().is_running(item)]
+    return rsc_id_list
 
 
-def resources_started(args):
+def resources_started(args=None):
```
nick-wang, line 60: Since all the caller are calling without `arg` and always call `resources(["started"])`, so maybe no need to have `args=None`, use `resources_started()` should enough.

crmsh/completers.py
```diff
@@ -49,22 +49,19 @@ def resources(args=None):
...
     return resources(["started"])
 
 
-def resources_stopped(args):
+def resources_stopped(args=None):
```
nick-wang, line 64: Same as `resources_started`, `resources_started()` should enough.

preflight_check/main.py
```diff
@@ -0,0 +1,250 @@
...
+    group_mutual.add_argument('--split-brain-iptables', dest='sp_iptables', action='store_true',
+                              help='Make split brain by blocking corosync ports')
+    parser.add_argument('-l', '--kill-loop', dest='loop', action='store_true',
+                        help='Kill process in loop')
+
+    other_options = parser.add_argument_group('other options')
+    other_options.add_argument('-y', '--yes', dest='yes', action='store_true',
+                               help='Answer "yes" if asked to run the test')
+    other_options.add_argument('-h', '--help', dest='help', action='store_true',
+                               help='show this help message and exit')
```
nick-wang, line 199: Start with upper case like others, 'Show this ...`

preflight_check/utils.py
```diff
@@ -0,0 +1,234 @@
...
+    returns Boolean and pid
+    """
+    # find pids of running processes
+    pids = [pid for pid in os.listdir('/proc') if pid.isdigit()]
+    for pid in pids:
+        try:
+            pid_file = os.path.join('/proc', pid, 'cmdline')
+            with open(pid_file, 'rb') as f:
+                data = f.read()
+                procname = os.path.basename(crmshutils.to_ascii(data).replace('\x00', ' ').split(' ')[0])
```
nick-wang, line 228: How about use library `psutil.process_iter()` or even using shell command `pidof` to replace path through the `/proc`?

preflight_check/check.py
```diff
@@ -0,0 +1,246 @@
...
+            else:
+                task_inst.warn("{} is disabled".format(timekeeper))
+
+            if crmshutils.service_is_active(timekeeper):
+                task_inst.info("{} is active".format(timekeeper))
+            else:
+                task_inst.warn("{} is not active".format(timekeeper))
+
+
+def check_watchdog():
```
nick-wang, line 70: Besides device, can check kernel module `softdog` as well. Just a hint or warning is enough.

preflight_check/task.py
nick-wang: But anyway, it is the same `self.target_kill`

preflight_check/check.py
```diff
@@ -0,0 +1,246 @@
...
+    """
+    task_inst = task.TaskCheck("Checking nodes")
+    with task_inst.run():
+        rc, outp, errp = crmshutils.get_stdout_stderr("crm_mon -1")
+        if rc != 0:
+            task_inst.error("run \"crm_mon -1\" error: {}".format(errp))
+        # check DC
+        res = re.search(r'Current DC: (.*) \(', outp)
+        if res:
+            task_inst.info("DC node: {}".format(res.group(1)))
```
nick-wang, line 207: How about use a function to parse all information(DC node, Online nodes, Offline nodes, Unclean nodes, etc...) of cluster, call the function to get information in everywhere.

preflight_check/task.py
```diff
@@ -0,0 +1,484 @@
...
+
+    def un_block_iptables(self):
+        for node in self.peer_nodelist:
+            self.info("Trying to recover {} communication ip".format(node))
+            for ip in crmshutils.get_iplist_from_name(node):
+                crmshutils.get_stdout_stderr(config.BLOCK_IP.format(action='D', peer_ip=ip))
+
+    def run(self):
+        """
+        Fire to fence node and start a thread to monitor the result
```
nick-wang, line 470: I think "Fence node and ..." is enough.

nick-wang (changes requested):
@liangxin1300 Sorry for the long waiting. Finally, i finish the 1st round review, please check the comments, thank.

preflight_check/config.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

preflight_check/main.py
liangxin1300: The reason I write it in this way was to keep the block size of `run` function short and clear

preflight_check/check.py
liangxin1300: Changed!

preflight_check/check.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

preflight_check/task.py
```diff
@@ -0,0 +1,484 @@
...
+        Dump join result
+        """
+        self.build_base_result()
+        self.result['Fence action'] = self.fence_action
+        self.result['Fence timeout'] = self.fence_timeout
+        from . import main
+        main.ctx.task_list = self.prev_task_list + [self.result]
+        utils.json_dumps()
+
+    def to_report(self):
```
nick-wang, line 181: Put these empty function `to_report` from each `TaskXXX` object to the `Task`.

preflight_check/task.py
liangxin1300: Changed

preflight_check/main.py
liangxin1300: Changed

preflight_check/check.py
liangxin1300: Changed

liangxin1300:
Other changes for now:
- Drop watchdog check

Makefile.am
liangxin1300: We can change this when we decide where to put it

crmsh/completers.py
liangxin1300: But the caller `@command.completers(compl.resources_started)`, inside it:
```
def completers(*fns):
    '''
    Use to set a list of positional tab completers for the command.
    Each completer gets as its argument the command line entered so far,
    and returns a list of possible completions.
    '''
    def cfn(args):
        nargs = len(args) - 1
        if nargs == 0:
            return [args[0]]
        if nargs <= len(fns):
            return fns[nargs-1](args)
        return []

    def inner(fn):
        setattr(fn, '_completer', cfn)
        return fn
    return inner
```

`fns[nargs-1](args)` it contains 1 positional argument

preflight_check/utils.py
liangxin1300: We have similar discuss before https://github.com/ClusterLabs/crmsh/pull/185
Anyway, current code works well

preflight_check/task.py
liangxin1300: Changed.
Thanks!

Makefile.am
nick-wang: sure.

nick-wang (approved):
Thanks for the pr and the change. It looks good!

PR #665, opened 2020-10-29
Replace travisCI as github actions

liangxin1300:
## Problem
Recently, speed of travisCI was too slow to tolerate, see https://travis-ci.org/github/ClusterLabs/crmsh/builds/739286101

## Solution
Using `github actions`.
As far as I can tell, benefits might include:
1. more faster: 6 hrs vs 11 mins(recently) , and 22 mins vs 11 mins(22 mins was historical best speed for travisCI) for the same test process
2. more parallel jobs: 5 jobs vs 10+ jobs
3. directly UI: directly see the results at the checks tabs in each PR

## Related docs
For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
For more information about secrets see: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets


.github/workflows/crmsh-ci.yml
```diff
@@ -0,0 +1,203 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
+    - name: Test with pytest in tox
+      run: |
+        tox
```
tomschr, line 37: Do you want to have the verbose option here?

```suggestion
        tox -v
```

arbulu89 (approved):
@liangxin1300 It looks good to me.
I don't have any preference between travis/github actions.
If you see that github actions work better, go ahead. I'm totally fine with it.
Good job

.github/workflows/crmsh-ci.yml
liangxin1300: Thanks! Changed

liangxin1300:
> @liangxin1300 It looks good to me.
> I don't have any preference between travis/github actions.
> If you see that github actions work better, go ahead. I'm totally fine with it.
> Good job

Thanks for comment

PR #669, opened 2020-10-30
[crmsh-4.1] Replace travisCI as github actions

liangxin1300:
backport #665


nick-wang (approved):
@liangxin1300 Didn't try github action before. It is fine if github action play better and a back port

PR #673, opened 2020-11-09
Minor change for _get_sbd_device_interactive function

liangxin1300:
Changes include:

- Using .format instead of %s
- Meaningful variable names
- Using print_error_msg instead of term.render
- Continue if sbd device value is None to avoid crash(bsc#1178333)
- Add unit test codes for _get_sbd_device_interactive function


nick-wang (approved):
Looks good to me.

PR #677, opened 2020-11-11
Add lock for multi nodes join in parallel

liangxin1300:
## Motivation
For multi nodes scenarios, while multi nodes joining to the init node at the same time, the results will be undefined and risky.
It might cause errors on ssh key deployment, csync2 or config files not synced.

## Solution
By touching a lock file to make multi nodes join process as one-by-one joining.

## Logic
Assuming we have 3 nodes(A, B, C), A is the init node, just finish the init process.

1. A lock file will be touched on init node A when the first join node B try to join
2. B starts another thread to update the timestamp of the lock file(in 10s interval)
3. Another join nodes C detect(in 15s interval) lock file exists and timestamp updates, waiting...
4. B finished joining, remove lock file
5. If C found no local file exist, touch the lock file and starts joining
6. If C found lock file always exists and the timestamp not update, remove that lock file and start joining

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -30,6 +30,7 @@ from . import corosync
 from . import tmpfiles
 from . import clidisplay
 from . import term
+from . import join_lock
 
 
 LOG_FILE = "/var/log/crmsh/ha-cluster-bootstrap.log"
@@ -1711,8 +1712,6 @@ def join_ssh(seed_host):
     if not invoke("ssh root@{} crm cluster init -i {} ssh_remote".format(seed_host, _context.default_nic_list[0])):
         error("Can't invoke crm cluster init -i {} ssh_remote on {}".format(_context.default_nic_list[0], seed_host))
 
-    setup_passwordless_with_other_nodes(seed_host)
-
 
 def swap_public_ssh_key(remote_node):
     """
@@ -2338,10 +2337,14 @@ def bootstrap_join(context):
             _context.cluster_node = cluster_node
 
         join_ssh(cluster_node)
-        join_remote_auth(cluster_node)
-        join_csync2(cluster_node)
-        join_ssh_merge(cluster_node)
-        join_cluster(cluster_node)
+
+        lock_inst = join_lock.JoinLock(cluster_node)
+        with lock_inst.lock():
+            setup_passwordless_with_other_nodes(cluster_node)
+            join_remote_auth(cluster_node)
+            join_csync2(cluster_node)
+            join_ssh_merge(cluster_node)
+            join_cluster(cluster_node)
 
     status("Done (log saved to %s)" % (LOG_FILE))
 
--- a/crmsh/join_lock.py
+++ b/crmsh/join_lock.py
@@ -1,0 +1,140 @@
+# Copyright (C) 2020 Xin Liang <XLiang@suse.com>
+# See COPYING for license information.
+
+
+import re
+import time
+from contextlib import contextmanager
+
+from . import bootstrap
+from . import utils
+from . import config
+
+
+class SSHError(Exception):
+    pass
+
+
+class JoinLock(object):
+    """
+    Class to manage lock for multiple nodes join in parallel
+    """
+
+    JOIN_LOCK_DIR = "/tmp/.crmsh_join_lock_directory"
+    MKDIR_CMD = "mkdir {}".format(JOIN_LOCK_DIR)
+    RM_CMD = "rm -rf {}".format(JOIN_LOCK_DIR)
+    SSH_TIMEOUT = 10
+    SSH_OPTION = "-o ConnectTimeout={} -o StrictHostKeyChecking=no".format(SSH_TIMEOUT)
+    SSH_EXIT_ERR = 255
+    MIN_JOIN_TIMEOUT = 120
+    WAIT_INTERVAL = 10
+
+    def __init__(self, init_node):
+        """
+        Init function
+        """
+        self.init_node = init_node
+        # only the lock owner can unlock
+        self.lock_owner = False
+
+    @property
+    def join_timeout(self):
+        """
+        Get join_timeout from config.core
+        """
+        try:
+            value = int(config.core.join_timeout)
+        except ValueError:
+            raise ValueError("Invalid format of core.join_timeout(should be a number)")
+        if value < self.MIN_JOIN_TIMEOUT:
+            raise ValueError("Minimum value of core.join_timeout should be {}".format(self.MIN_JOIN_TIMEOUT))
+        return value
+
+    def _run(self, cmd):
+        """
+        Run command on target node, consider specific exceptions
+        """
+        cmd_with_ssh = "ssh {} root@{} \"{}\"".format(self.SSH_OPTION, self.init_node, cmd)
+        rc, out, err = utils.get_stdout_stderr(cmd_with_ssh)
+        if rc == self.SSH_EXIT_ERR:
+            raise SSHError(err)
+        return rc, out, err
+
+    def _create_lock_dir(self):
+        """
+        Create lock directory, mkdir command was atomic
+        """
+        rc, _, _ = self._run(self.MKDIR_CMD)
+        return rc == 0
+
+    def _get_online_nodelist(self):
+        """
+        Get the online node list from init node
+        """
+        rc, out, err = self._run("crm_node -l")
+        if rc != 0 and err:
+            raise RuntimeError(err)
+        return re.findall('[0-9]+ (.*) member', out)
+
+    def _lock_or_wait(self):
+        """
+        Try to claim lock on init node,
+        wait if failed to claim
+        exit if reached the join_timeout
+        """
+        warned_once = False
+        online_list = []
+        pre_online_list = []
+        expired_error_str = "Cannot continue since the lock directory exists at the init node ({}:{})".format(self.init_node, self.JOIN_LOCK_DIR)
+
+        current_time = int(time.time())
+        timeout = current_time + self.join_timeout
+        while current_time <= timeout:
+
+            # Try to claim the lock
+            if self._create_lock_dir():
+                # Success
+                self.lock_owner = True
+                break
+
+            # Might lose claiming lock again, start to wait again
+            online_list = self._get_online_nodelist()
+            if pre_online_list and pre_online_list != online_list:
+                warned_once = False
+                current_time = int(time.time())
+                timeout = current_time + self.join_timeout
+                continue
+            else:
+                pre_online_list = online_list
+
+            if not warned_once:
+                warned_once = True
+                bootstrap.warn("Other node still joining, wait at most {}s...".format(self.join_timeout))
+
+            time.sleep(self.WAIT_INTERVAL)
+            current_time = int(time.time())
+
+        else:
+            raise TimeoutError("Join process failed after {} seconds. {}".format(self.join_timeout, expired_error_str))
+
+    @contextmanager
+    def lock(self):
+        """
+        Create lock directory on target node
+        """
+        try:
+            self._lock_or_wait()
+            yield
+        except SSHError as err:
+            bootstrap.error(str(err))
+        except:
+            raise
+        finally:
+            self.unlock()
+
+    def unlock(self):
+        """
+        Remove the lock directory on target node
+        """
+        if self.lock_owner:
+            self._run(self.RM_CMD)
--- a/crm.conf.in
+++ b/crm.conf.in
@@ -19,6 +19,7 @@
 ; dot = dot
 ; ignore_missing_metadata = no
 ; report_tool_options =
+; join_timeout = 120
 
 ; obscure_pattern option is the persisent configuration of CLI.
 ; Example, for the high security concern, obscure_pattern = passw* | ip
--- a/crmsh/config.py
+++ b/crmsh/config.py
@@ -239,6 +239,7 @@ DEFAULTS = {
         'dot': opt_program('', ('dot',)),
         'ignore_missing_metadata': opt_boolean('no'),
         'report_tool_options': opt_string(''),
+        'join_timeout': opt_string('120'),
         'obscure_pattern': opt_string('passw*')
     },
     'path': {
--- a/data-manifest
+++ b/data-manifest
@@ -184,12 +184,14 @@ test/unittests/test.conf
 test/unittests/test_corosync.py
 test/unittests/test_gv.py
 test/unittests/test_handles.py
+test/unittests/test_join_lock.py
 test/unittests/test_objset.py
 test/unittests/test_parallax.py
 test/unittests/test_parse.py
 test/unittests/test_report.py
 test/unittests/test_scripts.py
 test/unittests/test_time.py
+test/unittests/test_ui_cluster.py
 test/unittests/test_utils.py
 test/update-expected-output.sh
 utils/crm_clean.py
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -579,13 +579,12 @@ class TestBootstrap(unittest.TestCase):
             bootstrap.join_ssh(None)
         mock_error.assert_called_once_with("No existing IP/hostname specified (use -c option)")
 
-    @mock.patch('crmsh.bootstrap.setup_passwordless_with_other_nodes')
     @mock.patch('crmsh.bootstrap.error')
     @mock.patch('crmsh.bootstrap.invoke')
     @mock.patch('crmsh.bootstrap.swap_public_ssh_key')
     @mock.patch('crmsh.bootstrap.configure_local_ssh_key')
     @mock.patch('crmsh.utils.start_service')
-    def test_join_ssh(self, mock_start_service, mock_config_ssh, mock_swap, mock_invoke, mock_error, mock_swap_other):
+    def test_join_ssh(self, mock_start_service, mock_config_ssh, mock_swap, mock_invoke, mock_error):
         bootstrap._context = mock.Mock(default_nic_list=["eth1"])
         mock_invoke.return_value = False
 
@@ -596,7 +595,6 @@ class TestBootstrap(unittest.TestCase):
         mock_swap.assert_called_once_with("node1")
         mock_invoke.assert_called_once_with("ssh root@node1 crm cluster init -i eth1 ssh_remote")
         mock_error.assert_called_once_with("Can't invoke crm cluster init -i eth1 ssh_remote on node1")
-        mock_swap_other.assert_called_once_with("node1")
 
     @mock.patch('crmsh.bootstrap.warn')
     @mock.patch('crmsh.bootstrap.fetch_public_key_from_remote_node')
--- a/test/unittests/test_join_lock.py
+++ b/test/unittests/test_join_lock.py
@@ -1,0 +1,206 @@
+"""
+Unitary tests for crmsh/join_lock.py
+
+:author: xinliang
+:organization: SUSE Linux GmbH
+:contact: XLiang@suse.de
+
+:since: 2020-11-15
+"""
+
+# pylint:disable=C0103,C0111,W0212,W0611
+
+import os
+import unittest
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from crmsh import join_lock, config
+
+
+class TestJoinLock(unittest.TestCase):
+    """
+    Unitary tests for crmsh.join_lock.JoinLock
+    """
+
+    @classmethod
+    def setUpClass(cls):
+        """
+        Global setUp.
+        """
+
+    def setUp(self):
+        """
+        Test setUp.
+        """
+        self.lock_inst = join_lock.JoinLock("node1")
+
+    def tearDown(self):
+        """
+        Test tearDown.
+        """
+
+    @classmethod
+    def tearDownClass(cls):
+        """
+        Global tearDown.
+        """
+
+    def test_join_timeout_error_format(self):
+        config.core.join_timeout = "pwd"
+        with self.assertRaises(ValueError) as err:
+            self.lock_inst.join_timeout
+        self.assertEqual("Invalid format of core.join_timeout(should be a number)", str(err.exception))
+
+    def test_join_timeout_min_error(self):
+        config.core.join_timeout = "12"
+        with self.assertRaises(ValueError) as err:
+            self.lock_inst.join_timeout
+        self.assertEqual("Minimum value of core.join_timeout should be 120", str(err.exception))
+
+    def test_join_timeout(self):
+        config.core.join_timeout = "130"
+        self.assertEqual(self.lock_inst.join_timeout, 130)
+
+    @mock.patch('crmsh.utils.get_stdout_stderr')
+    def test_run_error(self, mock_run):
+        mock_run.return_value = (255, "output", "error data")
+        with self.assertRaises(join_lock.SSHError) as err:
+            self.lock_inst._run("test_cmd")
+        self.assertEqual("error data", str(err.exception))
+        mock_run.assert_called_once_with('ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no root@node1 "test_cmd"')
+
+    @mock.patch('crmsh.utils.get_stdout_stderr')
+    def test_run(self, mock_run):
+        mock_run.return_value = (0, "output data", None)
+        rc, out, err = self.lock_inst._run("test_cmd")
+        self.assertEqual(mock_run.return_value, (rc, out, err))
+        mock_run.assert_called_once_with('ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no root@node1 "test_cmd"')
+
+    @mock.patch('crmsh.join_lock.JoinLock._run')
+    def test_create_lock_dir(self, mock_run):
+        mock_run.return_value = (0, None, None)
+        rc = self.lock_inst._create_lock_dir()
+        self.assertEqual(rc, True)
+        mock_run.assert_called_once_with(join_lock.JoinLock.MKDIR_CMD)
+
+    @mock.patch('crmsh.join_lock.JoinLock._run')
+    def test_get_online_nodelist_error(self, mock_run):
+        mock_run.return_value = (1, None, "error data")
+        with self.assertRaises(RuntimeError) as err:
+            self.lock_inst._get_online_nodelist()
+        self.assertEqual("error data", str(err.exception))
+        mock_run.assert_called_once_with("crm_node -l")
+
+    @mock.patch('crmsh.join_lock.JoinLock._run')
+    def test_get_online_nodelist(self, mock_run):
+        output = """
+        1084783297 15sp2-1 member
+        1084783193 15sp2-2 lost
+        1084783331 15sp2-3 member
+        """
+        mock_run.return_value = (0, output, None)
+        res = self.lock_inst._get_online_nodelist()
+        self.assertEqual(res, ["15sp2-1", "15sp2-3"])
+        mock_run.assert_called_once_with("crm_node -l")
+
+    @mock.patch('crmsh.join_lock.JoinLock._create_lock_dir')
+    @mock.patch('crmsh.join_lock.JoinLock.join_timeout', new_callable=mock.PropertyMock)
+    @mock.patch('time.time')
+    def test_lock_or_wait_break(self, mock_time, mock_time_out, mock_create):
+        mock_time.return_value = 10000
+        mock_time_out.return_value = 120
+        mock_create.return_value = True
+
+        self.lock_inst._lock_or_wait()
+        self.assertEqual(self.lock_inst.lock_owner, True)
+
+        mock_time.assert_called_once_with()
+        mock_time_out.assert_called_once_with()
+
+    @mock.patch('time.sleep')
+    @mock.patch('crmsh.bootstrap.warn')
+    @mock.patch('crmsh.join_lock.JoinLock._get_online_nodelist')
+    @mock.patch('crmsh.join_lock.JoinLock._create_lock_dir')
+    @mock.patch('crmsh.join_lock.JoinLock.join_timeout', new_callable=mock.PropertyMock)
+    @mock.patch('time.time')
+    def test_lock_or_wait_timed_out(self, mock_time, mock_time_out, mock_create,
+            mock_get_nodelist, mock_warn, mock_sleep):
+        mock_time.side_effect = [10000, 10120, 10500]
+        mock_time_out.side_effect = [120, 120, 120]
+        mock_create.side_effect = [False, False]
+        mock_get_nodelist.side_effect = ["node1", "node1"]
+
+        with self.assertRaises(TimeoutError) as err:
+            self.lock_inst._lock_or_wait()
+        self.assertEqual("Join process failed after 120 seconds. Cannot continue since the lock directory exists at the init node (node1:/tmp/.crmsh_join_lock_directory)", str(err.exception))
+
+        mock_time.assert_has_calls([mock.call(), mock.call(), mock.call()])
+        mock_time_out.assert_has_calls([mock.call(), mock.call(), mock.call()])
+        mock_create.assert_has_calls([mock.call(), mock.call()])
+        mock_get_nodelist.assert_has_calls([mock.call(), mock.call()])
+        mock_warn.assert_called_once_with("Other node still joining, wait at most 120s...")
+        mock_sleep.assert_has_calls([mock.call(10), mock.call(10)])
+
+    @mock.patch('time.sleep')
+    @mock.patch('crmsh.bootstrap.warn')
+    @mock.patch('crmsh.join_lock.JoinLock._get_online_nodelist')
+    @mock.patch('crmsh.join_lock.JoinLock._create_lock_dir')
+    @mock.patch('crmsh.join_lock.JoinLock.join_timeout', new_callable=mock.PropertyMock)
+    @mock.patch('time.time')
+    def test_lock_or_wait_again(self, mock_time, mock_time_out, mock_create,
+            mock_get_nodelist, mock_warn, mock_sleep):
+        mock_time.side_effect = [10000, 10010, 10020]
+        mock_time_out.side_effect = [120, 120, 120]
+        mock_create.side_effect = [False, False, True]
+        mock_get_nodelist.side_effect = [["node1"], ["node1", "node2"]]
+
+        self.lock_inst._lock_or_wait()
+
+        mock_time.assert_has_calls([mock.call(), mock.call(), mock.call()])
+        mock_time_out.assert_has_calls([mock.call(), mock.call(), mock.call()])
+        mock_create.assert_has_calls([mock.call(), mock.call(), mock.call()])
+        mock_get_nodelist.assert_has_calls([mock.call(), mock.call()])
+        mock_warn.assert_called_once_with("Other node still joining, wait at most 120s...")
+        mock_sleep.assert_called_once_with(10)
+
+    @mock.patch('crmsh.join_lock.JoinLock.unlock')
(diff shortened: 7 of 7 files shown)


crmsh/bootstrap.py
```diff
@@ -1692,6 +1693,106 @@ def init():
     init_network()
 
 
+class JoinLock(object):
+    """
+    Class to manage lock for multiple nodes join in parallel
+    """
+
+    JOIN_LOCK_FILE = "/usr/share/crmsh/.join_lock_file"
+    DETECT_CMD = "test -f {}".format(JOIN_LOCK_FILE)
```
arbulu89, line 1702: Do we need to use this `bash` commands here? 
There are not pythonic options to run this simple operations?

crmsh/bootstrap.py
```diff
@@ -1692,6 +1693,106 @@ def init():
...
+                        # remove lock file on target node
+                        self._run(self.RM_CMD)
+                        break
+                else:
+                    origin_timestamp = timestamp
+
+                if not warned_once:
+                    warned_once = True
+                    warn("Node {} still joining, wait...".format(current_node))
+            time.sleep(15)
```
arbulu89, line 1760: I would put this `15` in a constant

crmsh/bootstrap.py
```diff
@@ -2335,10 +2436,18 @@ def bootstrap_join(context):
             _context.cluster_node = cluster_node
 
         join_ssh(cluster_node)
-        join_remote_auth(cluster_node)
-        join_csync2(cluster_node)
-        join_ssh_merge(cluster_node)
-        join_cluster(cluster_node)
+
+        lock_inst = JoinLock(cluster_node)
```
arbulu89, line 2440: We could put this in some kind of `context manager` or `decorator`

crmsh/bootstrap.py
arbulu89: I guess that you need them because the commands are executed remotely, but it's quite ugly

crmsh/bootstrap.py
```diff
@@ -1692,6 +1693,106 @@ def init():
...
+    def _wait(self):
+        """
+        Wait 15 seconds while detect target node has lock file
+        """
+        warned_once = False
+        origin_timestamp = None
+
+        while self._detect():
+            # tail the last line in lock file
+            rc, out = self._run(self.TAIL_CMD)
```
arbulu89, line 1740: Maybe we could create methods for all of these `self._run` methods, like `_get_tail`.
To make it a bit more meaningful. The same you do with `_detect`.
This way, for example with `TAIL_CMD`, we could put the all of this logic inside the method, and only return the timestamp as `int`
It is not wrong anyway

crmsh/bootstrap.py
```diff
@@ -1692,6 +1693,106 @@ def init():
...
+        # start thread to update timestamp in lock file
+        th = threading.Thread(target=self._update)
+        th.start()
+
+    def unlock(self):
+        """
+        Stop the update thread and remove the lock file on target node
+        """
+        self.thread_stop_event.set()
+        self._run(self.RM_CMD)
```
arbulu89, line 1793: We should run the command before setting the event flag, right?

crmsh/bootstrap.py
```diff
@@ -1692,6 +1693,106 @@ def init():
...
+    def _update(self):
+        """
+        Update timestamp(10s interval) in lock file on target node, format is:
+            nodename timestamp
+        """
+        while not self.thread_stop_event.is_set():
+            # TODO utils.run_cmd_on_remote have issue to run in thread
+            if not invoke("ssh root@{} \"{}\"".format(self.target_node, self.UPDATE_CMD)):
+                error("Can't invoke {} on {}".format(self.UPDATE_CMD, self.target_node))
+            time.sleep(10)
```
arbulu89, line 1771: Set this `10` in a constant

arbulu89 (commented):
@liangxin1300 Why do we need such a complex locking mechanism? Adding timestamps etc.
Besides that, adding threads is not desirable, more complexity.

Anyway, if there is not better option..
Some comments:
- I would put this lock mechanism in other file, like `join_lock.py`, to make it easier to spot
-

crmsh/bootstrap.py
liangxin1300: OK, changed!

crmsh/bootstrap.py
liangxin1300: Changed!

crmsh/bootstrap.py
liangxin1300: I think it's not a big issue, besides that remove first might lead to update timstamp command failed

crmsh/bootstrap.py

liangxin1300:
> * I would put this lock mechanism in other file, like `join_lock.py`, to make it easier to spot
Hi @arbulu89 , Already changed this point, thanks!

crmsh/bootstrap.py
liangxin1300: OK, changed, create some meaningful and small methods, and put part of logic into `_get_last_timestamp`

crmsh/bootstrap.py
```diff
@@ -2335,10 +2334,14 @@ def bootstrap_join(context):
...
 
         join_ssh(cluster_node)
-        join_remote_auth(cluster_node)
-        join_csync2(cluster_node)
-        join_ssh_merge(cluster_node)
-        join_cluster(cluster_node)
+
+        lock_inst = join_lock.JoinLock(cluster_node)
+        with lock_inst.lock():
+            setup_passwordless_with_other_nodes(cluster_node)
```
liangxin1300, line 2340: Consider on this time other node was joining, we should call `setup_passwordless_with_other_nodes` here to make sure init node already has latest membership

arbulu89 (changes requested):
@liangxin1300 Good job.
I think that we need to improve the locking message. It is not totally atomic. If two nodes start in the same exact time we will have an issue.
I think we need to have an extra check here.

crmsh/join_lock.py
```diff
@@ -0,0 +1,139 @@
...
+                bootstrap.error("Can't invoke {} on {}".format(self.UPDATE_CMD, self.target_node))
+            time.sleep(self.UPDATE_INTERVAL)
+
+    @contextmanager
+    def lock(self):
+        """
+        Create lock file on target node
+        """
+        # wait if other node was joining
+        self._wait()
```
arbulu89, line 118: We need to change some logic here. The `touch` (or maybe other system) must be the 1st thing we do. Otherwise, if 2 nodes start the lock in the same moment, both will pass and get the log.
We need to have something atomic.
Maybe we need to combine the `touch` with a `mkdir`. This second will fail if the folder is there. The touch is quite risky, as it works when the file doesn't exists and when it exists. And it is used after the wait.

arbulu89 (commented):
@liangxin1300 Still some doubts about the code.
I see potential issues in some places.
For example: We check `if res < self.UPDATE_INTERVAL:` to see if the lock is available or not.
What if the SSH connections are slow? If the SSH command used to write the timestamp takes some few seconds. Could this generate some issue?

What if the 2 machines don't have the same time? Is this scenario even possible?

Maybe we should use other comparison mechanism, like a counter. I guess you use the timestamp to see if there is progress or not, and the node didn't hang there. The counter should work in the same way, and without depending to external variables like the timestamp and network quality

I don't know, I see potential issues, and this feature will affect all of the `join` processes. 
It must be rock solid, and I see weak points in some places.

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        while not self.thread_stop_event.is_set():
+            try:
+                self._run(self.UPDATE_CMD)
+            except Exception as err:
+                bootstrap.print_error_msg(str(err))
+                pid = os.getpid()
+                os.kill(pid, signal.SIGQUIT)
+                break
+            time.sleep(self.UPDATE_INTERVAL)
```
liangxin1300, line 141: `self._run` with update timestamp shell command might raise any exceptions(maybe include ssh timed out)
My thinking was to terminate the whole join process.
But it's hard to catch child thread's exception in main thread without join(which will block the join process).
So I use signal to kill the join pid
Seems a bit complex, but it works

crmsh/join_lock.py
```diff
@@ -0,0 +1,134 @@
...
+        origin_timestamp = None
+
+        while not self._create_lock_dir():
+
+            result = self._get_last_timestamp()
+            if not result:
+                time.sleep(self.DETECT_INTERVAL)
+                continue
+
+            current_node, timestamp = result.values()
```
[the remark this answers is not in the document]
liangxin1300: Changed

crmsh/join_lock.py
```diff
@@ -0,0 +1,134 @@
...
+    UPDATE_INTERVAL = 10
+
+    def __init__(self, target_node):
+        """
+        Init function
+        """
+        self.thread_stop_event = None
+        self.target_node = target_node
+
+    def _run(self, cmd, ignore_error_str=None):
```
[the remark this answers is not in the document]
liangxin1300: Changed

liangxin1300:
> @liangxin1300 Still some doubts about the code.
> I see potential issues in some places.
> For example: We check `if res < self.UPDATE_INTERVAL:` to see if the lock is available or not.
> What if the SSH connections are slow? If the SSH command used to write the timestamp takes some few seconds. Could this generate some issue?

Code now will raise `TimeoutError` after 10s(less than detect interval) and terminate the whole join process

> 
> What if the 2 machines don't have the same time? Is this scenario even possible?

No, the timestamp in join lock file was always from 1 node
> 
> Maybe we should use other comparison mechanism, like a counter. I guess you use the timestamp to see if there is progress or not, and the node didn't hang there. The counter should work in the same way, and without depending to external variables like the timestamp and network quality
> 
> I don't know, I see potential issues, and this feature will affect all of the `join` processes.
> It must be rock solid, and I see weak points in some places.

@arbulu89 
I tried to improve it, for now, I can not find another way to implement this `lock` like scenarios between nodes.
If this does make sense, I will continue to adjust UT to make it pass

Anyway, thanks for your suggestions!

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+        self.target_node = target_node
+
+    def _run(self, cmd):
+        """
+        Run command on target node, consider specific exceptions
+        """
+        cmd_with_ssh = "ssh {} root@{} \"{}\"".format(self.SSH_OPTION, self.target_node, cmd)
+        rc, out, err = utils.get_stdout_stderr(cmd_with_ssh)
+        if rc != 0 and err:
+            if re.search("tail: cannot open .*: No such file or directory", err):
```
arbulu89, line 49: What if the systems are in other language and with this the error message?
Do we know if these messages are only in english?

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+        file_not_found_once = False
+        warned_once = False
+        origin_timestamp = None
+
+        while not self._create_lock_dir():
+
+            try:
+                result = self._get_last_timestamp()
+            # consider lock file might not be created yet
+            except FileNotFoundError:
```
arbulu89, line 104: This looks strange. Exceptions used as normal code flow

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+
+    def _run(self, cmd):
+        """
+        Run command on target node, consider specific exceptions
+        """
+        cmd_with_ssh = "ssh {} root@{} \"{}\"".format(self.SSH_OPTION, self.target_node, cmd)
+        rc, out, err = utils.get_stdout_stderr(cmd_with_ssh)
+        if rc != 0 and err:
+            if re.search("tail: cannot open .*: No such file or directory", err):
+                raise FileNotFoundError(err)
```
arbulu89, line 50: This methods looks really suspicious as it is using `Exceptions` as normal code flow. Exceptions should exceptions.
Besides that, it does much more than just running commands.

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+                os.kill(pid, signal.SIGQUIT)
+                break
+            time.sleep(self.UPDATE_INTERVAL)
+
+    @contextmanager
+    def lock(self):
+        """
+        Create lock directory on target node and start a thread to update timestamp
+        """
+        signal.signal(signal.SIGQUIT, signal_handler)
```
arbulu89, line 148: What is this `signal.signal` doing? I don't get it

arbulu89 (commented):
Hi @liangxin1300 
I still see really strange things on the code that I don't understand.
- Exceptions used in normal code flow
- This signal code piece
- System output messages used by the code

crmsh/join_lock.py
liangxin1300: That's the point, thanks for reminding me!
Let me re-thinking

crmsh/join_lock.py
liangxin1300: To catch SIGQUIT, which might from `_update_timestamp` child thread, then `signal_handler` raise RuntimeError("Error in JoinLock update timestamp thread, exit!") to terminate the whole join process

I did this because it's hard to catch child thread's exception in main thread without using thread.join(which will block the bootstrap join process).
So I use signal to kill the join pid
Seems a bit complex, but it works

crmsh/join_lock.py
MalloZup: yep https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+    def __init__(self, target_node):
+        """
+        Init function
+        """
+        self.thread_stop_event = threading.Event()
+        self.target_node = target_node
+
+    def _run(self, cmd):
+        """
+        Run command on target node, consider specific exceptions
```
MalloZup, line 44: in this function we should check for error code due to ssh.
The funcion should then imho return the error code of the `executed` command which is then  controlled by the place is executed.
So we replace this regex on output with err-code.

Example:

```python
  elif re.search("mkdir: cannot create directory .*: File exists", err):
                raise FileExistsError(err)
```
should be removed from this function and checked  in the appropriate place.

Now I personally dunno if `utils.get_stdout_stderr` function provides actually 2 return codes.
One of the `ssh` command and one for the `executed` command. this should imho the proper way to handle this

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+            origin_timestamp = timestamp
+
+            if not warned_once:
+                warned_once = True
+                bootstrap.warn("Node {} still joining, wait...".format(locked_node))
+
+            time.sleep(self.DETECT_INTERVAL)
+
+    def _update_timestamp(self):
+        """
```
MalloZup, line 128: as complete outsider of the project and rfc, I think it would be useful to have the rationale of the update timestamp function.

As I do understand right now, I think the update timestamp is used in a way to see if the Connection is still alive right?

crmsh/join_lock.py
```diff
@@ -0,0 +1,167 @@
...
+                    break
+                else:
+                    file_not_found_once = True
+                    time.sleep(self.DETECT_INTERVAL)
+                    continue
+
+            locked_node, timestamp = result.values()
+            if origin_timestamp:
+                res = int(timestamp) - int(origin_timestamp)
+                if res < self.UPDATE_INTERVAL:
```
nick-wang, line 116: what if the time is inconsistent and have minutes differents between the nodes?

nick-wang, line 116: what if the time between nodes have huge different like minutes?

nick-wang:
Just my two cents for the method to lock multiple nodes join together. The current solution with local file can slove 3 nodes issues, but hard to extend to more. 
For example, we have 4 nodes(or more) cluster, A and B are already in cluster, C and D are ready to join. May have problem if C join A, while concurrently D join B...

Other possible solutions for this, but also have drawbacks like extra dependencies or complex code.
1. Using distributed memory based database or libraries, like redis, zookeeper. It is easier to manage/implement, but will require extra dependencies. Also not sure how it works with 2 nodes cluster.
2. Create a process on the `joined` node to manage the lock, the process will exit when join process finished or after timeout.
The process should use single instance mode in the cluster, so it will check both localhost or other setuped nodes before creating. It just manage the lock without taking care of the join process, otherwise may need more code change of the logic.

Compare to this two solutions, the 2nd is better cause no extra dependency needed.

nick-wang:
What if the `join` process die(node reboot) without unlock(remove the dir on the init node) first?
In this case, no nodes can join the cluster.

stefanotorresi:
> What if the `join` process die(node reboot) without unlock(remove the dir on the init node) first?
> In this case, no nodes can join the cluster.

I wouldn't try to solve this problem in the join phase for now. How common would this be? I wouldn't expect nodes dying during join all the time. :wink: 
We could provide an additional `crmsh` command to clean-up the locks on the init node side, so if this happens, a user will need to clean the locks up and try again.

liangxin1300:
@arbulu89 @stefanotorresi @MalloZup @nick-wang 
I've removed `_update_timestamp` codes and related logic, the whole logic in `join_lock.py` looks quite simple now
1. Try to `mkdir` on init node, if success, do the join process; else, wait until success
2. Consider `255` as an customer `SSHError`, which might include ssh timed out exceptions

Other than that, consider the lock directory might exists for ever if join node died, I added this option to `crm cluster join` command:
```
-r, --remove-lock     Remove the join lock directory on init node
```
To remove the join lock directory to avoid blocked for ever

I will complete UT if this change make sense.

Thanks very much!

crmsh/join_lock.py
```diff
@@ -0,0 +1,81 @@
...
+        """
+        self.target_node = target_node
+
+    def _run(self, cmd):
+        """
+        Run command on target node, consider specific exceptions
+        """
+        cmd_with_ssh = "ssh {} root@{} \"{}\"".format(self.SSH_OPTION, self.target_node, cmd)
+        rc, out, err = utils.get_stdout_stderr(cmd_with_ssh)
+        if rc == 255:
```
arbulu89, line 40: This `255` should go in a constants to be more meaningful. What `255` does mean?

arbulu89 (commented):
@liangxin1300 Personally, I wouldn't add a new parameter. It looks risky, and the user might use it incorrectly.
What I would do instead, is to run a command that will destroy the created folder after a timeout (the timeout might be configured in the configuration file). We could put 2 minutes by default or so.

Something like.
1. When we get the lock. Together with creating the dir, we run:
```
(sleep 30;rm -rf JOIN_LOCK_DIR) &
```
And we get the `pid` of the process.

2. When we unlock, kill the process in background
```
kill pid
```

3. To be sure that we don't have side effects, the lock folder should be created in the `tmp` folder, so if something happens to the main node (reboot) for example, we don't have the lock file after the reboot

Opinions @nick-wang @MalloZup @stefanotorresi

crmsh/join_lock.py
liangxin1300: @arbulu89 
man ssh `ssh exits with the exit status of the remote command or with 255 if an error occurred`

liangxin1300:
@arbulu89 Instead of destroy the lock directory by the join node self, I propose wait in the `_lock_or_wait` loop on the `second join node`, to check whether beyond the wait timeout, if so, remove the lock directory and continue
That should be more simple
What do you think?

arbulu89:
> @arbulu89 Instead of destroy the lock directory by the join node self, I propose wait in the `_lock_or_wait` loop on the `second join node`, to check whether beyond the wait timeout, if so, remove the lock directory and continue
> That should be more simple
> What do you think?

Ok, it looks reasonable and would achieve the same goal, to have access to the lock after the timeout. 
And yes, it looks simpler.
This wouldn't require many changes, and it would have similar pros/cons.
Yes, go ahead with it.
Maybe we could add in the first warn message the whole timeout, so the user is aware of what is the maximum time he should wait

stefanotorresi:
I'm not quite clear if I understood correctly, but it doesn't sound right to me and I think this would bring us to square one.
If a node attempts to join, and they cannot acquire the lock, they just have to wait, because another join might be in progress.
If the waiting time exceeds a timeout, the waiting stops: the join failed, no further attempts will be retried.
Under no circumstance a join attempt should be able to override the lock by removing possibly existing locks, as this would defeat the purpose of the lock itself.
The timeout just serves the purpose to stop retrying the lock acquisition. It should not trigger a lock cleanup, because it comes from something that is trying to acquire the lock, not something that is holding the lock.
The cleanup should be only performed by the side that holds the lock.

Does this make sense to you folks?

arbulu89:
> I'm not quite clear if I understood correctly, but this doesn't sound right to me and I think it would bring us back to square one.
> If a node attempts to join, and they cannot acquire the lock, they just have to wait and retry, because another join might be in progress.
> If the waiting time exceeds a timeout, the retry loop stops: the join failed, no further attempts will be made.
> Under no circumstance a join attempt should be able to override the lock by removing possibly existing locks, as this would defeat the purpose of the lock itself.
> The timeout just serves the purpose to stop retrying the lock acquisition. It should not trigger a lock cleanup, because it comes from something that is trying to acquire the lock, not something that is holding the lock.
> The cleanup should only be performed by the side that holds the lock.
> 
> Does this make sense to you folks?

So, we are in the same, if something happens to the node that is joining (reboot?) and lock is not freed, nobody else will be able to join we will be in problems.

Somehow, the lock must be removed eventually. We can trigger that in the way I commented (it looks more complicated) or just on the join. The result would be exactly the same.

We just need to be really aware, that the lock is created in a 3rd node, so we need to have safe mechanism to rescue rare situations, like reboots or users stopping the join process in a hard way.

stefanotorresi:
That's why I proposed the solution of removing the locks manually with a dedicated command: it has to be a very deliberate action that should only be performed if one is _absolutely certain_ that there are no joins in progress.
If you try to automate this, you cannot make such guarantee, unless we implement a much more complicated distributed lock service.

liangxin1300:
@stefanotorresi @arbulu89 @MalloZup @zzhou1 
Well, already changed as:
1. Add `join_lock_timeout = 120` in core section in /etc/crm/crm.conf
2. And the minimum value for join_lock_timeout will be 120s(will raise ValueError if value < 120)
3. Will inform user as a warning if need wait: 
`"Other node still joining, wait at most {}s...".format(self.join_lock_timeout)`
4. When reached the timeout, remove the lock directory, and create the new lock directory

zzhou1:
Some thoughts:

- Keep rephasing the use case what we try to achieve, what could be over-engineered to react on the failure situation/corner case, where is the balance with the simplest code

- From the use case perspective, the assumption behind either the timeout or a clear action(eg. -r) is tring to solve the same problem: "to recycle the dir lock in brutal force when things go wrong."

- One step back, what situation makes us declare the join operation goes wrong?
  a) while joining, the node get rebooted. This is the same as Ctl+C any time during the joining phase. I don't think the current crmsh can solve it well. I incline to treat this as the corner case and no need spend too much effort to deal with it. The ideal solution must be very different than the joining design of the current crmsh.
  b) 2 minute timeout is not secure to declare the failure and recycle the lock. I did experience 10+ minutes joining process about 6 months ago due to the incorrect hypervisor configuration. Again, this is another corner case, and I incline to think `join_lock_timeout` is slightly error prone, confuses the situation of this corner case, and might waste the additional explanation to users.

- In summary of my view, 
  -- I've no good gut feeling to deal well with these corner cases due to the nature of crmsh currently. And, keep in mind the use case of this topic is solved. I agree the warning message "Other node still joining,..." is a must-have. But I'm less motivated to recycle the lock.
  -- If I have to chose one approach, I slightly prefer to the determinate action(eg. -r) since it really relies on the system admin to do trouble shooting, or recontruct the whole cluster from the scratch.

arbulu89:
>       -- If I have to chose one approach, I slightly prefer to the determinate action(eg. -r) since it really relies on the system admin to do trouble shooting, or recontruct the whole cluster from the scratch even.

Keep in mind that this goes in total contra direction with the `Make our tools usage simpler` statement. Adding more flags, more scenarios, and more things the admin must know won't make this happen.
I don't even know why the user should know that there is a lock mechanism. This is just an internal implementation detail, and exposing such flag is just a solution because we weren't able to solve it internally.

I know that there might not be any silver bullet here, but the most important thing, even over adding this feature, is that we don't create more unstable scenarios. Just fixing one to creating other is not a good deal.

@liangxin1300 I would suggest we stop implementing new things until we have the architecture defined. I see we are adding more and more commits everytime someone suggests a new thing.

Now we have much more data, we might need to join in a video call and concrete things.
What do you think?

zzhou1:
In my view, I can split the proposed code has two parts: 
1. one part is to have locking in place to protest the main use case for the bug report
2. the second part is to clean up the lock for the "corner cases"

> 
> Keep in mind that this goes in total contra direction with the `Make our tools usage simpler` statement. 

> Adding more flags, more scenarios, and more things the admin must know won't make this happen.

The most important part of this PR is the first part about locking. It does solve the main issue and make tool more friendly, and also doesn't make it more complex. I think we all agree this.

Debating is on the second part to clean up the lock. It is much less important since 
a) they are the corner cases 
b) and more important fact is as long as we run into these corner cases, cleaning up the lock in brutal force doesn't seems to clean up the mess behind the scene. This applies to both approaches, either join_lock_timeout, or "-r"

If we don't reach to the consensus, we could leave the second part. To merge the first part does makes sense to me.

Finally, I don't reject either approach for the second part. I believe you can find the right balance between simplicity (for both code and user expereince) and the perfectness.

stefanotorresi:
> Keep in mind that this goes in total contra direction with the `Make our tools usage simpler` statement. Adding more flags, more scenarios, and more things the admin must know won't make this happen.

This is not an simple problem, we can't expect to present an easy solution without a complex implementation, the complexity will always be there, somewhere.

I know the manual cleanup is far from perfect UX, but at least it ensures that the locking mechanism always works, and it makes the code simpler!
Having a timeout on the acquisition side is basically like having a "KICK TO OPEN" sign on the door.

Are you sure this is going to make things actually easier for the user?
What will happen when a node brute forces its way into joining when it shouldn't? Would this be a simple scenario for them to understand and recover from?
What's the likelihood of this to happen in the first place?

If we really want automatic lock releases, we need a proper, distributed lock server, which will require a lot of work.

MalloZup:
I agree with @zzhou1 on splitting.

1   one part is to have locking in place to protest the main use case for the bug report
2   the second part is to clean up the lock for the "corner cases"

I think as already mentioned before, we are stretching the limit of the design of `crmsh` which unfortunately  doesn't have a distributed model/consensu design ( SSH) which can properly solve this situation for all corner cases or complexity properly

Given the way of crmsh imho is used, I would not be against that if there is a lockfile we just error out and print 
``` another node is trying or tried to join. please ensure it and remove the lockfile manually or with -r command`.

Imho in this way we are much more simple as solution.

**Comment by arbulu89** [comment]
@stefanotorresi @MalloZup @zzhou1 @liangxin1300 
Be aware that most of these kind of issues with errors while the joining process starts in parallel are coming because users are trying to automatize the usage of `crmsh`. When users are running the join process manually, this issue is anecdotical, as they will just repeat the join after the failure, or they won't even run the command in parallel. So, I would even exclude this use case.

So, I think we need to centralize this on the scenarios where `crm join` is executed within automation projects. **This means, adding a new `-r` value most probably won't help anyway. We need a mechanism that fits in the root cause. I don't want users adding more workaround on this to their automation projects.**

That's why I suggest this kind of timeout operations, where the code can deal with the scenario, where to nodes are joining together, without user interaction. For this, a timeout (or similar technique) must exist in my opinion.

If even with this context, you all think that we should go with the `admin deals with it` approach, it's find, but I think we need to implement the solution based on the error.

**Comment by zzhou1** [comment]
In my view, as long as we get into the steps to "reclaim the lock", it "must be" something wrong already. "join_lock_timeout" not likely to clean up the mess. Instead, it makes it even worse.

I incline to claim these "corner cases", eg. rebooting or Ctl+C during rejoining as UNDEFINED before we have the concrete idea which is beyond this PR. 

If we count in the timeout of the joining procedure for automation for example, I value more on "join_timeout" to report error and exit instead of reclaiming the lock.

**Comment by nick-wang** [comment]
I think we can't find a perfect lock solution without a complex distributed locking mechanism. Add the `-r` option is more like a compromise solution for the simplicity, cause we can't implement this to fix the parallel join issue by adding another bug of corner case of abnormal exit when joining.
At least we get proceed the original unsolved issue to the `-r` command that admin can use to fix the exception. In an idea  world, won't need admin to interrupt.

If we agree to add more complexity, i would recommend an automatically method like @arbulu89 proposed to avoid the user to abuse the `-r` command in all cases. Also doesn't let user aware the `lock` which is an internal function that user should not know. 
But we need more comprehensive consideration for the `timeout`, at least for example:
1. The **joined** need to clear the lock file when starting. (this part is missing from the latest POC)
2. After a timeout calculated on **joined** node, the node should remove the lock file of previous **joining** node. (this part is included in POC)
3. Find a proper default **timeout** in code and doc, let user aware this and tuning in special cases

**Comment by liangxin1300** [comment]
@nick-wang @stefanotorresi @zzhou1 @arbulu89 @MalloZup 
To make a conclusion for our meeting and make a summary for the solution, please feel free to correct me if I'm wrong:

0. Add `join_timeout` in /etc/crm/crm.conf, detault value is 120s
1. Add this option on `crm cluster join` sub-command:
   ```
   -r, --remove-lock     Remove the join lock directory on init node
   ```

2. Each join node try to use `mkdir` on init node to claim lock
3. Race winner will continue the join process, then finally, remove the lock directory itself created
4. For other nodes just wait and try to claim the lock in interval, give a warning once like `Other node still joining, wait...`

5. **[corner case]** When on the join process, we should have a `timer` to measure whether the join process reached the value of `join_timeout`, if so, exit this join process with a TimeoutError, leave the lock directory `un-removed`,  let system admin to decide what to do

6. **[corner case]** When other nodes waiting more than this `join_timeout`, give a hint to user like `The node which is joining might have problem, please check`

**Comment by stefanotorresi** [comment]
@liangxin1300 that sums it up perfectly, thanks!!

**Comment by stefanotorresi** [comment]
At a glance, the workflow is close enough to what I expected. I'll leave the implementation review to others. :)
Nice work so far, by the way!

**Review by arbulu89 — CHANGES_REQUESTED** [review]
@liangxin1300 Some technical improvements

1- As @MalloZup commented, I would remove the usage of the named file.
It doesn't really add much value to know which node has the lock. In fact, I wouldn't restore the timeout. Otherwise, the timeout is not static, it's dynamic making things unclear. Better to increase the timeout if they have issues


2- The timeout system doesn't look good. We cannot use this `+1` system, and repeat all the ssh commands every repetition. We need something better. The correct approach would be:
- Before starting we get the current time using `time.time()`
- We get the timeout doing `current_time + self.join_timeout`
- The loop condition should be `while current_time <= timeout and not self._create_lock_dir:`
- At the end of the loop we update the `current_time`
- We add a more logical sleep, like 10 seconds
- If we get the lock, we run use `break` to finish the loop
- We add the `else` statement to the `while` (yes, it is strange hehe). This `else` is executed if we finish the loop without a break. Here we need to raise the error

File: crmsh/join_lock.py [hunk 20]
```diff
@@ -0,0 +1,145 @@
...
+            # might lose claiming lock again, start to wait again
+            if pre_join_node and pre_join_node != join_node:
+                count = 0
+                warned_once = False
+            else:
+                pre_join_node = join_node
+
+            # wait over join_timeout, exit
+            if count >= expired_count:
+                raise TimeoutError("Exit now, since the joining node({}) is ongoing more than {}s. {}".format(join_node, expired_count, expired_error_str))
```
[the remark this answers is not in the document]
liangxin1300: OK
Changed

crmsh/join_lock.py
```diff
@@ -0,0 +1,145 @@
...
+        """
+        Try to claim lock on init node,
+        wait if failed to claim
+        exit if reached the join_timeout
+        """
+        warned_once = False
+        join_node = None
+        pre_join_node = None
+        expired_count = self.join_timeout
+        expired_error_str = "Joining process can not continue since the lock directory exists at the init node({}:{})".format(self.init_node, self.JOIN_LOCK_DIR)
```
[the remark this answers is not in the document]
liangxin1300: OK
Changed

crmsh/join_lock.py
```diff
@@ -0,0 +1,145 @@
...
+
+    @property
+    def join_timeout(self):
+        """
+        Get join_timeout from config.core
+        """
+        try:
+            value = int(config.core.join_timeout)
+        except ValueError:
+            raise ValueError("Invalid format of core.join_timeout(should be number)")
```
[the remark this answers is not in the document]
liangxin1300: OK
Changed

crmsh/join_lock.py
```diff
@@ -0,0 +1,145 @@
...
+        self._touch_my_name()
+
+    @contextmanager
+    def lock(self):
+        """
+        Create lock directory on target node
+        """
+        try:
+            self._lock_or_wait()
+            yield
```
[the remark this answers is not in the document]
liangxin1300: Necessary to build the `lock` contextmanager

crmsh/join_lock.py
```diff
@@ -0,0 +1,154 @@
...
+                break
+
+            # Might lose claiming lock again, start to wait again
+            online_list = self._get_online_nodelist()
+            if pre_online_list and pre_online_list != online_list:
+                warned_once = False
+                self._reset_timer()
+                continue
+            else:
+                pre_online_list = online_list
```
liangxin1300, line 123: Assuming there were 3+ nodes join the init node in parallel, some nodes might lose claiming again after one of them successfully joined the cluster, since there were still at least 2 nodes try to join.
So, when we know the membership changed(by `_get_online_nodelist`), just reset timer and `warned_once` flag and continue looping

liangxin1300:
> @liangxin1300 Some technical improvements
> 
> 1- As @MalloZup commented, I would remove the usage of the named file.
> It doesn't really add much value to know which node has the lock. In fact, I wouldn't restore the timeout. Otherwise, the timeout is not static, it's dynamic making things unclear. Better to increase the timeout if they have issues
> 
> 2- The timeout system doesn't look good. We cannot use this `+1` system, and repeat all the ssh commands every repetition. We need something better. The correct approach would be:
> 
> * Before starting we get the current time using `time.time()`
> * We get the timeout doing `current_time + self.join_timeout`
> * The loop condition should be `while current_time <= timeout and not self._create_lock_dir:`
> * At the end of the loop we update the `current_time`
> * We add a more logical sleep, like 10 seconds
> * If we get the lock, we run use `break` to finish the loop
> * We add the `else` statement to the `while` (yes, it is strange hehe). This `else` is executed if we finish the loop without a break. Here we need to raise the error. Some docs: https://www.w3schools.com/python/gloss_python_while_else.asp

Thanks @arbulu89 , these are really great practice!

crmsh/join_lock.py
```diff
@@ -0,0 +1,154 @@
...
+        if rc != 0 and err:
+            raise RuntimeError(err)
+        return re.findall('[0-9]+ (.*) member', out)
+
+    def _reset_timer(self):
+        """
+        Reset timer
+        """
+        self.current_time = int(time.time())
+        self.timeout = self.current_time + self.join_timeout
```
arbulu89, line 82: This `self.timeout` should be initialized in the `init` method, as `None`. Otherwise is difficult to track

arbulu89 (approved):
@liangxin1300 Great work!

Anyway, some personal comments (you can decide to implement there or not hehe)
- I think that the methods `_reset_timer`, `_increase_timer` and `_timed_out` shouldn't exist, and this code should go in the `_lock_or_wait` itself. They don't help at all, and make the code difficult to understand and follow. And they are used just once. 
- I don't like the `timeout` refresh. I know that you have your reasons. If the timeout is refreshed every time a new join is joined, it is not the timeout we configured, it is something else. If I set 120s, I would expect that after this time the operation fails.

PR #680, opened 2020-11-16
Fix: bootstrap: disable corosync-qdevice if not configured

liangxin1300:
cluster init without qdevice should "systemctl disable corosync-qdevice" if it is enabled, and the similarly for cluster join


nick-wang (approved):
Approved!

PR #681, opened 2020-11-17
[crmsh-4.1] Add lock for multi nodes join in parallel

liangxin1300:
backport #677 

Different from master branch: in `join_lock.py`, to avoid circular-imports, import bootstrap in which function needed


nick-wang (approved):
Looks good to me.

PR #682, opened 2020-11-25
Fix: utils: improve disable_service and enable_service function

liangxin1300:
In #680, we should check whether `corosync-qdevice` is available firstly, to avoid error if that package didn't installed

So I've improved `ServiceManager.disable_service` and `ServiceManager.enable_service` to make the action more solid


nick-wang (approved):
Approved, except the property change is not a must for this issue. But it does looks better.

PR #683, opened 2020-11-30
Dev: bootstrap: check watchdog while already in use

liangxin1300:
## Problem
`bootstrap.check_watchdog` will return False when cluster was running with sbd.
That will cause confused result while using preflight check to check watchdog.

Or, I was wondering whether this case possible: some other process which was not related with HA stack was using /dev/watchdog, while using bootstrap to configure sbd will failed because `bootstrap.check_watchdog` return False.(@zzhou1 @gao-yan )

## Reason
`wdctl` will return these when this watchdog was using
```
wdctl: /dev/watchdog: watchdog already in use, terminating.
wdctl: cannot open /dev/watchdog: Device or resource busy
```

## Solution
1. Using `lsof` and then grep this watchdog device whether being used, if yes, return True
2. Use `wdctl` to verify

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1497,6 +1497,11 @@ def check_watchdog():
     Verify watchdog device. Fall back to /dev/watchdog.
     """
     watchdog_dev = detect_watchdog_device()
+    # In case this watchdog already in use
+    _, out, _ = utils.get_stdout_stderr("lsof")
+    if re.search(" {}\n".format(watchdog_dev), out):
+        return True
+    # not in use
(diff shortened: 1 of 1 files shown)


zzhou1 (changes requested):
Keep rephrase the use cases. I can see two:
a) bootstrap to double check the designated `watchdog_dev` is free, and can be used by sbd.service
b) preflight check tries to verify if watchdog is used by sbd.service alone, not any other process.

They are slightly different. 

The current check_watchdog() in the main branch fits a) well, but not for b). 
The new proposed change fits b), but breaks a) in theory, though it is likely be doomed as a corner case. 

I feel the code is more clean beside more safe to add the new function for b), eg. verify_sbd_watchdog(), and leave the current check_watchdog() alone.

PR #684, opened 2020-12-06
[crmsh-3.0] Add lock for multi nodes join in parallel

liangxin1300:
backport #677 
Different from master branch: 
* in `join_lock.py`, to avoid circular-imports, import bootstrap in which function needed
* in `join_lock.py`, define TimeoutError class since it doesn't exist in python2.7


nick-wang (approved):
Looks good to me for the back port

PR #686, opened 2020-12-10
Dev: bootstrap: improve invoke to print error to give specific hints

liangxin1300:
## Problem
When meeting error on bootstrap process, error message might make user confused like:
```Shell
# crm cluster join -c crmsh15sp2-1
WARNING: No watchdog device found. If SBD is used, the cluster will be unable to start without a watchdog.
Do you want to continue anyway (y/n)? y
WARNING: Failed to detect firewall: Could not open ports tcp=30865|5560|7630|21064, udp=
  Generating SSH key
  Configuring SSH passwordless with root@crmsh15sp2-1
ERROR: cluster.join: Failed to run "cat /root/.ssh/id_rsa.pub | ssh -oStrictHostKeyChecking=no root@crmsh15sp2-1 'cat >> /root/.ssh/authorized_keys'"
```
User cannot find reason why this command failed

## Solution
In bootstrap.py, `invoke` was commonly used, it record log about what has executed, the stdout and the stderr after running command.

I change the return code of `invoke` function, return `(boolean, stdout, stderr)` tuple, print `stderr` when necessary.

Also for the important place like first ssh command in join process when swapping ssh key with init node, ssh might failed in **Azure** environment, so here I give more details, like:
```Shell
# crm cluster join -c crmsh15sp2-1 -y
WARNING: No watchdog device found. If SBD is used, the cluster will be unable to start without a watchdog.
WARNING: Failed to detect firewall: Could not open ports tcp=30865|5560|7630|21064, udp=
  Configuring SSH passwordless with root@crmsh15sp2-1
ERROR: cluster.join: Failed to append contents of /root/.ssh/id_rsa.pub to crmsh15sp2-1: root@crmsh15sp2-1: Permission denied (publickey).

    crmsh has no way to help you to setup up passwordless ssh among nodes at this time. 
    As the hint, likely, `PasswordAuthentication` is 'no' in /etc/ssh/sshd_config. 
    Given in this case, users must setup passwordless ssh beforehand, or change it to 'yes' and manage passwords properly
```

Besides that, for those place might only need return code of invoke, I changed the function `invokerc`, wrapped `invoke`, just return True/False

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -278,8 +278,9 @@ Configure SBD:
         if self.diskless_sbd:
             return
         for dev in self._sbd_devices:
-            if not invoke("sbd -d {} create".format(dev)):
-                error("Failed to initialize SBD device {}".format(dev))
+            rc, _, err = invoke("sbd -d {} create".format(dev))
+            if not rc:
+                error("Failed to initialize SBD device {}: {}".format(dev, err))
 
     def _update_configuration(self):
         """
@@ -336,12 +337,12 @@ Configure SBD:
             return
         if utils.service_is_enabled("sbd.service"):
             if self._get_sbd_device_from_config():
-                if not invoke("crm configure primitive stonith-sbd stonith:external/sbd pcmk_delay_max=30s"):
+                if not invokerc("crm configure primitive stonith-sbd stonith:external/sbd pcmk_delay_max=30s"):
                     error("Can't create stonith-sbd primitive")
-                if not invoke("crm configure property stonith-enabled=true"):
+                if not invokerc("crm configure property stonith-enabled=true"):
                     error("Can't enable STONITH for SBD")
             else:
-                if not invoke("crm configure property stonith-enabled=true stonith-watchdog-timeout=5s"):
+                if not invokerc("crm configure property stonith-enabled=true stonith-watchdog-timeout=5s"):
                     error("Can't enable STONITH for diskless SBD")
 
     def join_sbd(self, peer_host):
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -989,9 +987,15 @@ def append_to_remote_file(fromfile, remote_node, tofile):
...
+    crmsh has no way to help you to setup up passwordless ssh among nodes at this time. 
+    As the hint, likely, `PasswordAuthentication` is 'no' in /etc/ssh/sshd_config. 
+    Given in this case, users must setup passwordless ssh beforehand, or change it to 'yes' and manage passwords properly
+    """
     cmd = "cat {} | ssh -oStrictHostKeyChecking=no root@{} 'cat >> {}'".format(fromfile, remote_node, tofile)
-    if not invoke(cmd):
-        error("Failed to run \"{}\"".format(cmd))
+    rc, _, err = invoke(cmd)
+    if not rc:
+        error("Failed to append contents of {} to {}: {}\n{}".format(fromfile, remote_node, err, err_details_string))
```
zzhou1, line 998: This improvement is much more informative to the end user. One minor comment, how about to tweak the code a little:

error("Failed to append contents of {} to {}\n    \"{}\"\n{}".format(fromfile, remote_node, err, err_details_string))

hence, to put the raw error information in another dedicated line, ie,

ERROR: cluster.join: Failed to append contents of /root/.ssh/id_rsa.pub to crmsh15sp2-1
   "root@crmsh15sp2-1: Permission denied (publickey)."

zzhou1 (approved):

PR #687, opened 2020-12-10
Fix: bootstrap: make sure sbd device UUID was the same between nodes

liangxin1300:
## Problem
SBD devices might have the same name but different UUID between cluster nodes, that might lead to stonith action failed

## Solution

- During join process, for each configured sbd device, compare local UUID with UUID on init node
- For outer tools (like preflight check) want to verify sbd device on an existing cluster, calling `bootstrap.SBDManager.verify_sbd_device`, will raise `ValueError` when having this problem

@arbulu89 @gao-yan @zzhou1 Any suggestions?
Thanks!

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -201,7 +201,36 @@ Configure SBD:
         return result_list
 
     @staticmethod
-    def _verify_sbd_device(dev_list):
+    def _get_device_uuid(dev, node=None):
+        """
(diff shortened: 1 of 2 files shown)


gao-yan (commented):
The idea and the logic look sensible to me. Thanks!

liangxin1300:
@gao-yan Could you please approve this? Thank you!

gao-yan (approved):

PR #688, opened 2020-12-11
Low: bootstrap: check cluster was running on init node

liangxin1300:
I make this PR based on these thinking:
- Detecting whether init node has cluster service running, on logical side, shouldn't be the task inside `setup_passwordless_with_other_nodes` function
- Instead, we should do this detecting job just after finishing `join_ssh`, that is, detect after finishing swapping the ssh key with the init node
- Considering the `init lock` I mentioned yesterday, if on init process we create `/tmp/.crmsh_join_lock_directory`, join node will wait. But I don't want to use this `init lock` to protect unready init node, instead, just to protect extra qdevice setup process. 

@arbulu89 , what do you think?

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1935,12 +1935,6 @@ def setup_passwordless_with_other_nodes(init_node):
 
     Should fetch the node list from init node, then swap the key
     """
-    # Check whether pacemaker.service is active on init node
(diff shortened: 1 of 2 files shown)


arbulu89 (approved):
Thanks @liangxin1300 ,
It makes sense. We shouldn't create the lock if we already know that the cluster service is not up and running. This will save some operations in the init node.

About your last question. I think the `lock` should protect any action that might cause conflicts or errors. We should prevent joining nodes while we are still initializing and or adding qdevice (as this applies changes to the configuration files)

PR #689, opened 2020-12-13
Use class Watchdog to simplify watchdog config

liangxin1300:
## Motivation
see #714

## Changes & Solution

- Add **class Watchdog** to find valid watchdog device name
- Drop `-w` option on join side
- Option `-w` on init side could be a watchdog device name or driver name
```
-w WATCHDOG, --watchdog WATCHDOG
                        Use the given watchdog device or driver name
```
- How to use class Watchdog?
```Python
# On init side
self._watchdog_inst = Watchdog(_input=_context.watchdog)
self._watchdog_inst.init_watchdog()

# On join side
self._watchdog_inst = Watchdog(peer_host=peer_host)
self._watchdog_inst.join_watchdog()
```
- In class Watchdog, will parse these in order:

> 1. value of `-w` option
> 2. result from /etc/sysconfig/sbd (value of SBD_WATCHDOG_DEV)
> 3. the first valid/unused device from result of sbd query-watchdog
> 4. softdog

- If to handle a driver name, and that driver was not loaded now, run:
```
invoke("echo {} > /etc/modules-load.d/watchdog.conf".format(driver))
invoke("systemctl restart systemd-modules-load")
```

- If value of `-w` neither a valid device or driver name, exit with error:
`Should provide valid watchdog device or driver name by -w option`

- Drop original way/functions to verify/check watchdog

## User case on Azure
Node instance on Azure doesn't contain any watchdog device by default
- On init side, run
`crm cluster init -s <device> -w softdog -y` 
or,
`crm cluster init -s <device> -y`

> 1.  Then softdog will be loaded
> 2.  and its watchdog device will write to `/etc/sysconfig/sbd`
> 3.  Then continue sbd config process

- On join side, run
`crm cluster join -c <init_node> -y`

> 1.  File `/etc/sysconfig/sbd` will be synced, 
> 2. got watchdog device from this config file, 
> 3. found no this watchdog device exist, 
> 4. then query which driver name did this device using on init node by ssh command, 
> 5. then loaded it on join node
> 6. Then continue sbd config process

## Test rpm
https://build.opensuse.org/package/show/home:XinLiang:crmsh_testing/crmsh

@zzhou1 @arbulu89 @gao-yan @sjvudp Please take a look and give me feedback, thank you!

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -59,6 +59,7 @@ class Context(object):
         '''
         Initialize attributes
         '''
+        self.type = None # init or join
         self.quiet = None
         self.yes_to_all = None
         self.template = None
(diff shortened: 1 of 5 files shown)


zzhou1:
> Option -w could be a watchdog device name or driver name(on both init and join side)

-w is not applicable on the join node, right?

liangxin1300:
> > Option -w could be a watchdog device name or driver name(on both init and join side)
> 
> -w is not applicable on the join node, right?

Consider the compatibility, we should keep this option on join side
Now that, `-w` option on join side can also be used to configure watchdog device/driver, to handle the corner case that watchdog device/driver was different between nodes

zzhou1:
> > > Option -w could be a watchdog device name or driver name(on both init and join side)
> > 
> > 
> > -w is not applicable on the join node, right?
> 
> Consider the compatibility, we should keep this option on join side

I look this is a "false" assumption about "compatibility". I suspect only very few people use it as of now.

> Now that, `-w` option on join side can also be used to configure watchdog device/driver, to handle the corner case that watchdog device/driver was different between nodes

This is the more corner case. crmsh should not be too flexible as the UI tool in my view. The core value is to make the usage intuitive rather to create the chaos.

More important, crmsh is not designed to manage a heterogenous cluster configuration. 

To this very specific case, how could crmsh work with csync2 to manage cluster nodes with different /etc/sysconfig/sbd?

crmsh/bootstrap.py
```diff
@@ -149,6 +149,176 @@ def init_sbd_manager(self):
...
+    def _set_watchdog_info(self):
+        """
+        Set watchdog info through sbd query-watchdog command
+        Content in self._watchdog_info_dict: {device_name: driver_name}
+        """
+        rc, out, err = utils.get_stdout_stderr(self.QUERY_CMD)
+        if rc == 0 and out:
+            # output format might like:
+            #   [1] /dev/watchdog\nIdentity: Software Watchdog\nDriver: softdog\n
+            self._watchdog_info_dict = dict(re.findall("\[[0-9]\] (/dev/.*)\n.*\nDriver: (.*)", out))
```
arbulu89, line 217: Can we have `[11]` entry here? If this is possible the regular expression would fail. We could just add use `[0-9]+` I think

crmsh/bootstrap.py
```diff
@@ -149,6 +149,176 @@ def init_sbd_manager(self):
...
+        else:
+            error("Failed to run {}: {}".format(self.QUERY_CMD, err))
+
+    def _get_device_through_driver(self, driver_name):
+        """
+        Get watchdog device name which has driver_name
+        """
+        for k, v in self._watchdog_info_dict.items():
```
arbulu89, line 225: I would avoid using `k` and `v`, use more meaningfull names (`device`, `name`, maybe?)

crmsh/bootstrap.py
```diff
@@ -149,6 +149,176 @@ def init_sbd_manager(self):
...
+        return None
+
+    def _get_driver_through_device_remotely(self, dev_name):
+        """
+        Given watchdog device name, get driver name on remote node
+        """
+        cmd = "ssh -o StrictHostKeyChecking=no root@{} {}".format(self._peer_host, self.QUERY_CMD)
+        rc, out, err = utils.get_stdout_stderr(cmd)
+        if rc == 0 and out:
+            device_driver_dict = dict(re.findall("\[[0-9]\] (/dev/.*)\n.*\nDriver: (.*)", out))
```
arbulu89, line 237: This regular expression is repeated. Maybe better to add in a constant

crmsh/bootstrap.py
```diff
@@ -149,6 +149,176 @@ def init_sbd_manager(self):
...
+        return None
+
+    def _set_input(self):
+        """
+        If self._input was not provided by option:
+          1. Try to get it from sbd config file
+          2. Try to get the first valid device from result of sbd query-watchdog
+          3. Set the self._input as softdog
+        """
+        if not self._input:
```
arbulu89, line 261: Do we need that many `if not self._input` here? 
I guess we could just return as soon as we have a self._input, without having this strange `if` cases

crmsh/bootstrap.py
```diff
@@ -149,6 +149,176 @@ def init_sbd_manager(self):
...
+    @staticmethod
+    def _get_watchdog_device_from_sbd_config():
+        """
+        Try to get watchdog device name from sbd config file
+        """
+        conf = utils.parse_sysconfig(SYSCONFIG_SBD)
+        return conf.get("SBD_WATCHDOG_DEV")
+
+    @staticmethod
+    def _driver_was_loaded(driver):
```
arbulu89, line 201: This name is strange. Maybe better `_driver_is_loaded`. Past statements are not usually used

arbulu89 (commented):
@liangxin1300 The code and structure looks really good.
I have added some minor comments.

Finally, should be directly remove the `-w` option from the `join` process. Maybe, we should first send a message saying that now it is deprecated or similar. Just wondering, to not break already existing processes

crmsh/bootstrap.py
liangxin1300: Good catch!
Changed!

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: Changed!

crmsh/bootstrap.py
liangxin1300: Changed

crmsh/bootstrap.py
liangxin1300: Thanks for pointing this!
Changed

liangxin1300:
> @liangxin1300 The code and structure looks really good.
> I have added some minor comments.
> 
> Finally, should be directly remove the `-w` option from the `join` process. Maybe, we should first send a message saying that now it is deprecated or similar. Just wondering, to not break already existing processes

@arbulu89 Thanks for review!
Already finished UT codes for Watchdog class, and keeping `-w` option on join side and give a warn `WARNING: -w option is deprecated and will be removed in future versions`

arbulu89 (approved):
Fantastic @liangxin1300 
Great job

PR #694, opened 2020-12-15
Use bootstrap_lock.InitLock to protect steps after cluster running

liangxin1300:
## Problem
Since qdevice setup on init process will change corosync.conf, at this time, if other node want to join this node, that will cause undefined result like corosync.conf wouldn't be synced.

Ideally, init process should adopt a lock mechanism to protect qdevice setup process on init node, then any join node should waiting until init node totally finished

## Solution
By using the lock mechanism mentioned by #677 could resolve this problem.
To reuse the code and to consolidate the related conception, I've made these changes:

- Rename `join_lock.py` as `bootstrap_lock.py`
- Extract commonly used methods into base class `Lock`
- Init process use `LocalLock.lock`
- Join process use `RemoteLock.lock`


crmsh/bootstrap_lock.py
```diff
@@ -26,37 +26,24 @@ class JoinLock(object):
...
-    MIN_JOIN_TIMEOUT = 120
-    WAIT_INTERVAL = 10
 
-    def __init__(self, init_node):
+    def __init__(self, init_node=None):
         """
         Init function
         """
         self.init_node = init_node
+        self.is_joining = True if self.init_node else False
```
arbulu89, line 35: I wouldn't call this `joining`. I would call it `remote`. Basically we are implementing a remote lock now

crmsh/bootstrap_lock.py
```diff
@@ -67,6 +54,68 @@ def _create_lock_dir(self):
...
 
+    def unlock(self):
+        """
+        Remove the lock directory
+        """
+        if self.lock_owner:
+            self._run(self.RM_CMD)
+
+
+class InitLock(Lock):
```
arbulu89, line 65: I would rename this `LocalLock`

crmsh/bootstrap_lock.py
```diff
@@ -67,6 +54,68 @@ def _create_lock_dir(self):
...
+            yield
+        except RuntimeError as err:
+            bootstrap.error(str(err))
+        except:
+            raise
+        finally:
+            self.unlock()
+
+
+class JoinLock(Lock):
```
arbulu89, line 92: Rename to `RemoteLock`

crmsh/bootstrap_lock.py
```diff
@@ -67,6 +54,68 @@ def _create_lock_dir(self):
...
+    """
+    Class inherited from Lock class, to manage lock for multiple nodes join in parallel
+    """
+
+    MIN_JOIN_TIMEOUT = 120
+    WAIT_INTERVAL = 10
+
+    def __init__(self, init_node):
```
arbulu89, line 100: De we need this `__init__` method?

arbulu89 (changes requested):
@liangxin1300 It looks good in general.
I think we need to do some renaming.
I would rename to `LocalLock` and `RemoteLock`. They are not attached to specific processes, they are just generic locks, so having these names would be better if we need to reuse them in the future.

Besides that, I will that we need to have some specific testing for this cases. I don't know here, on in QA side, but these changes must be rock solid

crmsh/bootstrap_lock.py
liangxin1300: Changed

crmsh/bootstrap_lock.py
liangxin1300: Changed

crmsh/bootstrap_lock.py
liangxin1300: Changed

crmsh/bootstrap_lock.py
liangxin1300: No, already removed, thanks!

liangxin1300:
I appended behave test for multi node joining:
```
  Scenario: Multi nodes join in parallel
    Given   Cluster service is "stopped" on "hanode1"
    And     Cluster service is "stopped" on "hanode2"
    And     Cluster service is "stopped" on "hanode3"
    When    Run "crm cluster init -y" on "hanode1"
    Then    Cluster service is "started" on "hanode1"
    And     Show cluster status on "hanode1"
    When    Run "crm cluster join -c hanode1 -y" on "hanode2,hanode3"
    And     Cluster service is "started" on "hanode3"
    And     Online nodes are "hanode1 hanode2 hanode3"
    And     Show cluster status on "hanode1"
    And     File "/etc/corosync/corosync.conf" was synced in cluster
```

arbulu89 (commented):
@liangxin1300 I still see many mentions to `init` and `join` in docstrings and warning messages.
Maybe we should change them to be `local/remote`. If we need to have specific messages on top of them, we can capture them (maybe create specific Exceptions) and apply the specific messags.
What do you think?

test/features/bootstrap_bugs.feature
```diff
@@ -64,3 +64,18 @@ Feature: Regression test for bootstrap bugs
     And     Service "hawk.service" is "stopped" on "hanode2"
     When    Run "crm corosync get nodelist.node.ring0_addr" on "hanode1"
     Then    Expected "10.10.10.3" not in stdout
+
+  @clean
+  Scenario: Multi nodes join in parallel(bsc#1175976)
```
arbulu89, line 69: This looks great!!!
Maybe we would need an additional test to check the `joining` nodes wait if the `init` node is still in progress.
This way, we would cover all of the major functionalities that we were looking for with this lock mechanism

liangxin1300: For this scenario: ` Multi nodes join in parallel(bsc#1175976)`
Maybe we could find a way to let the join process insist/sleep over 120s, then other join node will exit
I'm not sure how to make this happen without changing source code...

liangxin1300:
> @liangxin1300 It looks good in general.
> I think we need to do some renaming.
> I would rename to `LocalLock` and `RemoteLock`. They are not attached to specific processes, they are just generic locks, so having these names would be better if we need to reuse them in the future.

@arbulu89 I suggest we return back to `InitLock` and `JoinLock`, since I think they were created to solve specific problems when in init/join scenarios, so they were bind so closely with specific logic, see `_lock_or_wait` function which logic was still complex and only match for join process.

In short, they might be hard to reuse for generic locks for other purpose

arbulu89:
> > @liangxin1300 It looks good in general.
> > I think we need to do some renaming.
> > I would rename to `LocalLock` and `RemoteLock`. They are not attached to specific processes, they are just generic locks, so having these names would be better if we need to reuse them in the future.
> 
> @arbulu89 I suggest we return back to `InitLock` and `JoinLock`, since I think they were created to solve specific problems when in init/join scenarios, so they were bind so closely with specific logic, see `_lock_or_wait` function which logic was still complex and only match for join process.
> 
> In short, they might be hard to reuse for generic locks for other purpose

@liangxin1300 Ok, we can do that, but I honestly find this a worst implementation.

The lock just protects some piece of code that is included within, it should be agnostic.
If for some reason we need to have specific things inside, this could inherit the code from the generic lock.

Some options: 
1 . We could just have `Lock`, `LocalLock` and `RemoteLock`. And if we need specifics `JoinLock` that inherits from `RemoteLock` and maybe changes the `wait` method.
2. We could just have `Lock`, `LocalLock` and `RemoteLock` and add them the used specific parameters from outside (like the timeout) and the reclaiming would be done outside too (as it stills changes the lock behaviour)

If you want, we can continue with your suggestion, and maybe have a follow up task to clean up things and make a better implementation.

liangxin1300:
@arbulu89 Here was the behave test to simulate join node timed out:
```
  @clean
  Scenario: Multi nodes join in parallel timed out(bsc#1175976)
    Given   Cluster service is "stopped" on "hanode1"
    And     Cluster service is "stopped" on "hanode2"
    And     Cluster service is "stopped" on "hanode3"
    When    Run "crm cluster init -y" on "hanode1"
    Then    Cluster service is "started" on "hanode1"
    And     Show cluster status on "hanode1"
    # Try to simulate the join process hanging on hanode2 or hanode2 died
    # Just leave the lock directory unremoved
    When    Run "mkdir /tmp/.crmsh_join_lock_directory" on "hanode1"
    When    Try "crm cluster join -c hanode1 -y" on "hanode3"
    Then    Except "ERROR: cluster.join: Join process failed after 120 seconds. Cannot continue since the lock directory exists at the init node (hanode1:/tmp/.crmsh_join_lock_directory)"
    When    Run "rm -rf /tmp/.crmsh_join_lock_directory" on "hanode1"
```

liangxin1300:
> > > @liangxin1300 It looks good in general.
> > > I think we need to do some renaming.
> > > I would rename to `LocalLock` and `RemoteLock`. They are not attached to specific processes, they are just generic locks, so having these names would be better if we need to reuse them in the future.
> > 
> > 
> > @arbulu89 I suggest we return back to `InitLock` and `JoinLock`, since I think they were created to solve specific problems when in init/join scenarios, so they were bind so closely with specific logic, see `_lock_or_wait` function which logic was still complex and only match for join process.
> > In short, they might be hard to reuse for generic locks for other purpose
> 
> @liangxin1300 Ok, we can do that, but I honestly find this a worst implementation.
> 
> The lock just protects some piece of code that is included within, it should be agnostic.
> If for some reason we need to have specific things inside, this could inherit the code from the generic lock.
> 
> Some options:
> 1 . We could just have `Lock`, `LocalLock` and `RemoteLock`. And if we need specifics `JoinLock` that inherits from `RemoteLock` and maybe changes the `wait` method.
> 2. We could just have `Lock`, `LocalLock` and `RemoteLock` and add them the used specific parameters from outside (like the timeout) and the reclaiming would be done outside too (as it stills changes the lock behaviour)
> 
> If you want, we can continue with your suggestion, and maybe have a follow up task to clean up things and make a better implementation.

Nice suggestions!
Thanks @arbulu89 , I will reconsider it

liangxin1300:
@arbulu89 For more generic using concern, I've made these changes:

- Define `lock_timeout` in crm.conf instead of `join_timeout`
- Rename `join_lock.py` as `lock.py`
- On init side, using
```
```

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+    RM_CMD = "rm -rf {}".format(LOCK_DIR)
+    SSH_TIMEOUT = 10
+    SSH_OPTION = "-o ConnectTimeout={} -o StrictHostKeyChecking=no".format(SSH_TIMEOUT)
+    SSH_EXIT_ERR = 255
+
+    def __init__(self, remote_node=None, lock_func=None):
+        """
+        Init function
+        """
+        self.remote_node = remote_node
```
arbulu89, line 44: This usage of `remote_node` looks out of place. Shouldn't be only implemented in the `RemoteLock` class. Otherwise it's pretty strange to implement other class

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+        self.lock_func = lock_func if lock_func else self._lock_or_fail
+        # only the lock owner can unlock
+        self.lock_owner = False
+
+    def _run(self, cmd):
```
arbulu89, line 49: I think this method should re-implemented in the `RemoteLock` class. This one, should only run the operations locally, and the re-implemented method in `RemoteLock` only in remote node, without this initial condition

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+
+    LOCK_DIR = "/tmp/.crmsh_lock_directory"
+    MKDIR_CMD = "mkdir {}".format(LOCK_DIR)
+    RM_CMD = "rm -rf {}".format(LOCK_DIR)
+    SSH_TIMEOUT = 10
+    SSH_OPTION = "-o ConnectTimeout={} -o StrictHostKeyChecking=no".format(SSH_TIMEOUT)
+    SSH_EXIT_ERR = 255
+
+    def __init__(self, remote_node=None, lock_func=None):
```
arbulu89, line 40: I don't think we need to pass `lock_func` here. This should be part of the class, and the `RemoteLock` just would re implement the method.

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+from . import utils
+from . import config
+from .msg import common_warn
+
+
+class SSHError(Exception):
+    """
+    Custom exception for ssh error
+    """
+    pass
```
arbulu89, line 18: We don't need this `pass`. Just remove it, the exception will work fine

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+class ClaimLockError(Exception):
+    """
+    Custom exception if claiming lock failed or wait lock release timed out
+    """
+    pass
```
arbulu89, line 25: Remove this `pass`

arbulu89 (changes requested):
@liangxin1300 I see some really strange things about inheritance here:
- `Lock` method shouldn't have any reference to `remote` values
- `_run` method should be different for both classes, each of them having their own logic. I doesn't makes sense to wrap everything in `Lock` class
- Each of them should have their own `lock_func`, and `RemoteLock` should re-implement the method, not pass it as a variable to the class. We are not using the power of inheritance otherwise, we are just hacking the classes

crmsh/lock.py
```diff
@@ -0,0 +1,169 @@
...
+
+class Lock(object):
+    """
+    A base class define a lock mechanism used to exclude other nodes
+    """
+
+    LOCK_DIR = "/tmp/.crmsh_lock_directory"
+    MKDIR_CMD = "mkdir {}".format(LOCK_DIR)
+    RM_CMD = "rm -rf {}".format(LOCK_DIR)
+    SSH_TIMEOUT = 10
```
arbulu89, line 36: I would say these `SSH` constants should be only available in the `RemoteLock` class. The local lock doesn't use SSH at all

liangxin1300: Changed

crmsh/lock.py
liangxin1300: Changed

crmsh/lock.py
liangxin1300: Changed

crmsh/lock.py
liangxin1300: Changed

crmsh/lock.py
liangxin1300: Changed

crmsh/lock.py
liangxin1300: Changed

liangxin1300:
> @liangxin1300 I see some really strange things about inheritance here:
> 
> * `Lock` method shouldn't have any reference to `remote` values
> * `_run` method should be different for both classes, each of them having their own logic. I doesn't makes sense to wrap everything in `Lock` class
> * Each of them should have their own `lock_func`, and `RemoteLock` should re-implement the method, not pass it as a variable to the class. We are not using the power of inheritance otherwise, we are just hacking the classes

Thank you @arbulu89 , good suggestions!

crmsh/bootstrap.py
```diff
@@ -2337,13 +2343,16 @@ def bootstrap_join(context):
...
-            join_cluster(cluster_node)
+        lock_inst = lock.RemoteLock(cluster_node)
+        try:
+            with lock_inst.lock():
+                setup_passwordless_with_other_nodes(cluster_node)
+                join_remote_auth(cluster_node)
+                join_csync2(cluster_node)
+                join_ssh_merge(cluster_node)
+                join_cluster(cluster_node)
+        except (lock.SSHError, lock.ClaimLockError) as err:
```
arbulu89, line 2354: I see that `ValueError` is raised in `RemoteLock`. I guess we need to capture it here too

arbulu89 (approved):
Excellent @liangxin1300 , I find a brilliant piece of code.

(Just have a look in one comment I did, I cannot leave any PR without comments hehe)

arbulu89:
Commit message of the PR changed afterwards. Have a look at: https://github.com/liangxin1300/crmsh/commits/20201216_init_lock

PR #695, opened 2020-12-16
Dev: bootstrap: use ping to test host is reachable before joining

liangxin1300:
Purpose to create this PR: to distinguish `host unreachable error` and some `ssh error` like `Permission denied (publickey)` raised later when `join_ssh`

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2631,4 +2631,13 @@ def disable_exception_traceback():
     sys.tracebacklimit = 0
     yield
     sys.tracebacklimit = default_value  # revert changes
+
+
+def ping_node(node):
+    """
+    Check if the remote node is reachable
+    """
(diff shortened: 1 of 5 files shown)


nick-wang (approved):
Good change!

crmsh/utils.py
```diff
@@ -2631,4 +2631,13 @@ def disable_exception_traceback():
     sys.tracebacklimit = 0
     yield
     sys.tracebacklimit = default_value  # revert changes
+
+
+def ping_node(node):
+    """
+    Check whether remote node reachable
```
arbulu89, line 2638: `Check if the remote node is reachable`

arbulu89 (approved):
Thanks @liangxin1300 
It looks good

liangxin1300:
Thank you @nick-wang @arbulu89 !

PR #716, opened 2020-12-30
Fix: ui_resource: change return code and error to warning for some unharmful actions

liangxin1300:


crmsh/ui_resource.py
```diff
@@ -496,7 +496,8 @@ def ms_to_sec(m):
             return m[:len(m)-3]
 
         if rsc not in compl.resources():
-            context.fatal_error("Resource {} not exists in this cluster".format(rsc))
+            context.warning("Resource {} not exists in this cluster".format(rsc))
+            return
```
gao-yan, line 500: This looks good to me.

crmsh/ui_resource.py
```diff
@@ -515,7 +516,7 @@ def ms_to_sec(m):
             import re
             failcount_res = re.findall(r'fail-count-{}#(.*)_([0-9]+)'.format(rsc), out)
             if not failcount_res:
-                context.fatal_error("No failcount on node {} for resource {}".format(node, rsc))
+                return
```
gao-yan, line 519: Strictly,  it probably should return success here only if the specified `value` is `0`.

gao-yan (commented):
Thanks for the changes, @liangxin1300. Some comments in below.

crmsh/ui_resource.py
liangxin1300: I think it's fine to return success here, since no failcount here is unharmful

gao-yan: Like said, it's not really an usual case that users manipulate failcounts and set them to non-zero value, but if crmsh happens to receive such a request and is refusing to set it to non-zero, it as least should tell users they are not getting the expected result .

liangxin1300: Thanks for pointing this @gao-yan ! Already changed as `return True if value and int(value) == 0 else False`

gao-yan (approved):

PR #722, opened 2021-01-12
[crmsh-3.0] Fix: bootstrap: make sure sbd device UUID was the same between nodes 

liangxin1300:
backport #687, all the same except add function `utils.list_cluster_nodes_except_me` from master branch


nick-wang (approved):
Done review for the backport. Looks good!

PR #723, opened 2021-01-13
Detect and correct SBD device to support Azure Site Recover

nick-wang:
# Background:

In Azure Site Recovery, device may change ID after fail over, which led to a SBD/clusterfailure.

# Solution:

Leverage the preflight check feature. Detect the SBD device inconsistent before starting. If possible, find a suitable candidate to replace the broken/obsolete SBD device with a valid device. This is fit to the Azure site recover scenario.

# Commits in this PR:

- SBD device detect function
- Ability to fix the SBD device path
- A fix to `--yes` option, which is not working in previous crmsh version
- Show `--help` when no input for the preflight check

# Test RPM:
[Test rpm for the SBD device correction feature](https://build.opensuse.org/package/show/home:wanghaisu:branches:network:ha-clustering:Factory/crmsh)

# How to use:
`/usr/share/crmsh/preflight_check/ha-cluster-preflight-check -f`

# Expected result:
```
$ /usr/share/crmsh/preflight_check/ha-cluster-preflight-check -f

============ Check the SBD device ============
[2021/01/13 03:46:20]INFO: Checking SBD device [Pass]
  WARNING: SBD device '/dev/disk/by-id/scsi-SLIO-ORG_IBLOCK_e5ce22b5-4b40-4abe-9d26-fcced0ac9646' is not exist.
  INFO: Found '/dev/disk/by-id/scsi-SLIO-ORG_IBLOCK_221aea97-baf2-4504-b51a-f5aeef1281c2' with SBD header exist.
==============================================
Case:                Replace SBD_DEVICE with new candidate
Original SBD device: /dev/disk/by-id/scsi-SLIO-ORG_IBLOCK_e5ce22b5-4b40-4abe-9d26-fcced0ac9646
New SBD device:      /dev/disk/by-id/scsi-SLIO-ORG_IBLOCK_221aea97-baf2-4504-b51a-f5aeef1281c2

Run (y/n)? y
[2021/01/13 03:46:25]INFO: SBD DEVICE change succeed
```

```
$ tail /etc/sysconfig/sbd 

SBD_OPTS=
SBD_DEVICE='/dev/disk/by-id/scsi-SLIO-ORG_IBLOCK_221aea97-baf2-4504-b51a-f5aeef1281c2'

```

# Limitation:

- Only support one SBD device in the current version
- Lack of unit test case


nick-wang:
Per discussion with Xin and Roger, we decided to perform `check` when no parameter input. 
Then use `-c` option to replace `-f` option.

preflight_check/task.py
```diff
@@ -558,3 +564,94 @@ def wait(self):
         if not result:
             self.thread_stop_event.set()
             # should be an error here
+
+
+class TaskFixSBD(Task):
```
liangxin1300, line 569: Better to add __doc__ to describe what if this class for

preflight_check/check.py
```diff
@@ -1,17 +1,149 @@
...
+    return candidates[index]
+
+
+def check_sbd():
+    """
+    Check the sbd device and find a possible fix for incorrect disk
+
+    Only support one path SBD_DEVICE in the current version
+    """
+    print("\n============ Check the SBD device ============")
```
liangxin1300, line 80: How about s/Check/Checking/, to keep consistent with other items

preflight_check/check.py
```diff
@@ -1,17 +1,149 @@
...
+        utils.msg_info("SBD configuration file {} not found.".format(config.SBD_CONF))
+        return ""
+
+    sbd_options = utils.conf_parser(config.SBD_CONF)
+
+    if not "SBD_DEVICE" in sbd_options:
+        utils.msg_info("SBD DEVICE not used.")
+        return ""
+
+    dev = sbd_options["SBD_DEVICE"]
```
liangxin1300, line 91: How about move above lines(81 to here) into block of line 95(after task_inst.run)?

preflight_check/check.py
```diff
@@ -1,17 +1,149 @@
...
+from . import config
+
+
+def fix(context):
+    """
+    Check configuration and fix the abnormal options
+    """
+    if context.check_conf:
+        can = check_sbd()
+        if can != "":
```
liangxin1300, line 20: I suggest here, we use a more clear name, instead of `can`, like `need_correct`, and the value of `need_correct` should be True/False, that is, we need also change the return type of `check_sbd`
@nick-wang  What do you think?

preflight_check/check.py
```diff
@@ -1,17 +1,149 @@
...
+    return num
+
+
+def find_candidate_sbd(dev):
+    """
+    Find the devices that already have SBD header
+
+    return the path of candidate SBD device
+    """
+    from glob import glob
```
liangxin1300, line 47: Better to move this line at the head of file

preflight_check/check.py
```diff
@@ -1,10 +1,144 @@
...
+    Check configuration and fix the abnormal options
+    """
+    if context.fix_conf:
+        can = check_sbd()
+        if can != "":
+            correct_sbd(context, can)
+    print()
+
+
+def _find_match_count(str1, str2):
```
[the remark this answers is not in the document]
nick-wang: Done

preflight_check/check.py
```diff
@@ -1,10 +1,144 @@
...
+    for i in range(leng):
+        if str1[i] == str2[i]:
+            num += 1
+        else:
+            break
+
+    return num
+
+
+def find_candidate_sbd(dev):
```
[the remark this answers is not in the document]
nick-wang: Done

preflight_check/config.py
```diff
@@ -5,3 +5,5 @@
 REMOVE_PORT = "firewall-cmd --zone=public --remove-port={port}/udp"
 ADD_PORT = "firewall-cmd --zone=public --add-port={port}/udp"
 FENCE_HISTORY = "stonith_admin -h {node}"
+SBD_CONF = "/etc/sysconfig/sbd"
+SBD_CHECK_CMD = "hexdump -C {dev} -n 7|head -1 |cut -d '|' -f 2"
```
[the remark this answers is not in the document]
nick-wang: Done

preflight_check/utils.py
```diff
@@ -232,3 +232,57 @@ def get_process_status(s):
...
+            value = value.strip("'")
+            value = value.strip('"')
+            value = value.strip()
+
+            result[l.split(sp)[0].strip()] = value
+
+        return result
+
+
+def is_valid_sbd(dev):
```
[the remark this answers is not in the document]
nick-wang: Can't use that function in here, cause it won't need to `_compare_device_uuid` in this case.

preflight_check/utils.py
```diff
@@ -232,3 +232,57 @@ def get_process_status(s):
             # a process may have died since we got the list of pids
             pass
     return False, -1
+
+
+def conf_parser(f, sp="="):
```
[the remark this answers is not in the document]
nick-wang: Done.

preflight_check/task.py
nick-wang: Done.

preflight_check/check.py
nick-wang: Done with correct name.
The return string need to used later, so can't use `True/False` instead.

nick-wang:
Add prefix `Dev:` and `Fix:` to the commits.

liangxin1300 (approved):
Great work @nick-wang 
Thanks for the new feature for ASR and the fix for preflight_check!

PR #729, opened 2021-01-21
Fix: hb_report: run lsof with specific ocfs2 device

liangxin1300:
## Problem
During dumping ocfs2 info in hb_report, running `lsof` directly might hanging for a long time,  and the most of output might have no related with ocfs2

## Solution
Run `lsof` with specific ocfs2 device

Test rpm link: https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_testing3/crmsh

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -1643,12 +1643,30 @@ def dump_D_process():
     return out_string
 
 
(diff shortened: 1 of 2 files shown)


nick-wang (approved):
Good change and code, also get positive feedback from customer.

PR #730, opened 2021-01-21
[crmsh-4.1] Fix: hb_report: run lsof with specific ocfs2 device

liangxin1300:
backport #729


nick-wang (approved):
Good change and code, also get positive feedback from customer.

PR #734, opened 2021-01-28
Add entry limitation for dump_ocfs2 function

liangxin1300:
Checking before dumping:

1. If has `mounted.ocfs2` command
2. if any ocfs2 partitions exist by checking output of `mounted.ocfs2`

RPM link: https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_testing6/crmsh

### DIFF
--- a/hb_report/utillib.py
+++ b/hb_report/utillib.py
@@ -1663,6 +1663,15 @@ def lsof_ocfs2_device():
 def dump_ocfs2():
     ocfs2_f = os.path.join(constants.WORKDIR, constants.OCFS2_F)
     with open(ocfs2_f, "w") as f:
+        rc, out, err = crmutils.get_stdout_stderr("mounted.ocfs2 -d")
+        if rc != 0:
+            f.write("Failed to run \"mounted.ocfs2 -d\": {}".format(err))
+            return
+        # No ocfs2 device, just header line printed
+        elif len(out.split('\n')) == 1:
+            f.write("No ocfs2 partitions found")
+            return
+
         f.write(dump_D_process())
         f.write(lsof_ocfs2_device())
 


nick-wang (commented):
@liangxin1300 Can we just run `rc, out, err = crmutils.get_stdout_stderr("mounted.ocfs2 -d")` directly without check `mount.ocfs2` exist first? 
I think the `rc != 0` when command `mount.ocfs2` not found, just write the error message can easier to debug.

liangxin1300:
> @liangxin1300 Can we just run `rc, out, err = crmutils.get_stdout_stderr("mounted.ocfs2 -d")` directly without check `mount.ocfs2` exist first?
> I think the `rc != 0` when command `mount.ocfs2` not found, just write the error message can easier to debug.

Make sense!
Already changed and synced with #737 , thank you @nick-wang !

hb_report/utillib.py
```diff
@@ -1663,6 +1663,15 @@ def lsof_ocfs2_device():
 def dump_ocfs2():
     ocfs2_f = os.path.join(constants.WORKDIR, constants.OCFS2_F)
     with open(ocfs2_f, "w") as f:
+        rc, out, err = crmutils.get_stdout_stderr("mounted.ocfs2 -d")
+        if rc != 0:
```
arbulu89, line 1667: How is this check working? I mean, what happens if this returns an error.
Should the code just continue?

liangxin1300: If running this command failed, error message will be recorded in `ocfs2.txt` then this function return, other functions not related with ocfs2 dumping continued

hb_report/utillib.py
```diff
@@ -1663,6 +1663,15 @@ def lsof_ocfs2_device():
...
+            f.write("Failed to run \"mounted.ocfs2 -d\": {}".format(err))
+            return
+        # No ocfs2 device, just header line printed
+        elif len(out.split('\n')) == 1:
```
liangxin1300, line 1671: If no ocfs2 device exist, the output will just one line:
```
# mounted.ocfs2 -d
Device Stack Cluster F UUID Label
```

nick-wang (approved):
Thanks for the change, looks good to me.

PR #737, opened 2021-01-28
[crmsh-4.1] Add entry limitation for dump_ocfs2 function

liangxin1300:
backport #734


nick-wang (approved):
Thanks for the change, looks good to me.

PR #741, opened 2021-01-31
Adjust qdevice configure process with diskless sbd

liangxin1300:
## Poblem
Configuring qdevice with diskless SBD during init process will cause init node rebooted
And, removing qdevice with diskless SBD will also got the node rebooted

## Reason
When setup qdevice, the current order was:

- Add qdevice configuration
- Run corosync reload (then quorum lost)
- Certification process (take a relative long time, then node got rebooted)
- Start qdevice service

When removing qdevice, the current order was:

- Stop qdevice service (then quorum lost)
- Remove configuration (the node got rebooted)
- Run corosync reload

In short, Diskless SBD require quorum, when reload qdevice configuration, quorum will lost for a while, then then node got rebooted

## Solution
**Process 1**. Setup qdevice with bootstrap init process from beginning
**Process 2**. Append qdevice on an existing running cluster via bootstrap qdevice stage

**For process 1**, as the qdevice was configuring from the beginning without any RA configured, it's no harm to **restart cluster** service when all configuration was ready; In this way, diskless SBD service will also restated, without any risk to fence the node

**For process 2**, we restrict qdevice stage must be deployed on a 2+ nodes cluster, otherwise, crmsh will raise an error like "**Cannot configure qdevice stage on a single node cluster**"

**For removal process**, we restrict command "crm cluster remove --qdevice" only works when 2+ nodes online, otherwise, for signle node cluster raise an error like "**Cannot remove qdevice service from a single node cluster**"

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -22,6 +22,7 @@ import shutil
 from string import Template
 from lxml import etree
 from pathlib import Path
+from enum import Enum
 from . import config
 from . import utils
 from . import xmlutil
@@ -45,8 +46,15 @@ PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
 USER_LIST = ["root", "hacluster"]
+QDEVICE_ADD = "add"
+QDEVICE_REMOVE = "remove"
 
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
+
+class QdevicePolicy(Enum):
+    QDEVICE_RELOAD = 0
+    QDEVICE_RESTART = 1
+    QDEVICE_RESTART_LATER = 2
 
 
 class Context(object):
@@ -79,6 +87,7 @@ class Context(object):
         self.qdevice_heuristics = None
         self.qdevice_heuristics_mode = None
         self.qdevice_rm_flag = None
+        self.qdevice_reload_policy = QdevicePolicy.QDEVICE_RESTART
         self.shared_device = None
         self.ocfs2_device = None
         self.cluster_node = None
@@ -1828,6 +1837,34 @@ Configure Administration IP Address:
     wait_for_resource("Configuring virtual IP ({})".format(adminaddr), "admin-ip")
 
 
+def evaluate_qdevice_quorum_effect(mode):
+    """
+    While adding/removing qdevice, get current expected votes and actual total votes,
+    to calculate after adding/removing qdevice, whether cluster has quorum
+    return different policy
+    """
+    quorum_votes_dict = utils.get_quorum_votes_dict()
+    expected_votes = int(quorum_votes_dict["Expected"])
+    actual_votes = int(quorum_votes_dict["Total"])
+    if mode == QDEVICE_ADD:
+        expected_votes += 1
+    elif mode == QDEVICE_REMOVE:
+        actual_votes -= 1
+
+    if utils.is_quorate(expected_votes, actual_votes):
+        # safe to use reload
+        return QdevicePolicy.QDEVICE_RELOAD
+    elif utils.has_resource_running():
+        # will lose quorum, and with RA running
+        # no reload, no restart cluster service
+        # just leave a warning
+        return QdevicePolicy.QDEVICE_RESTART_LATER
+    else:
+        # will lose quorum, without RA running
+        # safe to restart cluster service
+        return QdevicePolicy.QDEVICE_RESTART
+
+
 def init_qdevice():
     """
     Setup qdevice and qnetd service
@@ -1836,6 +1873,10 @@ def init_qdevice():
     if not _context.qdevice_inst:
         utils.disable_service("corosync-qdevice.service")
         return
+
+    if _context.stage == "qdevice":
+        utils.check_all_nodes_reachable()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD)
 
     status("""
 Configure Qdevice/Qnetd:""")
@@ -1874,8 +1915,15 @@ def start_qdevice_service():
 
     status("Enable corosync-qdevice.service in cluster")
     utils.cluster_run_cmd("systemctl enable corosync-qdevice")
-    status("Starting corosync-qdevice.service in cluster")
-    utils.cluster_run_cmd("systemctl start corosync-qdevice")
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        status("Starting corosync-qdevice.service in cluster")
+        utils.cluster_run_cmd("systemctl restart corosync-qdevice")
+    elif _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RESTART:
+        status("Restarting cluster service")
+        utils.cluster_run_cmd("crm cluster restart")
+        wait_for_cluster()
+    else:
+        warn("To use qdevice service, need to restart cluster service manually on each node")
 
     status("Enable corosync-qnetd.service on {}".format(qnetd_addr))
     qdevice_inst.enable_qnetd()
@@ -1895,7 +1943,8 @@ def config_qdevice():
         corosync.add_nodelist_from_cmaptool()
     status_long("Update configuration")
     update_expected_votes()
-    utils.cluster_run_cmd("crm corosync reload")
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        utils.cluster_run_cmd("crm corosync reload")
     status_done()
 
 
@@ -2596,10 +2645,14 @@ def remove_qdevice():
     if not confirm("Removing QDevice service and configuration from cluster: Are you sure?"):
         return
 
+    utils.check_all_nodes_reachable()
+    _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_REMOVE)
+
     status("Disable corosync-qdevice.service")
     invoke("crm cluster run 'systemctl disable corosync-qdevice'")
-    status("Stopping corosync-qdevice.service")
-    invoke("crm cluster run 'systemctl stop corosync-qdevice'")
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        status("Stopping corosync-qdevice.service")
+        invoke("crm cluster run 'systemctl stop corosync-qdevice'")
 
     status_long("Removing QDevice configuration from cluster")
     qnetd_host = corosync.get_value('quorum.device.net.host')
@@ -2607,8 +2660,15 @@ def remove_qdevice():
     qdevice_inst.remove_qdevice_config()
     qdevice_inst.remove_qdevice_db()
     update_expected_votes()
-    invoke("crm cluster run 'crm corosync reload'")
     status_done()
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        invoke("crm cluster run 'crm corosync reload'")
+    elif _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RESTART:
+        status("Restarting cluster service")
+        utils.cluster_run_cmd("crm cluster restart")
+        wait_for_cluster()
+    else:
+        warn("To remove qdevice service, need to restart cluster service manually on each node")
 
 
 def bootstrap_remove(context):
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2649,4 +2649,46 @@ def ping_node(node):
     rc, _, err = get_stdout_stderr("ping -c 1 {}".format(node))
     if rc != 0:
         raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+
+
+def is_quorate(expected_votes, actual_votes):
+    """
+    Given expected votes and actual votes, calculate if is quorated
+    """
+    return int(actual_votes)/int(expected_votes) > 0.5
+
+
+def get_stdout_or_raise_error(cmd):
+    """
+    Common function to get stdout from cmd or raise exception
+    """
+    rc, out, err = get_stdout_stderr(cmd)
+    if rc != 0:
+        raise ValueError("Failed to run \"{}\": {}".format(cmd, err))
+    return out
+
+
+def get_quorum_votes_dict():
+    """
+    Return a dictionary which contain expect votes and total votes
+    """
+    out = get_stdout_or_raise_error("corosync-quorumtool -s")
+    return dict(re.findall("(Expected|Total) votes:\s+(\d+)", out))
+
+
+def has_resource_running():
+    """
+    Check if any RA is running
+    """
+    out = get_stdout_or_raise_error("crm_mon -1")
+    return re.search("No active resources", out) is None
+
+
+def check_all_nodes_reachable():
+    """
+    Check if all cluster nodes are reachable
+    """
+    out = get_stdout_or_raise_error("crm_node -l")
+    for node in re.findall("\d+ (.*) \w+", out):
+        ping_node(node)
 # vim:ts=4:sw=4:et:
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -1386,8 +1386,8 @@ class TestBootstrap(unittest.TestCase):
     @mock.patch('crmsh.corosync.QDevice.enable_qnetd')
     @mock.patch('crmsh.utils.cluster_run_cmd')
     @mock.patch('crmsh.bootstrap.status')
-    def test_start_qdevice_service(self, mock_status, mock_cluster_run, mock_enable_qnetd, mock_start_qnetd):
-        bootstrap._context = mock.Mock(qdevice_inst=self.qdevice_with_ip)
+    def test_start_qdevice_service_reload(self, mock_status, mock_cluster_run, mock_enable_qnetd, mock_start_qnetd):
+        bootstrap._context = mock.Mock(qdevice_inst=self.qdevice_with_ip, qdevice_reload_policy=bootstrap.QdevicePolicy.QDEVICE_RELOAD)
 
         bootstrap.start_qdevice_service()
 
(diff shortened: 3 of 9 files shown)


test/unittests/test_bootstrap.py
```diff
@@ -1763,67 +1824,6 @@ def test_bootstrap_remove(self, mock_context, mock_init, mock_active,
...
-            mock.call("crm cluster run 'crm corosync reload'")
-            ])
-        mock_long.assert_called_once_with("Removing QDevice configuration from cluster")
-        mock_get_value.assert_called_once_with("quorum.device.net.host")
-        mock_qdevice.assert_called_once_with("node1")
-        mock_qdevice_inst.remove_qdevice_config.assert_called_once_with()
-        mock_qdevice_inst.remove_qdevice_db.assert_called_once_with()
-        mock_expected.assert_called_once_with()
-        mock_done.assert_called_once_with()
-
```
liangxin1300, line 1826: Remove duplicated unit test codes

arbulu89:
@liangxin1300 What is the technical limitation for solution 2? I mean, why cannot we start qdevice in 1 node cluster?

liangxin1300:
> What is the technical limitation for solution 2? 

No technical limitation, solution 2 require more (refactor)works and time

>  I mean, why cannot we start qdevice in 1 node cluster?

Current code flow was:
On a single node cluster:
1. Add qdevice configuration
2. Run corosync reload (then quorum lost)
3. Certification process (take a relative long time, then node got rebooted)
4. Start qdevice service

So starting qdevice in 1 node cluster will cause quorum lost for a short time, which will lead to node rebooted when configuring diskless SBD. This is a race condition which we should avoid it

arbulu89:
> So starting qdevice in 1 node cluster will cause quorum lost for a short time, which will lead to node rebooted when configuring diskless SBD. This is a race condition which we should avoid it

Is this a normal behaviour? Maybe there is something we could improve in qdevice/qnetd for that (talking without knowing much, to raise options to think).

About the code, I see that that the limitation is not only applied to `diskless sbd`, it is applied always. Is this intentional?

liangxin1300:
> > So starting qdevice in 1 node cluster will cause quorum lost for a short time, which will lead to node rebooted when configuring diskless SBD. This is a race condition which we should avoid it
> 
> Is this a normal behaviour? Maybe there is something we could improve in qdevice/qnetd for that (talking without knowing much, to raise options to think).

Fixing this from root cause, we should configure qdevice **in a static way**, the first step will need to configure **nodelist section** in mcast mode(more details see **solution two** in confluence page)

Right now it cannot be avoided unless restart the cluster service.

- To start qdevice.service, we should add qdevice related configuration in corosync.conf, and **reload** the configuration
- At the time when `reloading`, the `expected votes` become `2`, the actual votes was still `1`(from the single node), **so the cluster lost quorum this time**
- Then qdevice.service started to provide one vote to quorum system, the actual vote become `2`, **the cluster contain quorum again**


> 
> About the code, I see that that the limitation is not only applied to `diskless sbd`, it is applied always. Is this intentional?

Yes, it is not only for diskless sbd, the purpose was to avoid quorum lost during bootstrap process

gao-yan (commented):
Been thinking about the workflows ...

AFAICS, this pull request is basically to disallow configuring/removing of qdevice in the cases where reload of corosync.conf could trigger unexpected situations in terms of quorum.

I'm not against these as very temporary restrictions. But the restrictions are kind of confusing.  As we can see, it's really not the cases to blame. The main problem here is the "reload" of corosync config. And the fact is qdevice configuration has never been claimed by upstream to be "changable during the runtime".

Depending on how difficult it'd be, what I'd prefer to do right now is to just drop the "reload" of corosync config on configuring/removing of qdevice. Instead, if any cluster nodes are online, we ask users "Configuring/Removing of qdevice requires restart of cluster to get applied.  Do you want bootstrap to restart cluster for you during the procedure, otherwise to restart cluster yourself after the procedure?"

And for qdevice setup run at the beginning together with init, we could just restart cluster service without asking, or MORE reasonably,  start the cluster only after corosync.conf has been completely settled with the nodelist and qdevice  configuration, and qnetd has been started.

liangxin1300:
@gao-yan Thanks for review!

> Been thinking about the workflows ...
> 
> AFAICS, this pull request is basically to disallow configuring/removing of qdevice in the cases where reload of corosync.conf could trigger unexpected situations in terms of quorum.
> 
> I'm not against these as very temporary restrictions. But the restrictions are kind of confusing. As we can see, it's really not the cases to blame. The main problem here is the "reload" of corosync config. And the fact is qdevice configuration has never been claimed by upstream to be "changable during the runtime".
> 

From my understanding, `changable during the runtime` was meaningless when trying to change **existing** qdevice configuration, like change `quorum.device.algorithm` as another value during the runtime. crmsh never try to use `reload` to change these qdevice value in this way

But as we can see **from the actual fact**, **reload** does work when reloading corosync.conf with qdevice configuration from 0 to 1 when setup, or reloading corosync.conf without qdevice configuration from 1 to 0 then removing. That was what redhat did in pcs.

The side effect caused by `reload` was quorum lost when just have one single node online

> Depending on how difficult it'd be, what I'd prefer to do right now is to just drop the "reload" of corosync config on configuring/removing of qdevice. Instead, if any cluster nodes are online, we ask users "Configuring/Removing of qdevice requires restart of cluster to get applied. Do you want bootstrap to restart cluster for you during the procedure, otherwise to restart cluster yourself after the procedure?"
> 

Let me summarize all possible scenarios:

- Scenario 1: **setup qdevice with bootstrap init process**, don't use `reload`, just restart cluster service, it should no harm to restart since no RA configured on this fresh setup init node
- Scenario 2: **setup qdevice via stage on an existing 1 node cluster**, don't use `reload`, detect if any RA running:
    If False, like scenario 1, restart cluster service;
    If True, ask user to confirm like `Qdevice require cluster service restart` in interactive mode; For -y/--yes option(non-interactive mode), don't restart cluster service, finish the process leaving a warning like `Qdevice already configured, please restart cluster service`
- Scenario 3, **setup qdevice via stage on an existing 2+ nodes cluster**, keep current workflow to **reload**
- Scenario 4, **remove qdevice from an existing 2+ nodes cluster**, keep current workflow to **reload**
- Scenario 5, **remove qdevice on an existing 1 node cluster**, detect if any RA running:
    If False, restart cluster service;
    If True, ask user to confirm like `Qdevice require cluster service restart` in interactive mode; For -y/--yes option(non-interactive mode), don't restart cluster service, finish the process leaving a warning like `Qdevice already removed, please restart cluster service`

> Of course for qdevice setup run at the beginning together with init, we could just restart cluster service without asking, or MORE reasonably, start the cluster only after corosync.conf has been completely settled with the nodelist and qdevice configuration, and qnetd has been started.

gao-yan:
I think they are marked as "not changeable during the runtime" mainly because corosync cannot help you determine/predict the results/impacts of the reload.

Technically it's not just the case of single node. The points are:

* Dynamically adding qdevice, which means increasing the practical "expected votes", might make a quorate cluster partition inquorate.

For example in a 3-node cluster, only 2 nodes are online/reachable, adding qdevice will make it inquorate as least for a while before receiving vote(s)  from qnetd.

* Dynamically removing qdevice, which means loss of the vote(s) from qdevice/qnetd, might make a quorate cluster partition inquorate.

For example, in a 4-node cluster, only 2 nodes are online/reachable. Removing qdevice will make it inquorate.

So essentially, if we are to support dynamic adding/removing, we will need to be able to well evaluate the situations and determine for what cases qdevice could be dynamically added/removed safely.

A workflow I can think of could be something like:

1) First of all before proceeding anything, check and make sure all the configured nodes must be reachable via ssh from where `qdevice` stage is running, no matter how many of them are online in terms of clustering. Otherwise we could not get corosync.conf synchronized.

2) If all the configured nodes are offline (also for single node cluster), we just proceed with static adding/removing of qdevice.

3) Otherwise if only a single node has been configured and is online,  tell user: adding/removing of qdevice requires restart of cluster.

4) Otherwise for a cluster with more than one node and any nodes are online, we could choose from two policies here:

a)  The simple policy: 

* All the configured cluster nodes must be online in order to add/remove qdevice dynamically. 

* Otherwise if not all the configured nodes are online, tell user it requires restart of the whole cluster. 

b) The delicate policy:

* For adding, calculate if the votes that the cluster partition currently has will still be the majority assuming the practical "expect votes" is increased by adding of qdevice while it hasn't received vote(s) from qnetd.

* For removing, calculate if the votes that are owned by cluster nodes themselves will still be the majority assuming qdevice is removed.

* Otherwise likewise, tell user it requires restart of the whole cluster.

liangxin1300:
Thanks @gao-yan ! These thinking are really helpful, let me make some demos to verify

liangxin1300:
**Case 1**: **setup qdevice with bootstrap init process**, don't use reload, just restart cluster service, it should no harm to restart since no RA configured on this fresh setup init node

crmsh/bootstrap.py
```diff
@@ -45,6 +45,9 @@
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
 USER_LIST = ["root", "hacluster"]
+QDEVICE_RELOAD = 0
```
arbulu89, line 48: This looks really ugly @liangxin1300 ,
Better use something like (in fact the best would be to use `enum` but as it is not native in python2 it would bring compatibility issues):
```
QDEVICE_STATES = {
    'reload': 0,
    'restart': 1,
    'restart_later': 2
}
```

crmsh/bootstrap.py
```diff
@@ -1828,6 +1832,34 @@ def init_admin():
...
+def evaluate_qdevice_quorum_effect(mode):
+    """
+    While adding/removing qdevice, get current expected votes and actual total votes,
+    to calculate after adding/removing qdevice, whether cluster has quorum
+    return different policy
+    """
+    quorum_votes_dict = utils.get_quorum_votes_dict()
+    expected_votes = int(quorum_votes_dict["Expected"])
+    actual_votes = int(quorum_votes_dict["Total"])
+    if mode == "add":
```
arbulu89, line 1844: This `add`, `remove` string look really week. Use constants instead

crmsh/bootstrap.py
```diff
@@ -1837,6 +1869,10 @@ def init_qdevice():
         utils.disable_service("corosync-qdevice.service")
         return
 
+    if _context.stage == "qdevice":
+        utils.check_all_nodes_reachable()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect("add")
```
arbulu89, line 1874: I would replace this `add` by a constant

crmsh/bootstrap.py
```diff
@@ -2596,19 +2640,30 @@ def remove_qdevice():
     if not confirm("Removing QDevice service and configuration from cluster: Are you sure?"):
         return
 
+    utils.check_all_nodes_reachable()
+    _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect("remove")
```
arbulu89, line 2644: Change `remove` by a constant

crmsh/utils.py
```diff
@@ -2649,4 +2649,46 @@ def ping_node(node):
     rc, _, err = get_stdout_stderr("ping -c 1 {}".format(node))
     if rc != 0:
         raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+
+
+def is_quorate(expected_votes, actual_votes):
```
arbulu89, line 2654: I wonder if it wouldn't be better to calculate the `expected_votes` and `actual_votes` in this method directly.
Is there any reason you decided them to pass in the parameters?

crmsh/utils.py
```diff
@@ -2649,4 +2649,46 @@ def ping_node(node):
...
+    """
+    Given expected votes and actual votes, calculate if is quorated
+    """
+    return int(actual_votes)/int(expected_votes) > 0.5
+
+
+def get_stdout_or_raise_error(cmd):
```
arbulu89, line 2661: I think we already have many methods like `get_stdout...`. Did you consider changing any of them?
In any case, this new method works fine too

arbulu89: By the way, I would pass the `0` constant in the parameter being `0` default. Otherwise, you might have some command in the future that returns something different than `0` as good, and this method might not work properly

crmsh/utils.py
```diff
@@ -2649,4 +2649,46 @@ def ping_node(node):
...
+    """
+    rc, out, err = get_stdout_stderr(cmd)
+    if rc != 0:
+        raise ValueError("Failed to run \"{}\": {}".format(cmd, err))
+    return out
+
+
+def get_quorum_votes_dict():
+    """
+    Get current expect votes and total votes
```
arbulu89, line 2673: A would add here in the docstring the value of the returned dictionaly. Otherwise, we cannot really undestand what it is returning

arbulu89 (changes requested):
@liangxin1300 Added some comments about the code itself.
Let's wait for @gao-yan to check the logic of the change too

crmsh/bootstrap.py
liangxin1300: @arbulu89 Using `enum` is OK, since qdevice only exist on sle15

crmsh/utils.py
liangxin1300: @arbulu89 This methond is called by `bootstrap.evaluate_qdevice_quorum_effect`
**Before** adding qdevice on a running cluster or removing qdevice from a cluster, I should evaluate whether the adding/removing action will cause quorum lost
So I need to create a new method to calculate that result giving the current expected_votes and actual_votes from `corosync-quorumtool -s`

crmsh/utils.py
liangxin1300: > I think we already have many methods like `get_stdout...`. Did you consider changing any of them?
> In any case, this new method works fine too

@arbulu89 I don't know what's your point here? Do you mean I should give this function a new name not starts with `get_stdout`?

liangxin1300:
> @liangxin1300 Added some comments about the code itself.
> Let's wait for @gao-yan to check the logic of the change too

@arbulu89 Nice suggestions! Thank you! I will change them later on today

crmsh/utils.py
arbulu89: No. I'm just saying that we have already multiple methods running comands and getting stdout, stderr, etc, and now we are adding a new one. I was just thinking loud, that we could have a generic method that handles all of them configuring with the incoming parameters. 
But forget about this by now hehe

crmsh/bootstrap.py
liangxin1300: Changed as using `enum`

crmsh/bootstrap.py

crmsh/bootstrap.py

crmsh/bootstrap.py

crmsh/utils.py
liangxin1300: Changed

crmsh/bootstrap.py
```diff
@@ -2596,19 +2645,30 @@ def remove_qdevice():
...
+    utils.check_all_nodes_reachable()
+    _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_REMOVE)
+
     status("Disable corosync-qdevice.service")
     invoke("crm cluster run 'systemctl disable corosync-qdevice'")
-    status("Stopping corosync-qdevice.service")
-    invoke("crm cluster run 'systemctl stop corosync-qdevice'")
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        status("Stopping corosync-qdevice.service")
+        invoke("crm cluster run 'systemctl stop corosync-qdevice'")
```
gao-yan, line 2655: Probably we could move this into the later block:
```
    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
        invoke("crm cluster run 'crm corosync reload'")
```
and put it right before `crm corosync reload`?

crmsh/bootstrap.py
```diff
@@ -2596,19 +2645,30 @@ def remove_qdevice():
...
     qdevice_inst.remove_qdevice_db()
     update_expected_votes()
-    invoke("crm cluster run 'crm corosync reload'")
     status_done()
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        invoke("crm cluster run 'crm corosync reload'")
+    elif _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RESTART:
+        status_long("Restarting cluster service")
+        utils.cluster_run_cmd("crm cluster restart")
+        status_done()
```
gao-yan, line 2669: Is this `status _done()` duplicate given that there's one above? Maybe remove this one and move the common one to the end of the function?

gao-yan (commented):
@liangxin1300 Thanks for the nice work I really like where this goes.

crmsh/utils.py
```diff
@@ -2649,4 +2649,46 @@ def ping_node(node):
     rc, _, err = get_stdout_stderr("ping -c 1 {}".format(node))
     if rc != 0:
         raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+
+
+def is_quorate(expected_votes, actual_votes):
+    """
+    Given expected votes and actual votes, calculate if is quorated
+    """
+    return int(actual_votes)/int(expected_votes) > 0.5
```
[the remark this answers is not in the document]
liangxin1300: I think `int(actual_votes)/int(expected_votes) > 0.5` will be safe under python3:)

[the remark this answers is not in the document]
liangxin1300: I think `int(actual_votes)/int(expected_votes) > 0.5` will be safe under python3:)

crmsh/bootstrap.py
```diff
@@ -1874,8 +1915,15 @@ def start_qdevice_service():
 
     status("Enable corosync-qdevice.service in cluster")
     utils.cluster_run_cmd("systemctl enable corosync-qdevice")
-    status("Starting corosync-qdevice.service in cluster")
-    utils.cluster_run_cmd("systemctl start corosync-qdevice")
+    if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
+        status("Starting corosync-qdevice.service in cluster")
+        utils.cluster_run_cmd("systemctl start corosync-qdevice")
```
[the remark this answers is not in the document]
liangxin1300:
```
    # Start qdevice service if qdevice already configured
    if utils.is_qdevice_configured() and not confirm("Qdevice is already configured - overwrite?"):
        start_qdevice_service()
        return
```
When configuring qdevice and detecting already configured, if `n` input for confirm, above codes will be sure to run again
But I think systemctl itself will be smart enough to handle that, right?
At least, I've tried to start service manually on the already running service, nothing harmful happened

crmsh/bootstrap.py
liangxin1300: Actually I did this intentionally.
`Stopping qdevice service -> then remove configuration -> then reload`
If
`remove configuration-> stop service and reload`
corosync-qdevice will generate error logs since the configure already broken

crmsh/bootstrap.py
liangxin1300: Not duplicated. Previous `status_done` was for previous `status_long`, to print `done` after `description...`

liangxin1300:
> @liangxin1300 Thanks for the nice work! I really like where this goes.

Thanks @gao-yan 
I will complete UT and functional test quickly based on this

crmsh/bootstrap.py
gao-yan: Alright, a bit surprised a stopping service cares about the configuration.

crmsh/bootstrap.py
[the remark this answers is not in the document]
gao-yan: The thing is if corosync-qdevice.service is already started, we'd like it to restart so that it gets the changed corosync.conf applied.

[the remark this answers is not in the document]
liangxin1300: @gao-yan Fine, changed as restart!

liangxin1300:
@arbulu89 @gao-yan Complete with some UT and functional tests
Please review
Then we can merge this!
Thanks!

test/unittests/test_utils.py
```diff
@@ -1250,3 +1250,55 @@ def test_ping_node(mock_run):
         utils.ping_node("node_unreachable")
     assert str(err.value) == 'host "node_unreachable" is unreachable: error data'
     mock_run.assert_called_once_with("ping -c 1 node_unreachable")
+
+
+def test_is_quorate():
+    assert utils.is_quorate(3, 2) is True
+    assert utils.is_quorate(3, 1) is False
```
gao-yan, line 1257: Not necessary, but a test like `utils.is_quorate(4, 2) is False` would also make sense.

gao-yan (approved):
Looks good to me. Thank.

arbulu89 (approved):
@liangxin1300 Nice!. Thanks

PR #742, opened 2021-02-05
Run hb_report under hacluster

liangxin1300:
## Problem
It didn't work if run **hb_report** or **crm cluster copy** under **hacluster** user, which was required from Hawk

## Solution

- Change function `bootstrap.configure_local_ssh_key` and `bootstrap.swap_public_ssh_key` to accept non-root user parameter; That is, for common user, bootstrap process also setup password-less among cluster nodes
- Assign `/usr/bin/sh` to hacluster's login shell, otherwise, it's impossible to run command remotely
- On interactive mode, behavior about changing hacluster's login shell and setup ssh access will be asked for confirm
- For `runtime` existing cluster, need to use `ssh` stage to setup password-less access for hacluster
  * on init node, run `crm cluster init ssh -y`
  * on join node, run `crm cluster join ssh -c <init node> -y`
  
- For `runtime` existing cluster, `/etc/corosync/corosync.conf` and `/etc/sysconfig/sbd` might has 400 mod, which will cause `Permission denied` exceptions when running hb_report under hacluster, #747 try to resolve this, but not works for `runtime` existing cluster, to avoid that, before running hb_report under hacluster:
  * Run `chmod 644 /etc/corosync/corosync.conf`
  * Run `chmod 644 /etc/sysconfig/sbd`

Test rpm link: https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_testing7/crmsh

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -31,6 +31,7 @@ from . import tmpfiles
 from . import clidisplay
 from . import term
 from . import lock
+from . import userdir
 
 
 LOG_FILE = "/var/log/crmsh/ha-cluster-bootstrap.log"
@@ -42,10 +43,8 @@ SYSCONFIG_FW = "/etc/sysconfig/SuSEfirewall2"
 SYSCONFIG_FW_CLUSTER = "/etc/sysconfig/SuSEfirewall2.d/services/cluster"
 PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
-RSA_PRIVATE_KEY = "/root/.ssh/id_rsa"
-RSA_PUBLIC_KEY = "/root/.ssh/id_rsa.pub"
-AUTHORIZED_KEYS_FILE = "/root/.ssh/authorized_keys"
 SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
+USER_LIST = ["root", "hacluster"]
 
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "storage", "sbd", "cluster", "vgfs", "admin", "qdevice")
 
@@ -95,6 +94,7 @@ class Context(object):
         self.args = None
         self.ui_context = None
         self.interfaces_inst = None
+        self.with_other_user = True
         self.default_nic_list = []
         self.default_ip_list = []
         self.local_ip_list = []
@@ -1134,22 +1134,66 @@ def init_ssh():
     Configure passwordless SSH.
     """
     utils.start_service("sshd.service", enable=True)
-    configure_local_ssh_key()
+    for user in USER_LIST:
+        configure_local_ssh_key(user)
 
 
-def configure_local_ssh_key():
+def key_files(user):
+    """
+    Find home directory for user and return key files with abspath
+    """
+    keyfile_dict = {}
+    home_dir = userdir.gethomedir(user)
+    keyfile_dict['private'] = "{}/.ssh/id_rsa".format(home_dir)
+    keyfile_dict['public'] = "{}/.ssh/id_rsa.pub".format(home_dir)
+    keyfile_dict['authorized'] = "{}/.ssh/authorized_keys".format(home_dir)
+    return keyfile_dict
+
+
+def is_nologin(user):
+    """
+    Check if user's shell is /sbin/nologin
+    """
+    with open("/etc/passwd") as f:
+        return re.search("{}:.*:/sbin/nologin".format(user), f.read())
+
+
+def change_user_shell(user):
+    """
+    To change user's login shell
+    """
+    if user != "root" and is_nologin(user):
+        if not _context.yes_to_all:
+            status("""
+User {} will be changed the login shell as /bin/bash, and
+be setted up authorized ssh access among cluster nodes""".format(user))
+            if not confirm("Continue?"):
+                _context.with_other_user = False
+                return
+        invoke("usermod -s /bin/bash {}".format(user))
+
+
+def configure_local_ssh_key(user="root"):
     """
     Configure ssh rsa key locally
 
-    If /root/.ssh/id_rsa not exist, generate a new one
-    Add /root/.ssh/id_rsa.pub to /root/.ssh/authorized_keys anyway, make sure itself authorized
+    If <home_dir>/.ssh/id_rsa not exist, generate a new one
+    Add <home_dir>/.ssh/id_rsa.pub to <home_dir>/.ssh/authorized_keys anyway, make sure itself authorized
     """
-    if not os.path.exists(RSA_PRIVATE_KEY):
-        status("Generating SSH key")
-        invoke("ssh-keygen -q -f {} -C 'Cluster Internal on {}' -N ''".format(RSA_PRIVATE_KEY, utils.this_node()))
-    if not os.path.exists(AUTHORIZED_KEYS_FILE):
-        open(AUTHORIZED_KEYS_FILE, 'w').close()
-    append_unique(RSA_PUBLIC_KEY, AUTHORIZED_KEYS_FILE)
+    change_user_shell(user)
+
+    private_key, public_key, authorized_file = key_files(user).values()
+    if not os.path.exists(private_key):
+        status("Generating SSH key for {}".format(user))
+        cmd = "ssh-keygen -q -f {} -C 'Cluster Internal on {}' -N ''".format(private_key, utils.this_node())
+        cmd = utils.add_su(cmd, user)
+        rc, _, err = invoke(cmd)
+        if not rc:
+            error("Failed to generate ssh key for {}: {}".format(user, err))
+
+    if not os.path.exists(authorized_file):
+        open(authorized_file, 'w').close()
+    append_unique(public_key, authorized_file)
 
 
 def init_ssh_remote():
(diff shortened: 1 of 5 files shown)


MalloZup (commented):
seems good to me but not familiar with crmsh codebase

crmsh/bootstrap.py
```diff
@@ -1134,22 +1133,59 @@ def init_ssh():
...
-def configure_local_ssh_key():
+def key_files(user):
+    """
+    Find home directory for user and return key files with abspath
+    """
+    home_dir = userdir.gethomedir(user)
+    private_key = "{}/.ssh/id_rsa".format(home_dir)
+    public_key = "{}/.ssh/id_rsa.pub".format(home_dir)
+    authorized_file = "{}/.ssh/authorized_keys".format(home_dir)
+    return private_key, public_key, authorized_file
```
arbulu89, line 1148: I wouldn't return a 3 elements list here. This is a potential issue if for some reason you check the return data, as you will need to change all the places where the method is used.
Better to return a dictionary or specific object, so you can use just the needed elements

crmsh/bootstrap.py
```diff
@@ -1134,22 +1133,59 @@ def init_ssh():
     Configure passwordless SSH.
     """
     utils.start_service("sshd.service", enable=True)
-    configure_local_ssh_key()
+    for user in ["root", "hacluster"]:
```
arbulu89, line 1136: This `["root", "hacluster"]` list looks something that should go in a constant, as it is repeated later

crmsh/bootstrap.py
```diff
@@ -1134,22 +1133,59 @@ def init_ssh():
...
+            if not confirm("Continue?"):
+                _context.with_other_user = False
+                return
+        invoke("usermod -s /usr/bin/sh {}".format(user))
+
+    private_key, public_key, authorized_file = key_files(user)
+    if not os.path.exists(private_key):
+        status("Generating SSH key for {}".format(user))
+        cmd = "ssh-keygen -q -f {} -C 'Cluster Internal on {}' -N ''".format(private_key, utils.this_node())
+        if user != "root":
```
arbulu89, line 1180: Maybe adding the next code in a method is a good idea:
```
if user != "root":
  cmd = utils.add_su(cmd, user)
```
It is reused in different places

arbulu89 (commented):
@liangxin1300 Added some comments about the code.
About the purpose of the changes, I cannot really judge without knowing what are the consequences of changing the `hacluster` user bash option

crmsh/bootstrap.py
MalloZup: yop valid point

liangxin1300: Thanks @arbulu89 , already changed!

crmsh/bootstrap.py
liangxin1300: Already put it into `add_su` function

crmsh/bootstrap.py

MalloZup (approved):

gao-yan (changes requested):
It generally looks good to me. Thanks for the nice work, @liangxin1300!

Only one nitpick.

gao-yan:
> * For `runtime` existing cluster, need to use `ssh` stage to setup password-less access for hacluster
>   
>   * on init node, run `crm cluster init ssh -y`
>   * on join node, run `crm cluster join ssh -c <init node> -y`

Probably it could be a further improvement. To make it as friendly as possible, given that there's an existing cluster, it'd be great to only require running such a `ssh` stage on any of nodes, which takes care of all the nodes.

> * For `runtime` existing cluster, `/etc/corosync/corosync.conf` and `/etc/sysconfig/sbd` might has 400 mod, which will cause `Permission denied` exceptions when running hb_report under hacluster, #747 try to resolve this, but not works for `runtime` existing cluster, to avoid that, before running hb_report under hacluster:
>   
>   * Run `chmod 644 /etc/corosync/corosync.conf`
>   * Run `chmod 644 /etc/sysconfig/sbd`
> 

Also to make it as easy as possible for users, I'm thinking probably `ssh` stage running for an existing cluster should as well fix the permissions of any theses existing files?

crmsh/bootstrap.py
```diff
@@ -1134,22 +1134,59 @@ def init_ssh():
...
-    append_unique(RSA_PUBLIC_KEY, AUTHORIZED_KEYS_FILE)
+    if user != "root" and is_nologin(user):
+        if not _context.yes_to_all:
+            status("""
+User {} will be changed the login shell as /usr/bin/sh, and
+be setted up authorized ssh access among cluster nodes""".format(user))
+            if not confirm("Continue?"):
+                _context.with_other_user = False
+                return
+        invoke("usermod -s /usr/bin/sh {}".format(user))
```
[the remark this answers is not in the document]
liangxin1300: Since 
```
ls -al /usr/bin/sh
lrwxrwxrwx 1 root root 9 Sep 10  2019 /usr/bin/sh -> /bin/bash
```
```
# ls -al /bin/sh
lrwxrwxrwx 1 root root 4 Sep 10  2019 /bin/sh -> bash
```
I decide to change this as `/bin/bash`

liangxin1300:
> > * For `runtime` existing cluster, need to use `ssh` stage to setup password-less access for hacluster
> >   
> >   * on init node, run `crm cluster init ssh -y`
> >   * on join node, run `crm cluster join ssh -c <init node> -y`
> 
> Probably it could be a further improvement. To make it as friendly as possible, given that there's an existing cluster, it'd be great to only require running such an `ssh` stage on any of the nodes, which takes care of all the nodes.

Good suggestion!  Thanks @gao-yan 
> 
> > * For `runtime` existing cluster, `/etc/corosync/corosync.conf` and `/etc/sysconfig/sbd` might has 400 mod, which will cause `Permission denied` exceptions when running hb_report under hacluster, #747 try to resolve this, but not works for `runtime` existing cluster, to avoid that, before running hb_report under hacluster:
> >   
> >   * Run `chmod 644 /etc/corosync/corosync.conf`
> >   * Run `chmod 644 /etc/sysconfig/sbd`
> 
> Also to make it as easy as possible for users, I'm thinking probably `ssh` stage running for an existing cluster should as well fix the permissions of any theses existing files?

I agree that will be more convenient for this whole issue.
But ssh key and the permissions changing are two kinds of things for crmsh, is that proper to put them together in `ssh` stage?

gao-yan:
> > > * For `runtime` existing cluster, `/etc/corosync/corosync.conf` and `/etc/sysconfig/sbd` might has 400 mod, which will cause `Permission denied` exceptions when running hb_report under hacluster, #747 try to resolve this, but not works for `runtime` existing cluster, to avoid that, before running hb_report under hacluster:
> > >   
> > >   * Run `chmod 644 /etc/corosync/corosync.conf`
> > >   * Run `chmod 644 /etc/sysconfig/sbd`
> > 
> > 
> > Also to make it as easy as possible for users, I'm thinking probably `ssh` stage running for an existing cluster should as well fix the permissions of any theses existing files?
> 
> I agree that will be more convenient for this whole issue.
> But ssh key and the permissions changing are two kinds of things for crmsh, is that proper to put them together in `ssh` stage?

Technically, no :-)  As we can see, the "breaking" changes of hawk will bring  users  much inconvenience requiring them to re-run ssh stage.  I was thinking combing this into ssh stage would make their lives a bit easier :-)  But of course, we could either add this step also into update note / TID ...

liangxin1300:
> > > > * For `runtime` existing cluster, `/etc/corosync/corosync.conf` and `/etc/sysconfig/sbd` might has 400 mod, which will cause `Permission denied` exceptions when running hb_report under hacluster, #747 try to resolve this, but not works for `runtime` existing cluster, to avoid that, before running hb_report under hacluster:
> > > >   
> > > >   * Run `chmod 644 /etc/corosync/corosync.conf`
> > > >   * Run `chmod 644 /etc/sysconfig/sbd`
> > > 
> > > 
> > > Also to make it as easy as possible for users, I'm thinking probably `ssh` stage running for an existing cluster should as well fix the permissions of any theses existing files?
> > 
> > 
> > I agree that will be more convenient for this whole issue.
> > But ssh key and the permissions changing are two kinds of things for crmsh, is that proper to put them together in `ssh` stage?
> 
> Technically, no :-) As we can see, the "breaking" changes of hawk will bring users much inconvenience requiring them to re-run ssh stage. I was thinking combing this into ssh stage would make their lives a bit easier :-) But of course, we could either add this step also into update note / TID ...

@arbulu89 , How do you think about this?

liangxin1300:
> > > * For `runtime` existing cluster, need to use `ssh` stage to setup password-less access for hacluster
> > >   
> > >   * on init node, run `crm cluster init ssh -y`
> > >   * on join node, run `crm cluster join ssh -c <init node> -y`
> > 
> > 
> > Probably it could be a further improvement. To make it as friendly as possible, given that there's an existing cluster, it'd be great to only require running such an `ssh` stage on any of the nodes, which takes care of all the nodes.
> 
> Good suggestion! Thanks @gao-yan

@gao-yan After estimating, I think this change might be big and might break compatibility and bring regression.
So I suggest keep this unchanged and try to clarify these in `update note`

crmsh/bootstrap.py
```diff
@@ -1134,22 +1134,59 @@ def init_ssh():
...
+
+def is_nologin(user):
+    """
+    Check if user's shell is /sbin/nologin
+    """
+    with open("/etc/passwd") as f:
+        return re.search("{}:.*:/sbin/nologin".format(user), f.read())
+
+
+def configure_local_ssh_key(user="root"):
```
arbulu89, line 1161: I think this `configure_local_ssh_key` method has become too big and complex. 
We could split the logic in different submethods

arbulu89 (approved):
@liangxin1300 
I see correct the code changes. But I'm talking about python code itself.
As commented, I don't have that much context about the change, so I would wait until @gao-yan approval to move forward

crmsh/bootstrap.py
liangxin1300: Thanks @arbulu89 
I moved the change shell codes into `change_user_shell` function
And I will adjust UT codes later

gao-yan (approved):

liangxin1300:
Loop #1033 here

PR #743, opened 2021-02-08
[WIP] New configure parser

liangxin1300:
## Highlights compare with original solution

- **Reuse std library**

This PR: Using python standard library [tokenize](https://docs.python.org/3/library/tokenize.html);
Original: Implemented similar functionality by [corosync.corosync_tokenizer](https://github.com/ClusterLabs/crmsh/blob/master/crmsh/corosync.py#L672)

- **API: more simple**
```Python
# In this PR
from crmsh import corosync

inst = corosync.Parser()
inst.load_config_file()
# get value
inst.get("totem.cluster_name")
# set value
inst.set("nodelist.node.nodeid", "2")
# add new empty node entry under nodelist section
inst.add_section("nodelist.node")
# write back to config file
inst.write_to_file()
```
```Python
# Original
from crmsh import corosync

with open("/etc/corosync/corosync.conf") as f:
    data = f.read()
inst = corosync.Parser(data)
# get value
inst.get("totem.cluster_name")
# set value
inst.set("nodelist.node.nodeid", "3")
# add new empty node entry under nodelist section
inst.add('nodelist', corosync.make_section('node', []))
# write back to config file
with open("/etc/corosync/corosync.conf", 'w') as f:
    f.write(inst.to_string())
```
- **API: enable get/set value by index**

corosync.conf might has multi entries for `interface/nodelist`, original API has no way to get/set them by index; I think this feature should be more convenience for corosync3 since corosync3 might has more interface or nodelist to configure

```Python
# In this PR
from crmsh import corosync

inst = corosync.Parser()
inst.load_config_file()
# get value by index
inst.get("nodelist.node.nodeid", 1)
# set value, the first entry of nodelist by default
inst.set("nodelist.node.nodeid", "2")
# set value by index, the second entry of nodelist
inst.set("nodelist.node.nodeid", "3", 1)
```
```Python
# Original
from crmsh import corosync

with open("/etc/corosync/corosync.conf") as f:
    data = f.read()
inst = corosync.Parser(data)
# No way to get value by index, always get the first one
inst.get("nodelist.node.nodeid")
# No way to set value by index, always set the first one
inst.set("nodelist.node.nodeid", "3")
```
- **API: more pythonic**
```Python
# In this PR, consolidate related methods into one class
    @classmethod
    def get_value(cls, path, index=0):
        """
        Class method to get value
        Return None if not found
        """
        inst = cls()
        inst.load_config_file()
        return inst.get(path, index)

    @classmethod
    def get_values(cls, path):
        """
        Class method to get value list matched by path
        Return [] if not matched
        """
        inst = cls()
        inst.load_config_file()
        return inst.get_all(path)

    @classmethod
    def set_value(cls, path, value, index=0):
        """
        Class method to set value and write back to file
        """
        inst = cls()
        inst.load_config_file()
        inst.set(path, value, index)
        inst.write_to_file()
```
```Python
# Original
def get_value(path):
    f = open(conf()).read()
    p = Parser(f)
    return p.get(path)


def get_values(path):
    f = open(conf()).read()
    p = Parser(f)
    return p.get_all(path)


def set_value(path, value):
    f = open(conf()).read()
    p = Parser(f)
    p.set(path, value)
    f = open(conf(), 'w')
    f.write(p.to_string())
    f.close()
```
- **Exception handling**
```Python
# In this PR when having miss-configured braces
>>> from crmsh import corosync
>>> p = corosync.Parser()
>>> p.load_config_file()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/crmsh/corosync.py", line 762, in load_config_file
    self._verify_config_file()
  File "/usr/lib/python3.6/site-packages/crmsh/corosync.py", line 756, in _verify_config_file
    raise ValueError("Missing closing brace")
ValueError: Missing closing brace
```
Original: no error raised when having miss-configured braces
- **Unit test**

This PR: with defining more specific and small methods, the inner logic should be easier to test, so in this PR, we will archive good test coverage
Original: lack of unit test

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -843,8 +843,10 @@ class Parser(object):
         The contextmanager for convert between config dict and list
         """
         self._config_list = list(self._config_dict.items())
-        yield
-        self._config_dict = dict(self._config_list)
+        try:
+            yield
+        finally:
+            self._config_dict = dict(self._config_list)
 
     @staticmethod
     def _section_item(key):
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -656,6 +656,7 @@ class Parser(object):
 
     def __init__(self, config_file=None):
         """
+        Init function
         """
         self._config_file = config_file if config_file else conf()
         self._config_dict = {}
@@ -843,46 +844,43 @@ class Parser(object):
             return (key, "}")
         return (key+"__END", "}")
 
-    def add_section(self, section_path):
-        """
-        """
+    def _find_last_match(self, section_path):
         # totem.cluster_name -> error
         # totem.new_section  -> before
         # quorum.device      -> before
         # totem.interface    -> after
         # nodelist.node      -> after
         # nodelist           -> after
-        after = None
-        last_match = ""
-        for key in self._inner_section_end_list:
-            if self._matched_inner_section_end(section_path, key):
-                last_match = key
-                after = True
-                print("1111")
+        after_match_flag = True
+        two_levels = False
+        while True:
+            for key in self._inner_section_end_list[::-1]:
+                if self._matched_inner_section_end(section_path, key):
+                    return key, after_match_flag
 
-        # add section from root
-        if not last_match:
-            parent_section = '.'.join(section_path.split('.')[:-1])
-            for key in self._inner_section_end_list:
-                if self._matched_inner_section_end(parent_section, key):
-                    last_match = key
-                    print(last_match)
-                    after = False
-                    print("2222")
+            if two_levels:
+                raise ValueError("No section {} exist".format(section_path))
+
+            section_path = '.'.join(section_path.split('.')[:-1])
+            if not section_path:
+                after_match_flag = True
+                return self._inner_section_end_list[-1], after_match_flag
             else:
-                last_match = self._inner_section_end_list[-1]
-                print("3333")
-                after = True
+                two_levels = True
+                after_match_flag = False
+
+
+    def add_section(self, section_path):
+        """
+        """
+        last_match, after = self._find_last_match(section_path)
+        offset = 1 if after else 0
 
         with self._operate_config_list():
             key = self._unused_inner_key(section_path)
             index = self._config_list.index(self._section_end_item(last_match))
-            if not after:
-                self._config_list.insert(index, self._section_item(key))
-                self._config_list.insert(index + 1, self._section_end_item(key))
-            else:
-                self._config_list.insert(index+1, self._section_item(key))
-                self._config_list.insert(index+2, self._section_end_item(key))
+            self._config_list.insert(index+offset, self._section_item(key))
+            self._config_list.insert(index+1+offset, self._section_end_item(key))
 
     def _find_and_insert(self, inner_key, value):
         """
@@ -911,6 +909,14 @@ class Parser(object):
             if not self._is_inner_section(self._inner_section_name(key)):
                 raise ValueError("No section {} exist".format(self._section_name(key)))
             self._find_and_insert(key, value)
+
+    def remove(self, path, index=0):
+        key = self._inner_key(path, index)
+        del self._config_dict[key]
+
+    def remove_section(self):
+        with self._operate_config_list():
+            pass
 
     def write_to_file(self):
         """
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -860,7 +860,7 @@ class Parser(object):
 
     def _find_last_match(self, section_path):
         """
-        When adding seciton, need to find an anchor
+        When adding section, need to find an anchor
         Then the new section name will insert just after or before this anchor
 
         Example:
@@ -962,7 +962,7 @@ class Parser(object):
 
     def to_string(self):
         """
-        Use tokenize.untokenize to convert self._config_dit to string
+        Use tokenize.untokenize to convert self._config_dict into string
         """
         token_list = []
         for index, (key, value) in enumerate(self._config_dict.items()):
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -651,6 +651,8 @@ class QDevice(object):
 
 class Parser(object):
     """
+    Class to parse config file which syntax is similar with corosync.conf
+    Parse corosync.conf by default
     """
     INDENT = 4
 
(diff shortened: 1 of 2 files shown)


crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
+    Class to parse config file which syntax is similar with corosync.conf
+    Parse corosync.conf by default
+    """
+    INDENT = 4
 
+    def __init__(self, config_file=None):
+        """
+        Init function
+        """
+        self._config_file = config_file if config_file else conf()
```
tomschr, line 663: The `conf()` function (defined `crmsh/corosync.py`) uses the Corosync config file as string. However, that string is hardcoded in other places:

```
$ grep -n /etc/corosync/corosync.conf crmsh/*.py 
crmsh/bootstrap.py:1249:include /etc/corosync/corosync.conf;
crmsh/corosync.py:22:    return os.getenv('COROSYNC_MAIN_CONFIG_FILE', '/etc/corosync/corosync.conf')
crmsh/utils.py:1112:    elif os.path.exists("/etc/corosync/corosync.conf") or is_program('corosync-cfgtool'):
crmsh/utils.py:1675:        cfg = os.getenv('COROSYNC_MAIN_CONFIG_FILE', '/etc/corosync/corosync.conf')
crmsh/utils.py:1694:        cfg = os.getenv('COROSYNC_MAIN_CONFIG_FILE', '/etc/corosync/corosync.conf')
```

I know, it's not directly related to this PR. Maybe create a constant and import that.

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
-    "Create a token sequence representing a section"
-    if not contents:
-        contents = []
-    sp = path.split('.')
-    name = sp[-1]
-    for t in contents:
-        if t.path and not t.path.startswith(path):
-            raise ValueError("%s (%s) not in path %s" % (t.path, t.key, path))
-    return [Token(_tBEGIN, sp, name)] + contents + [Token(_tEND, [])]
+    @staticmethod
```
tomschr, line 667: In my daily work, a staticmethod in Python is not really useful. In most cases you can move it out of the class and use the method as a dedicated, separate function. The only reason to use it is for organizational purposes.

If you want to keep it, you should think about a classmethod. You should also change the calling of the method, for example:

```python
# in crmsh/corosync.py: 
 return inner_key in self._inner_section_name_list
```

to this:

```python
# in crmsh/corosync.py: 
 return inner_key in Parser._inner_section_name_list
```

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
+        """
+        Extract key name from an inner key
+        """
+        # nodelist.node.nodeid__1 -> nodelist.node.nodeid
+        return re.sub("__[0-9]+", "", inner_key)
 
-    def all_paths(self):
-        """Returns all value paths"""
+    @property
+    def _inner_section_name_list(self):
```
tomschr, line 720: I'm not sure if it make sense to define a property which is basically "pseudo-private". Usually, in most cases, properties should be public.
If you want to make this property public, remove the underscore in front of the function name.

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
+        """
+        Get list of inner section name
+        """
+        return [key for key, value in self._config_dict.items() if value == "{"]
+
+    @property
+    def _inner_section_end_list(self):
```
tomschr, line 727: Same here.

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
-            return
-        self._tokens = self._tokens[:i] + self._tokens[(e+1):]
+    @contextmanager
+    def _operate_config_list(self):
+        """
+        The contextmanager for convert between config dict and list
+        """
+        self._config_list = list(self._config_dict.items())
+        yield
+        self._config_dict = dict(self._config_list)
```
tomschr, line 847: You should be careful here. If you have a context manager, anything that is executed in the with block can raise an exception. If an exception is raised, you need to make sure some "teardown" code is executed properly.

The safest way is to use a `try... finally` construct:

```suggestion
        self._config_list = list(self._config_dict.items())
        try:
            yield
        finally:
            self._config_dict = dict(self._config_list)
```

Maybe it's not really needed, but this is usually the structure how a contextmanager should look like. See the example in [contextlib.contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager).

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
+    def _section_end_item(key):
+        if key.endswith("__END"):
+            return (key, "}")
+        return (key+"__END", "}")
+
+    def _find_last_match(self, section_path):
         """
-        Remove section which contains key: value
-        Used to remove node definitions
+        When adding seciton, need to find an anchor
```
tomschr, line 861: Small typo:

```suggestion
        When adding section, need to find an anchor
```

crmsh/corosync.py
```diff
@@ -669,215 +649,370 @@ def remove_qdevice_db(self):
...
-                if t.token == _tEND:
-                    s += '\n'
-                yield s
-                if t.token == _tBEGIN:
-                    indent += 1
-                last = t
-                tstream = tstream[1:]
-        return ''.join(joiner(self._tokens))
+        """
+        Use tokenize.untokenize to convert self._config_dit to string
```
tomschr, line 963: Small typo:

```suggestion
        Use tokenize.untokenize to convert self._config_dict into string
```

tomschr (commented):
@liangxin1300 Thanks for your efforts. :+1: 

Just a few comments below. Maybe consider the "private" and public methods. Not entirely sure if all really needs to be "private". Probably there is a reason, so I leave it to you.

Also the staticmethod vs. classmethods topic. It may be a matter of taste.

crmsh/corosync.py
liangxin1300: @tomschr Thanks for pointing this
Already added to my backlog

crmsh/corosync.py
liangxin1300: No, I don't want it to be public, like its name, it was used internally

crmsh/corosync.py
liangxin1300: Also, used internally

crmsh/corosync.py
liangxin1300: Thanks, changed!

crmsh/corosync.py
liangxin1300: Thanks, changed!

crmsh/corosync.py
liangxin1300: Yeah, the big reason for using staticmethod here is easy to read/find under a related class

crmsh/corosync.py
liangxin1300: Thanks, changed!

crmsh/corosync.py
arbulu89: This line could be improve with:
```
self._config_file = config_file or conf()
```

crmsh/corosync.py
```diff
@@ -669,215 +649,372 @@ def remove_qdevice_db(self):
...
-    def all_paths(self):
-        """Returns all value paths"""
+    def _unused_inner_key(self, path):
+        """
+        Generate unused inner key
+        """
+        index = 0
+        key = self._inner_key(path, index)
+        while key in self._config_dict:
+            index += 1
```
arbulu89, line 746: You should manage this `index` variable with the `enumerate` option.
https://www.tutorialspoint.com/enumerate-in-python

arbulu89 (commented):
Hi @liangxin1300 ,
In my opinion the code looks extremely difficult for the purpose it has.
I find the usage of `tokenize` not really suitable here.
I did some similar code to parse/change the corosync configuration file:
https://github.com/SUSE/salt-shaptools/blob/master/salt/states/crmshmod.py#L301

This code basically does:
1. Loads the corosync configuration file as a dictionary
2. There, we can do any operation, as managing dictionaries is simple.
3. We can dump the dictionary to a corosync configuration file style
Check the `_convert2dict` and `_convert2corosync` methods. With these basic methods, we can do everything. We could complement the `set/get` commands to use the dictionary

PR #752, opened 2021-02-25
Fix: bootstrap: raise warning when configuring diskless SBD with less than 3 nodes

liangxin1300:
## Problem
Using diskless SBD with less than 3 nodes was not recommend, it'd better if bootstrap giving related warning while configuring diskless SBD

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -348,6 +348,8 @@ Configure SBD:
   are a good choice.  Note that all data on the partition you
   specify here will be destroyed.
 """
+    DISKLESS_SBD_WARNING = """Diskless SBD requires cluster with three or more nodes.
+If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."""
 
     def __init__(self, sbd_devices=None, diskless_sbd=False):
         """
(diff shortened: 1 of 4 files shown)


crmsh/bootstrap.py
```diff
@@ -525,6 +527,8 @@ def sbd_init(self):
         if not self._sbd_devices and not self.diskless_sbd:
             invoke("systemctl disable sbd.service")
             return
+        if self.diskless_sbd:
+            warn(self.DISKLESS_SBD_WARNING)
```
liangxin1300, line 531: Always give warning while configuring diskless SBD on init process

crmsh/bootstrap.py
```diff
@@ -563,6 +567,10 @@ def join_sbd(self, peer_host):
         dev_list = self._get_sbd_device_from_config()
         if dev_list:
             self._verify_sbd_device(dev_list, [peer_host])
+        else:
+            vote_dict = utils.get_quorum_votes_dict(peer_host)
+            if int(vote_dict['Total']) < 2:
+                warn(self.DISKLESS_SBD_WARNING)
```
liangxin1300, line 573: On join side, when finding that init node's actual votes less than 2, give warning

gao-yan: It seems to make more sense to check the "Expected votes", which means whether the "configuration"/"setup" is going to meet the requirement.

gao-yan (commented):
Helpful warning. Thanks.

crmsh/bootstrap.py
liangxin1300: Changed, thanks!

gao-yan (approved):
Looks good. Thanks.

Issue #685, opened 2020-12-08
"crm node status" displays XML in newly created cluster


sjvudp:
After setting up a new cluster, `crm node status` displayed XML like this:
~~~lang-txt
<nodes>
      <node id="123" UNAME="h23"/>
      ...
</nodes>
~~~
Expected non-XML output.
Version was crmsh-4-2-0+git.1604052559.2a348644-5.26.1 of SLES 15 SP2.

liangxin1300:
Hi @sjvudp ,

`crm node status` show xml, this behavior was by designed(see `crm node help status`), although I personally don't know what is the initial purpose to make this sub-command

You can get the node status by `crm node show` or using `crm_mon` or `crm_node` commands

Thanks!

liangxin1300:
BTW, @sjvudp , what kind of results do you what to get?

sjvudp:
Maybe I actually was confused, but thinking that `crm configure show xml ...` can show the XML of the configuration, couldn't `crm node show xml` do what `crm node status` does?
When using `crm node status` I was just trying out commands, and I was surprised that the default output is XML. I had thought crm shell would avoid (i.e.: hide from the user) XML as much as possible ;-)

zzhou1:
Sounds like to me `crm node status` is unleashed to the end user before the code is finished, or said before the use case get decribed clear enough.

Maybe good to remove it from CLI at the moment?

PR #753, opened 2021-02-25
Remove crm node status subcommand

liangxin1300:
I think it's useless to leave a such subcommand just to show xml, which might increase confusing, see #685


zzhou1 (approved):

zzhou1 (approved):
Agree to drop this at the moment. We lack of doc to describe the use case to move this on further.

PR #763, opened 2021-03-01
Dev: utils: remove unused utils.cluster_stack and its related codes

liangxin1300:
I think under current live branches, no need to detect if the cluster stack is corosync


arbulu89 (approved):
Nice @liangxin1300 
Thanks

PR #766, opened 2021-03-08
Dev: create README.md file to introduce preflight check

liangxin1300:
Preview: https://github.com/liangxin1300/crmsh/blob/20210308_intro_preflight_check/preflight_check/README.md


liangxin1300:
@zzhou1 Please review when you have time, thanks!

zzhou1 (commented):
I think I more prefer to let `-h` explain this tool set by itself.

Maybe twak and rename the execution result part as `example-execution-result.txt`

PR #771, opened 2021-03-09
Low: bootstrap: change ha-cluster-bootstrap log path

liangxin1300:
backport #576


nick-wang (approved):
Looks good, approved.

PR #772, opened 2021-03-10
Change lock directory under /run

liangxin1300:
## Problem
**Just when** finished setup two nodes cluster with diskless SBD, I reboot node two **immediately**, then node one also be fenced by sbd.

After reboot, I found the lock directory not be deleted on node one, which should already be deleted when node two finished bootstrap join 
## Solution
Should run `sync` command after remove lock directory, otherwise, the lock directory might not be deleted in time

### DIFF
--- a/crmsh/lock.py
+++ b/crmsh/lock.py
@@ -28,7 +28,7 @@ class Lock(object):
     A base class define a lock mechanism used to exclude other nodes
     """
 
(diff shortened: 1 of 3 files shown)


liangxin1300:
## Updated Solution
Change the lock directory under /run, then the lock directory will removed automatically after reboot

arbulu89 (approved):

PR #773, opened 2021-03-10
[crmsh-4.1] Change lock directory under /run

liangxin1300:
backport #772

### DIFF
--- a/crmsh/lock.py
+++ b/crmsh/lock.py
@@ -28,7 +28,7 @@ class Lock(object):
     A base class define a lock mechanism used to exclude other nodes
     """
 
-    LOCK_DIR = "/tmp/.crmsh_lock_directory"
+    LOCK_DIR = "/run/.crmsh_lock_directory"
     MKDIR_CMD = "mkdir {}".format(LOCK_DIR)
     RM_CMD = "rm -rf {}".format(LOCK_DIR)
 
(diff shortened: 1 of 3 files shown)


crmsh/lock.py
```diff
@@ -30,7 +30,7 @@ class Lock(object):
 
     LOCK_DIR = "/tmp/.crmsh_lock_directory"
     MKDIR_CMD = "mkdir {}".format(LOCK_DIR)
-    RM_CMD = "rm -rf {}".format(LOCK_DIR)
+    RM_CMD = "rm -rf {};sync".format(LOCK_DIR)
```
zzhou1, line 33: Two comments,
1. sync is the system wide, likely overkill the whole system. I would avoid using it.
2. /run is gurenteed to be cleaned per boot, and fits your finding well. Seems, /run/crm is created by some one. If it is created by crmsh.rpm, then, it's safe to use, 
LOCK_DIR = "/run/crm/.crmsh_lock_directory"

liangxin1300: Good suggestions! Already tried to use `/run/crm/.crmsh_lock_directory`, seems work

liangxin1300:
@arbulu89 What do you think?

crmsh/lock.py
liangxin1300: @zzhou1 `/run/crm` is created just when pacemaker service starting, not related with crmsh

zzhou1: > @zzhou1 `/run/crm` is created just when pacemaker service starting, not related with crmsh

Then, it is is safe for crmsh to use it directly.

liangxin1300:
> Refer to the existing directory of /run, it does not need hide the inode LOCK_DIR.
> I prefer something like /run/crmsh_bootstrap_lock

@zzhou1  By design, lock.py was a common used file lock mechanism, so I think it's suitable to keep the original file name

arbulu89 (approved):

PR #775, opened 2021-03-11
Fix: bootstrap: sync files before finished joining

liangxin1300:
## Problem
I found that corosync.conf will become empty when reset init node just after join process finished, so the cluster service wouldn't be started on init node then.

How to reproduce it:
1. On init node: crm cluster init -u -i eth1 -i eth2 -y
2. On join node: crm cluster join -c <init_node> -y -i eth1 -i eth2 -y
3. Just when join process finished, quickly force reset the init node
4. After init node started, corosync.conf become empty

## Solution

- Before finishing join process, run `sync <files>` command between cluster nodes, to make sure all these files' contents sync to disk
- Use `utils.str2file` to write data into file, `utils.str2file` could execute `os.fsync` internally

## Other changes
Change more string values as constants

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -41,6 +41,8 @@ CSYNC2_KEY = "/etc/csync2/key_hagroup"
 CSYNC2_CFG = "/etc/csync2/csync2.cfg"
 COROSYNC_AUTH = "/etc/corosync/authkey"
 SYSCONFIG_SBD = "/etc/sysconfig/sbd"
+SYSCONFIG_PCMK = "/etc/sysconfig/pacemaker"
+SYSCONFIG_NFS = "/etc/sysconfig/nfs"
 SYSCONFIG_FW = "/etc/sysconfig/SuSEfirewall2"
 SYSCONFIG_FW_CLUSTER = "/etc/sysconfig/SuSEfirewall2.d/services/cluster"
 PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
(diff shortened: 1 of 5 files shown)


liangxin1300:
@arbulu89 @zzhou1 Please have a look, thanks!

zzhou1 (commented):
Great, I approve this nicer general approach to deal with this issue for all files we care, than just sync corosync.conf in the previous proposal. 

I'd like to leave the detailed python review to Xabi for the final appove :)

liangxin1300:
@arbulu89 Would you please take a look at  this?
Thank you!

crmsh/bootstrap.py
```diff
@@ -1258,29 +1267,17 @@ def init_csync2():
...
-include /etc/ha.d/ldirectord.cf;
-include /etc/lvm/lvm.conf;
-include /etc/multipath.conf;
-include /etc/samba/smb.conf;
-include /etc/sysconfig/nfs;
-include /etc/sysconfig/pacemaker;
-include /etc/sysconfig/sbd;
-include /etc/pacemaker/authkey;
-include /etc/modules-load.d/watchdog.conf;
+%s
```
arbulu89, line 1278: Maybe it would be better to use `format` here. But both work

crmsh/bootstrap.py
```diff
@@ -2239,6 +2236,18 @@ def setup_passwordless_with_other_nodes(init_node):
             swap_public_ssh_key(node, user)
 
 
+def sync_files_to_disk():
+    """
+    Sync file content to disk between cluster nodes
+    """
+    files_string = ""
```
arbulu89, line 2243: You could replace this by something like:
```
files_string = ' '.join(filter(lambda f: os.path.isfile(f), FILES_TO_SYNC))
```

arbulu89 (approved):
@liangxin1300 Looks good.
Added some python code comments. Feel free to change

crmsh/bootstrap.py
Thanks!

crmsh/bootstrap.py
Thanks!

PR #784, opened 2021-03-19
Fix: bootstrap: get the peer node name correctly

liangxin1300:
## Problem
When configuring an alias peer host name in /etc/hosts, crm cluster join/remove might failed

## Solution
Change the logic of get_cluster_node_hostname function, always using `crm_node --name` to get the peer node's host name


arbulu89 (approved):
@liangxin1300 Nice!
I thought that the hostname to get there was of the new node, but thank you for claryfing that this checks if the 1st node is still online.

PR #785, opened 2021-03-19
[crmsh-4.1] Fix: bootstrap: get the peer node name correctly 

liangxin1300:
backport #784


nick-wang (approved):
Same to the master branch, approved.

PR #789, opened 2021-03-23
Move preflight check codes from /usr/share/crmsh into crmsh directory

liangxin1300:
## Changes include:
- Remove preflight_check in Makefile.am
- Move `crmsh/preflight_check` into `crmsh/crmsh/preflight_check`
- Append preflight_check to `packages` in setup.py
- Import path change in all realted unit test codes
- Remove preflight_check from tox.ini for `--cov` option

Test RPM: https://build.opensuse.org/package/show/home:XinLiang:branches:preflight_check/crmsh


zzhou1 (approved):
You are the phython expert and make it phythonic well ;)

PR #790, opened 2021-03-24
Dev: ui_cluster: enable/disable corosync-qdevice.service

liangxin1300:
When running `crm cluster disable`, if exist enabled qdevice service, then after reboot, both corosync and corosync-qdevice will all started

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -136,7 +136,9 @@ class Cluster(command.UI):
         Enable the cluster services on this node
         '''
         try:
-            utils.enable_service("pacemaker")
+            utils.enable_service("pacemaker.service")
+            if utils.is_qdevice_configured():
+                utils.enable_service("corosync-qdevice.service")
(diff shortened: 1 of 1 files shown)


crmsh/ui_cluster.py
```diff
@@ -74,6 +74,7 @@ class Cluster(command.UI):
     - Perform other callouts/cluster-wide devops operations
     '''
     name = "cluster"
+    ENABLE_SERVICES = ("pacemaker.service", "corosync-qdevice.service")
```
zzhou1, line 77: I wonder we have to add the condition if corosync-qdevice is set in corosync.conf, or not. Blindly manage qdevice.service is a little too much for the cluster without qdevice.

liangxin1300: Never mind @zzhou1 
Inside function `utils.enable_service`, the logic will check if the service is available

zzhou1 (approved):

PR #791, opened 2021-03-25
Support parse space in sbd device name

liangxin1300:
## Problem

1. Not support space in value of `-s` option: `crm cluster init -s "/dev/sda1 " -y`
2. If has space between multi sbd devices in /etc/sysconfig/sbd, join process will raise an error:
`ERROR: cluster.join:  /dev/sdb1 doesn't look like a block device`
3. On init side, there was also an error if input space on interactive mode:
```
Path to storage device (e.g. /dev/disk/by-id/...), or "none" for diskless sbd, use ";" as separator for multi path []/dev/sda1; /dev/sdb1
ERROR:  /dev/sdb1 doesn't look like a block device
```
Anyway, it's too strict, should loose the limitation

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -350,6 +350,7 @@ Configure SBD:
 """
     DISKLESS_SBD_WARNING = """Diskless SBD requires cluster with three or more nodes.
 If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."""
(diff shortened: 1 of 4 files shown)


liangxin1300:
@gao-yan Could you please take a look at this? Thanks!

gao-yan (approved):
Looks good to me. Thanks.

PR #797, opened 2021-04-01
Support --all or specific node to manage cluster and nodes

liangxin1300:
## Motivation
- To start and stop the pacemaker/corosync stack at the whole cluster level
- To avoid resources migration/node reset during the whole cluster shutdown procedure
## Changes
- Dev: ui_cluster: Support multi sub-commands with --all option or specific node
  * crm cluster start/stop/restart [--all | \<node\>... ]
  * crm cluster enable/disable [--all | \<node\>... ]
  * crm node standby/online [--all | \<node\>... ]
- Dev: doc: Consolidate help info for those sub-command using argparse
- Stop cluster procedure:
  1. When dlm configured and quorum is lost, set enable_quorum_fencing=0, enable_quorum_lockspace=0 for dlm config option,
     to avoid dlm hanging
  2. Stop pacemaker since it can make sure cluster has quorum until stop corosync
  3. Then, stop qdevice if is active
  4. At last, stop corosync
- Standby/online action for nodes:
  To avoid race condition for --all option, melt all standby/online values into one cib replace session


crmsh/ui_cluster.py
```diff
@@ -91,78 +138,86 @@ def __init__(self):
...
+        node_list = parse_option_for_nodes(context, *args)
+        for node in node_list[:]:
+            if not utils.service_is_active("corosync.service", remote_addr=node):
+                logger.info("Cluster services already stopped on {}".format(node))
+                node_list.remove(node)
+        if not node_list:
+            return
+
+        # When dlm configured and quorum is lost, before stop cluster service, should set
+        # enable_quorum_fencing=0, enable_quorum_lockspace=0 for dlm config option
```
zzhou1, line 173: 1. Make sense to me to split the code to a separate commit to deal with the dlm node when the cluster is inquorate.
2. also another new commit to add robustness  to do_stop() for https://github.com/ClusterLabs/crmsh/issues/872

zzhou1: > > another new commit to add robustness to do_stop() for #872
> @zzhou1 Can't get your point here
I added a simple reproducer in #872

crmsh/bootstrap.py
```diff
@@ -688,7 +689,7 @@ def start_pacemaker():
             SBDManager.is_delay_start():
         pacemaker_start_msg += "(waiting for sbd {}s)".format(SBDManager.get_suitable_sbd_systemd_timeout())
     with logger_utils.status_long(pacemaker_start_msg):
-        utils.start_service("pacemaker.service", enable=True)
+        utils.start_service("pacemaker.service", enable=True, node_list=node_list)
```
zzhou1, line 692: It's a bit tricky here when start --all or multiple nodes.

sle15sp3-1:~ # time crm cluster start --all
INFO: Starting pacemaker(waiting for sbd 36s)...done    
<--- This above line is doesn't match the real "time" consumed, and likely confuse users.
       It does not fit status_long() context. Maybe slightly different wording, ie. 
       INFO: Started pacemaker.service (NOTE: it will wait sbd.service in the background for 36s)

INFO: Cluster services started on sle15sp3-1
INFO: Cluster services started on sle15sp3-2
INFO: Cluster services started on sle15sp3-3

real    0m2.912s
user    0m0.455s
sys     0m0.089s

crmsh/ui_cluster.py
```diff
@@ -29,13 +29,60 @@ def parse_options(parser, args):
...
+            add_help=False,
+            formatter_class=RawDescriptionHelpFormatter)
+    parser.add_argument("-h", "--help", action="store_true", dest="help", help="Show this help message")
+    parser.add_argument("--all", help="To {} on all nodes".format(action), action="store_true", dest="all")
+
+    options, args = parse_options(parser, args)
+    if options is None or args is None:
+        raise utils.TerminateSubCommand
+    if options.all and args:
+        context.fatal_error("Should either use --all or specific node")
```
zzhou1, line 71: s/node/node(s)/

Ideally, the same for several other places.

test/features/steps/step_implenment.py
```diff
@@ -4,6 +4,7 @@
 import datetime
```
zzhou1, line 4: Ah, this need another commit to change the file name typo, 

test/features/steps/step_implenment.py 

to: 
test/features/steps/step_implementation.py
 
though low priority.

crmsh/bootstrap.py
liangxin1300:
```
# time crm cluster start --all
INFO: Starting pacemaker(waiting for sbd 36s)...done
INFO: Cluster services started on 15sp2-1
INFO: Cluster services started on 15sp2-2
INFO: Cluster services started on 15sp2-3

real    0m33.171s
user    0m2.819s
sys     0m0.879s
```
The real time will be slightly less than expected wait time, it's true; since the wait time will be 1.2*2*SBD_WATCHDOG_TIMEOUT

But I never saw this real time: 2.9s ...

crmsh/ui_cluster.py
liangxin1300: Done

test/features/steps/step_implenment.py
liangxin1300: Done

crmsh/ui_cluster.py
liangxin1300:
```suggestion
```

zzhou1: yes, indeed

crmsh/bootstrap.py
zzhou1: Yeah, with the latest version. I see the same as you. Good!

crmsh/ui_cluster.py
liangxin1300: Done

zzhou1 (approved):

crmsh/ui_cluster.py
```diff
@@ -91,78 +138,94 @@ def __init__(self):
...
-            logger.info("Cluster services started")
-        except IOError as err:
-            context.fatal_error(str(err))
-
-        # TODO: optionally start services on all nodes or specific node
+        node_list = parse_option_for_nodes(context, *args)
+        for node in node_list[:]:
+            if utils.service_is_active("pacemaker.service", remote_addr=node):
+                logger.info("Cluster services already started on {}".format(node))
+                node_list.remove(node)
```
gao-yan, line 149: Pacemaker.service being active doesn't necessarily mean corosync-qdevice.service is active as well? Should corosync-qdevice be checked for the full set of nodes as well?

crmsh/ui_cluster.py
```diff
@@ -91,78 +138,94 @@ def __init__(self):
...
+        if not node_list:
+            return
+
+        bootstrap.wait_for_cluster("Waiting for {} online".format(' '.join(node_list)), node_list)
+
+        # When dlm configured and quorum is lost, before stop cluster service, should set
+        # enable_quorum_fencing=0, enable_quorum_lockspace=0 for dlm config option
+        if utils.is_dlm_configured(node_list[0]) and not utils.is_quorate(node_list[0]):
+            logger.debug("Quorum is lost; Set enable_quorum_fencing=0 and enable_quorum_lockspace=0 for dlm")
+            utils.set_dlm_option(peer=node_list[0], enable_quorum_fencing=0, enable_quorum_lockspace=0)
```
gao-yan, line 182: IIUC, both `enable_quorum_fencing` and `enable_quorum_lockspace` need to be disabled for an inquorate dlm to gracefully stop? Could there be any risks?

gao-yan, line 182: IIUC, both `enable_quorum_fencing` and `enable_quorum_lockspace` need to be disabled for an inquorate dlm to gracefully stop? Could there be any risks?

crmsh/ui_cluster.py
gao-yan, line 149: Pacemaker.service being active doesn't necessarily mean corosync-qdevice.service is active as well, right? Should corosync-qdevice be checked for the full set of nodes as well?

gao-yan (commented):
Thanks for the nice work, @liangxin1300!

crmsh/ui_node.py
```diff
@@ -287,36 +290,87 @@ def do_print(uname):
...
+                # Create standby nvpair
+                standby_template_str = standby_template.format(node_id=node_id, value="on")
+                xml_item.append(etree.fromstring(standby_template_str))
+
+        cib_str = xmlutil.xml_tostring(cib)
+        # Consider both "nodes" and "status" section might contain different standby value at the same time
+        # Should replace all possible "off" values here
+        for node in node_list:
+            node_id = utils.get_nodeid_from_name(node)
+            cib_str = re.sub(constants.STANDBY_NV_RE.format(node_id=node_id, value="off"), r'\1value="on"\2', cib_str)
```
gao-yan, line 340: It doesn't seem to cover the cases where it's set with other valid boolean values "false/no/n/0"?  How about substituting that with standby=on anyway?

And technically, the id of such an nvpair doesn't have to contain the node_id. Could we rely on the results of xpath queries like done in above and do the modifications?

crmsh/ui_node.py
```diff
@@ -287,36 +290,87 @@ def do_print(uname):
...
+                logger.info("Node %s already online", node)
+                node_list.remove(node)
+        if not node_list:
+            return
+
+        cib = xmlutil.cibdump2elem()
+        cib_str = xmlutil.xml_tostring(cib)
+        for node in node_list:
+            node_id = utils.get_nodeid_from_name(node)
+            cib_str = re.sub(constants.STANDBY_NV_RE.format(node_id=node_id, value="on"), r'\1value="off"\2', cib_str)
```
gao-yan, line 368: Similarly, a more reliable way would be do xpath queries and modify/delete

crmsh/ui_cluster.py
liangxin1300: When using bootstrap to setup cluster with qdevice, qdevice service will be enabled, so after reboot, corosync-qdevice will be started
@zzhou1 What do you think?

crmsh/ui_cluster.py
```diff
@@ -91,78 +138,94 @@ def __init__(self):
...
+        for node in node_list[:]:
+            if utils.service_is_active("pacemaker.service", remote_addr=node):
+                logger.info("Cluster services already started on {}".format(node))
+                node_list.remove(node)
+        if not node_list:
+            return
+
+        bootstrap.start_pacemaker(node_list)
+        if utils.is_qdevice_configured():
+            utils.start_service("corosync-qdevice", node_list=node_list)
```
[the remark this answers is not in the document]
liangxin1300: `bootstrap.start_pacemaker` here is to start pacemaker.service, according the dependency, corosync.service will be started firstly, then start qdevice does make sense; Otherwise, start qdevice without corosync started will be failed:)

crmsh/ui_cluster.py
```diff
@@ -91,78 +138,94 @@ def __init__(self):
...
+                if utils.service_is_active("sbd.service", remote_addr=node):
+                    utils.stop_service("corosync", remote_addr=node)
+                    logger.info("Cluster services stopped on {}".format(node))
+                else:
+                    logger.info("Cluster services already stopped on {}".format(node))
+                node_list.remove(node)
+        if not node_list:
+            return
+
+        bootstrap.wait_for_cluster("Waiting for {} online".format(' '.join(node_list)), node_list)
```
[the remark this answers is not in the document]
liangxin1300: Yes.

- Just after `crm cluster start --all`, the status of all nodes are `UNCLEAN`, on this time, if execute `crm cluster stop --all`, I found the stop process will be hang, and on some node, the output of `crm_mon` will show all nodes' status are `pending`
- If all nodes already `Online`, in `wait_for_cluster` function, I made a check firstly, and the line `Waiting for .... online` will be quiet

crmsh/ui_cluster.py
```diff
@@ -91,78 +138,94 @@ def __init__(self):
...
-        Restarts the cluster services on this node
+        Common part for enable and disable
         '''
-        self.do_stop(context)
-        self.do_start(context)
+        node_list = parse_option_for_nodes(context, *args)
+        action = context.get_command_name()
+        utils.cluster_run_cmd("systemctl {} pacemaker.service".format(action), node_list)
+        if utils.is_qdevice_configured():
+            utils.cluster_run_cmd("systemctl {} corosync-qdevice.service".format(action), node_list)
```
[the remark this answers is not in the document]
liangxin1300: corosync-qdevice.service will not be started if not configured, right? So I think to check if configured then do the action will no harm?

crmsh/ui_node.py
```diff
@@ -287,36 +290,87 @@ def do_print(uname):
...
-            opts = "--lifetime='forever'"
-        return utils.ext_cmd(self.node_standby % (node, "on", opts)) == 0
+            lifetime_opt = lifetime
+            args = args[:-1]
+
+        # Parse node option
+        node_list = parse_option_for_nodes(context, *args)
+        for node in node_list[:]:
+            if utils.is_standby(node):
+                logger.info("Node %s already standby", node)
```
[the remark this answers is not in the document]
liangxin1300: Both "reboot" and "forever" lifetime, the status from `crm_mon` will be `standby`,
And using `crm node online` will revert the status for both cases

crmsh/ui_node.py
liangxin1300: Good point! Thanks!

Consider need more time to verify these corner cases, can I put these into my backlog and improve later?

crmsh/ui_cluster.py
zhaohem: yes, both. 
from my opinion, there doesn't have any risk. because this action is only triggered by cmd "crm cluster stop".
before this patch, cluster (by dlm_controld) deny to do any further action until quorum is true.
after this patch, cluster will directly stop, there is no other behavior.

crmsh/ui_node.py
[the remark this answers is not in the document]
gao-yan: I mean, for example if the user intends to put a node into "forever standby" which is already in "reboot standby"...

crmsh/ui_node.py
gao-yan: Sure. We can improve that later.

crmsh/ui_cluster.py
gao-yan: The biggest concern is, given that this cluster partition is inquorate, there might a quorate partition standing... For example this partition has been split and inquorate, but somehow hasn't been fenced. If we are shutting down this cluster partition, and in case these settings make the inquorate partition be able to acquire the access to lockspace and corrupt data, that would be a disaster. We should never sacrifice data integrity even for graceful shutdown...

crmsh/ui_cluster.py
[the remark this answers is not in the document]
gao-yan: Given that corosync-qdevice.service has these defined as well:
```
Wants=corosync.service
After=corosync.service
```
I'd expect pure start of qdevice.service will resolve the dependency as well.

crmsh/ui_cluster.py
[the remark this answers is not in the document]
gao-yan: I guess it depends on how is_qdevice_configured() does the check. Of course one could remove qdevice configuration from corosync.conf before stopping the running corosync-qdevice.service...

crmsh/ui_node.py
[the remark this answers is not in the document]
liangxin1300: > I mean, for example if an user intends to put a node into "forever standby" which is already in "reboot standby"...

Then here this action will be rejected and say "already standby", until online it firstly.
Do we need to support this change? I mean change the lifetime from "reboot" to "forever"?
Current code in production support that, bug in my view, that might cause confuse/conflicts, like:

1. crm node standby <node> reboot
2. crm node standby <node>
3. crm node online <node>
Then from crm_mon, we will see this node already in `standby` status

To make these actions more clear, I think we should reject above action and say "already standby"

crmsh/ui_cluster.py
ganghe: enable_quorum_lockspace is disabled, this will make dlm lockspace related operation can keep going when the cluster quorum is lost.

zhaohem: your comment makes sense to me.

zhaohem: your comment makes sense to me.
the better solution should add a prompt info for user, let user to make decision: whether crm modify enable_quorum_(fencing|lockspace) to graceful shutdown.

zhaohem: the better solution should add a prompt info for user, let user to make decision: whether crm modify enable_quorum_(fencing|lockspace) to graceful shutdown.

crmsh/ui_cluster.py
zzhou1: "node_list.remove(node)" can be wrong in the case pacemaker.service is up, but qdevice is not, in the use case of normal stop/start. No necessary in the case of reboot.

crmsh/ui_node.py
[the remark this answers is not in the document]
gao-yan: I wouldn't check if it's already in any kind of standby :-). I'd just set what the user wants, as long as "online" command could bring it out of any cases/combinations of standby.

crmsh/ui_cluster.py
gao-yan: > the better solution should add a prompt info for user, let user to make decision: whether crm modify enable_quorum_(fencing|lockspace) to graceful shutdown.

It's good idea. We should give clear heads-up/warning so that they know what they are doing.

crmsh/ui_cluster.py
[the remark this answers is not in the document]
zzhou1: I can see a little challenge for the current code to check and start pacemaker and qdevice altogether before report the correct information. This may need more lines of code than expect.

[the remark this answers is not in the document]
zzhou1: I can get the point to move corosync-qdevice.service start operation before pacemaker. There is no harm to have multiple systemctl start, in theory.

crmsh/ui_cluster.py
zzhou1: > The biggest concern is, given that this cluster partition is inquorate, there might a quorate partition standing... For example this partition has been split and inquorate, but somehow hasn't been fenced. If we are shutting down this cluster partition, and in case these settings make the inquorate partition be able to acquire the access to lockspace and corrupt data, that would be a disaster. We should never sacrifice data integrity even for graceful shutdown...

No risk here. These changes, "enable_quorum_fencing=0 and enable_quorum_lockspace=0", are intentionally put as the runtime by "crm cluster stop". It will revert back to the static default config after restart the dlm_controld daemon.

With that, it looks to me an INFO is enough to tell those runtime changes happens, and probably not as WARNING even.

gao-yan: We don't know whether one *really* is aware of the real situation about whether there's no other cluster partitions standing...

crmsh/ui_cluster.py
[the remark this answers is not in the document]
zzhou1: Good point. 

I can understand the wait experience in bootstrap. Well, probably no necessary to force user to wait `crm cluster start`.

I would rather think to remove wait in `crm cluster start`, and let `crm cluster stop` directly abort if the criteria not meet. It's a reasonable error handling, rather than have sysadmin really wait in front of the screen. BTW, not sure if you want to fix it in this PR, or track this as an issue and fix it later on.

For those scripts do want to 'wait', before we implement '--wait' option, the following example steps could help
crm cluster start
crm wait_for_startup
crm cluster stop

crmsh/ui_cluster.py
gao-yan: Thinking about it on the more safe/paranoid side, even if the user has confirmed to proceed, but allowing this simultaneously on multiple nodes is more like opening a Pandora's box... Since right after that,  during stop, this cluster partition might spit apart into more partitions...

If we go for it, we could ask user once, but we'd better proceed this specific procedure the serialized way: so
`set_dlm_option -> stop dlm/pacemaker` for only one node at a time, and proceed for another node only after it has succeeded on this one.

crmsh/ui_cluster.py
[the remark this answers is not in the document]
liangxin1300: Done

crmsh/ui_cluster.py
zzhou1: My previous point stays with the last standing single node situation. Given that the node is inquorate already, the proposed behavior for `crm cluster stop` (aka. crmsh -> set_dlm_option -> stop dlm/patcemaker) is just a "fencing" operation to protect data integrity, though no necessarily STONITH at the node level to do reboot. 

Well, for the situation with multiple inquorate nodes, aka. multiple partitions, then the code here do have problem for '--all' situation. Simply because set_dlm_option only applies to one node. Not sure, if it is simple enough to address it in this PR, or open an issue to clarify this in another PR.

crmsh/ui_cluster.py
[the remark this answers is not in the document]
zzhou1: Good point, I think I would naturally expect `crm cluster stop` to stop corosync-qdevice.service even if corosync.conf has no qdevice.

crmsh/ui_cluster.py
zhaohem: set_dlm_option is implemented by "dlm_tool set_config", which only run on a single node each time.

zzhou1: > set_dlm_option is implemented by "dlm_tool set_config", which only run on a single node each time.

Fine. 

Given the above situation, when all nodes are inquorate, what's the suggested graceful shutdown steps internally for "stop --all" ? My reading of the code is it only change the current local node, no repeat the same on other nodes. 

The situation is more fun, in theory for a big cluster, some nodes are quorate, some are not. What's the suggested internal steps for "stop --all"? @zhaohem

zzhou1: Maybe crmsh should never operate a cluster in the transient state at all by default? And ask user to answer Y/n, or user --force ?

crmsh/ui_node.py
liangxin1300: PR to improve this: https://github.com/ClusterLabs/crmsh/pull/893

crmsh/ui_cluster.py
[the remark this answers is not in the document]
liangxin1300: PR to improve this: #895

crmsh/ui_cluster.py
liangxin1300: Improve this in #898

PR #807, opened 2021-04-26
Add warnings for damage options in preflight_check

liangxin1300:
#### Changes
1. replace --yes as --force
```
-f, --force             Force to skip all prompts (Use with caution, this is destructive)
```
2. Add warnings for damage options
```
# crm analyze preflight_check --kill-corosync


==============================================
Testcase:          Force kill corosync
Looping Kill:      False
Expected State:    a) corosync process restarted
                   b) Or, this node fenced.

!!! WARNING WARNING WARNING !!!
THIS CASE MAY LEAD TO NODE BE FENCED.
TYPE Yes TO CONTINUE, OTHER INPUTS WILL CANCEL THIS CASE:
```

Screenshot from 2021-04-26 14-55-29

### DIFF
--- a/crmsh/preflight_check/task.py
+++ b/crmsh/preflight_check/task.py
@@ -25,13 +25,17 @@ class Task(object):
     Task is a base class
     Use for record the information of each test case
     """
+    REBOOT_WARNING = """!!! WARNING WARNING WARNING !!!
+THIS CASE MAY LEAD TO NODE BE FENCED.
+TYPE Yes TO CONTINUE, OTHER INPUTS WILL CANCEL THIS CASE: """
+
     def __init__(self, description, flush=False, quiet=False):
         """
         Init function
         flush, to print the message immediately
         """
         self.passed = True
-        self.yes = False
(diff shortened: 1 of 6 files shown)


crmsh/preflight_check/task.py
```diff
@@ -25,13 +25,17 @@ class Task(object):
     Task is a base class
     Use for record the information of each test case
     """
+    REBOOT_WARNING = """!!! WARNING WARNING WARNING !!!
+THIS CASE MAY LEAD TO NODE BE FENCED.
+TYPE Yes TO CONTINUE, OTHER INPUTS WILL CANCEL THIS CASE: """
```
zzhou1, line 30: Just to maintain the simliar habit to most of other CLI tools. How about:
s%CASE: %CASE [Yes/No] (No): %

zzhou1 (changes requested):
Need add the major change about the command layout
s/crm analyze preflight_check/crm cluster crash_test/

PR #809, opened 2021-05-06
Fix: bootstrap: change StrictHostKeyChecking=no as a constant

liangxin1300:


test/unittests/test_bootstrap.py
```diff
@@ -998,7 +998,7 @@ def test_append_to_remote_file(self, mock_invoke, mock_error):
         mock_invoke.return_value = (False, None, "error")
         error_string = 'Failed to append contents of fromfile to node1:\n"error"\n\n    crmsh has no way to help you to setup up passwordless ssh among nodes at this time. \n    As the hint, likely, `PasswordAuthentication` is \'no\' in /etc/ssh/sshd_config. \n    Given in this case, users must setup passwordless ssh beforehand, or change it to \'yes\' and manage passwords properly\n    '
         bootstrap.append_to_remote_file("fromfile", "node1", "tofile")
-        cmd = "cat fromfile | ssh -oStrictHostKeyChecking=no root@node1 'cat >> tofile'"
+        cmd = "cat fromfile | ssh -o StrictHostKeyChecking=no root@node1 'cat >> tofile'"
```
nick-wang, line 1001: Is that possible to import SSH_OPTION in test as well? or at least defined and reuse this in the beginning?

nick-wang (approved):
Besides one minor thing for the test, looks good to me.

test/unittests/test_bootstrap.py
liangxin1300: Changed

PR #815, opened 2021-05-08
Split class Watchdog into watchdog.py from bootstrap.py

liangxin1300:
bootstrap.py is huge, should split some classes into specific .py file, also make easier for maintain and testing


nick-wang (approved):
Approved, since it just move the same code to `crmsh/watchdog.py`

PR #818, opened 2021-05-21
Fix: bootstrap: update sbd watchdog timeout when using diskless SBD with qdevice

liangxin1300:
When using diskless SBD with qdevice, **SBD_WATCHDOG_TIMEOUT** in /etc/sysconfig/sbd should larger than qdevice sync_timeout, see https://github.com/corosync/corosync-qdevice/issues/21

#### Changed logic include:
- **When set diskless SBD and qdevice together**, like `crm cluster init -S --qnetd-hostname <qnetd_addr>`, set **SBD_WATCHDOG_TIMEOUT**=35, set **stonith-timeou**t as (SBD_WATCHDOG_TIMEOUT*2+20)s，stonith-watchdog-timeout=-1

- **When adding qdevice on a running cluster also with diskless SBD running**, like `crm cluster init qdevice --qnetd-hostname <qnetd_addr>`, set **SBD_WATCHDOG_TIMEOUT**=35, and sync the sbd config file, **stonith-timeou**t=90s (stonith-watchdog-timeout already been set as -1), then restart the cluster service on needed (auto restart when no RA running; hints manually restart when RA running)

- **When adding diskless SBD on a running cluster also with qdevice runing**, like `crm cluster init sbd -S -y`, get the value of qdevice sync_timeout(default is 30s), then plus with 5s as the value of **SBD_WATCHDOG_TIMEOUT**， set **stonith-timeout** as (SBD_WATCHDOG_TIMEOUT*2+20)s, and set stonith-watchdog-timeout=-1

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -205,6 +205,10 @@ Configure SBD:
     DISKLESS_SBD_WARNING = """Diskless SBD requires cluster with three or more nodes.
 If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."""
     PARSE_RE = "[; ]"
+    DISKLESS_CRM_CMD = "crm configure property stonith-enabled=true stonith-watchdog-timeout={} stonith-timeout={}"
+    SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE = 35
+    STONITH_WATCHDOG_TIMEOUT_DEFAULT = "10s"
+    STONITH_WATCHDOG_TIMEOUT_DEFAULT_S390 = "30s"
 
     def __init__(self, sbd_devices=None, diskless_sbd=False):
         """
@@ -217,7 +221,10 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         self.diskless_sbd = diskless_sbd
         self._sbd_devices = None
         self._watchdog_inst = None
-        self._stonith_watchdog_timeout = "10s"
+        self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
+        self._stonith_timeout = 60
+        self._sbd_watchdog_timeout = 0
+        self._is_s390 = "390" in os.uname().machine
 
     @staticmethod
     def _get_device_uuid(dev, node=None):
@@ -331,34 +338,53 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         Update /etc/sysconfig/sbd
         """
         shutil.copyfile(self.SYSCONFIG_SBD_TEMPLATE, SYSCONFIG_SBD)
+        self._determine_sbd_watchdog_timeout()
         sbd_config_dict = {
                 "SBD_PACEMAKER": "yes",
                 "SBD_STARTMODE": "always",
                 "SBD_DELAY_START": "no",
                 "SBD_WATCHDOG_DEV": self._watchdog_inst.watchdog_device_name
                 }
+        if self._sbd_watchdog_timeout > 0:
+            sbd_config_dict["SBD_WATCHDOG_TIMEOUT"] = str(self._sbd_watchdog_timeout)
         if self._sbd_devices:
             sbd_config_dict["SBD_DEVICE"] = ';'.join(self._sbd_devices)
         utils.sysconfig_set(SYSCONFIG_SBD, **sbd_config_dict)
         csync2_update(SYSCONFIG_SBD)
 
+    def _determine_sbd_watchdog_timeout(self):
+        """
+        When using diskless SBD, determine value of SBD_WATCHDOG_TIMEOUT
+        """
+        if not self.diskless_sbd:
+            return
+        # add sbd after qdevice started
+        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
+            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
+            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
+            if self._is_s390 and self._sbd_watchdog_timeout < 15:
+                self._sbd_watchdog_timeout = 15
+            self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
+        # add sbd and qdevice together from beginning
+        elif _context.qdevice_inst:
+            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+            self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
+
     def _determine_stonith_watchdog_timeout(self):
         """
         Determine value of stonith-watchdog-timeout
         """
-        conf = utils.parse_sysconfig(SYSCONFIG_SBD)
-        res = conf.get("SBD_WATCHDOG_TIMEOUT")
+        res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
         if res:
             self._stonith_watchdog_timeout = -1
-        elif "390" in os.uname().machine:
-            self._stonith_watchdog_timeout = "30s"
+        elif self._is_s390:
+            self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT_S390
 
     def _get_sbd_device_from_config(self):
         """
         Gets currently configured SBD device, i.e. what's in /etc/sysconfig/sbd
         """
-        conf = utils.parse_sysconfig(SYSCONFIG_SBD)
-        res = conf.get("SBD_DEVICE")
+        res = SBDManager.get_sbd_value_from_config("SBD_DEVICE")
         if res:
             return utils.re_split_string(self.PARSE_RE, res)
         else:
@@ -376,7 +402,8 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         else:
             warn("To start sbd.service, need to restart cluster service manually on each node")
             if self.diskless_sbd:
-                warn("Then run \"crm configure property stonith-enabled=true stonith-watchdog-timeout={}\" on any node".format(self._stonith_watchdog_timeout))
+                cmd = self.DISKLESS_CRM_CMD.format(self._stonith_watchdog_timeout, str(self._stonith_timeout)+"s")
+                warn("Then run \"{}\" on any node".format(cmd))
             else:
                 self.configure_sbd_resource()
 
@@ -445,7 +472,8 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
             if not invokerc("crm configure property stonith-enabled=true"):
                 error("Can't enable STONITH for SBD")
         else:
-            if not invokerc("crm configure property stonith-enabled=true stonith-watchdog-timeout={}".format(self._stonith_watchdog_timeout)):
+            cmd = self.DISKLESS_CRM_CMD.format(self._stonith_watchdog_timeout, str(self._stonith_timeout)+"s")
+            if not invokerc(cmd):
                 error("Can't enable STONITH for diskless SBD")
 
     def join_sbd(self, peer_host):
@@ -490,6 +518,41 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         """
         inst = cls()
         return inst._get_sbd_device_from_config()
+
+    @classmethod
+    def is_using_diskless_sbd(cls):
+        """
+        Check if using diskless SBD
+        """
+        inst = cls()
+        dev_list = inst._get_sbd_device_from_config()
+        if not dev_list and utils.service_is_active("sbd.service"):
+            return True
+        return False
+
+    @staticmethod
+    def update_configuration(sbd_config_dict):
+        """
+        Update and sync sbd configuration
+        """
+        utils.sysconfig_set(SYSCONFIG_SBD, **sbd_config_dict)
+        csync2_update(SYSCONFIG_SBD)
+
+    @staticmethod
+    def calculate_stonith_timeout(sbd_watchdog_timeout):
+        """
+        Calculate stonith timeout
+        """
+        return int(sbd_watchdog_timeout * 2 * 1.2)
+
+    @staticmethod
+    def get_sbd_value_from_config(key):
+        """
+        Get value from /etc/sysconfig/sbd
+        """
+        conf = utils.parse_sysconfig(SYSCONFIG_SBD)
+        res = conf.get(key)
+        return res
 
 
 _context = None
@@ -1562,7 +1625,7 @@ Configure Administration IP Address:
     wait_for_resource("Configuring virtual IP ({})".format(adminaddr), "admin-ip")
 
 
-def evaluate_qdevice_quorum_effect(mode):
+def evaluate_qdevice_quorum_effect(mode, diskless_sbd=False):
     """
     While adding/removing qdevice, get current expected votes and actual total votes,
     to calculate after adding/removing qdevice, whether cluster has quorum
@@ -1576,7 +1639,7 @@ def evaluate_qdevice_quorum_effect(mode):
     elif mode == QDEVICE_REMOVE:
         actual_votes -= 1
 
-    if utils.is_quorate(expected_votes, actual_votes):
+    if utils.is_quorate(expected_votes, actual_votes) and not diskless_sbd:
         # safe to use reload
         return QdevicePolicy.QDEVICE_RELOAD
     elif utils.has_resource_running():
@@ -1631,7 +1694,18 @@ def init_qdevice():
         return
     if _context.stage == "qdevice":
         utils.check_all_nodes_reachable()
-        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD)
+        using_diskless_sbd = SBDManager.is_using_diskless_sbd()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD, using_diskless_sbd)
+        # add qdevice after diskless sbd started
+        if using_diskless_sbd:
+            res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
+            if res:
+                sbd_watchdog_timeout = max(int(res), SBDManager.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE)
+            else:
+                sbd_watchdog_timeout = SBDManager.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+            stonith_timeout = SBDManager.calculate_stonith_timeout(sbd_watchdog_timeout)
+            SBDManager.update_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout)})
+            invokerc("crm configure property stonith-watchdog-timeout=-1 stonith-timeout={}s".format(stonith_timeout))
 
     status("""
 Configure Qdevice/Qnetd:""")
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2868,4 +2868,15 @@ def append_res_to_group(group_id, res_id):
     """
     cmd = "crm configure modgroup {} add {}".format(group_id, res_id)
     get_stdout_or_raise_error(cmd)
+
+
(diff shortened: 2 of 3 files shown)


crmsh/bootstrap.py
```diff
@@ -338,17 +340,34 @@ def _update_configuration(self):
...
+    def _determine_sbd_watchdog_timeout(self):
+        """
+        When using diskless SBD, Determine value of SBD_WATCHDOG_TIMEOUT
+        """
+        if not self.diskless_sbd:
+            return
+        # add sbd after qdevice started
+        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
+            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
+            self._sbd_watchdog_timeout += qdevice_sync_timeout
```
liangxin1300, line 365: When adding sbd service via sbd stage after cluster and qdevice running

crmsh/bootstrap.py
```diff
@@ -1800,7 +1844,11 @@ def init_qdevice():
         return
     if _context.stage == "qdevice":
         utils.check_all_nodes_reachable()
-        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD)
+        using_diskless_sbd = SBDManager.is_using_diskless_sbd()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD, using_diskless_sbd)
+        if using_diskless_sbd:
+            SBDManager.update_configuration({"SBD_WATCHDOG_TIMEOUT": "35"})
+            invokerc("crm configure property stonith-timeout=90s")
```
liangxin1300, line 1851: When adding qdevice stage after cluster and diskless SBD was runing

crmsh/bootstrap.py
```diff
@@ -338,17 +341,35 @@ def _update_configuration(self):
...
+            return
+        # add sbd after qdevice started
+        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
+            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
+            self._sbd_watchdog_timeout += qdevice_sync_timeout
+            self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
+        # add sbd and qdevice together
+        elif _context.qdevice_inst:
+            self._sbd_watchdog_timeout = 35
+            self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
```
liangxin1300, line 371: When adding diskless sbd and qdevice service together

crmsh/bootstrap.py
```diff
@@ -1800,7 +1849,13 @@ def init_qdevice():
         return
     if _context.stage == "qdevice":
         utils.check_all_nodes_reachable()
-        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD)
+        using_diskless_sbd = SBDManager.is_using_diskless_sbd()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD, using_diskless_sbd)
+        if using_diskless_sbd:
```
liangxin1300, line 1854: When adding qdevice stage on a running cluster which with diskless sbd running

liangxin1300:
@gao-yan Please review this when you have time. Thanks!

BTW, should we change this logic for s390?
```Python
    def _determine_stonith_watchdog_timeout(self):
        """
        Determine value of stonith-watchdog-timeout
        """
        conf = utils.parse_sysconfig(SYSCONFIG_SBD)
        res = conf.get("SBD_WATCHDOG_TIMEOUT")
        if res:
            self._stonith_watchdog_timeout = -1
        elif "390" in os.uname().machine:
            self._stonith_watchdog_timeout = "30s"
```

gao-yan (commented):
Thanks for the nice work, @liangxin1300!

crmsh/bootstrap.py
```diff
@@ -331,17 +334,36 @@ def _update_configuration(self):
...
+    def _determine_sbd_watchdog_timeout(self):
+        """
+        When using diskless SBD, determine value of SBD_WATCHDOG_TIMEOUT
+        """
+        if not self.diskless_sbd:
+            return
+        # add sbd after qdevice started
+        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
+            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
+            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
```
gao-yan, line 360: I think we need to consider s390(x) in here. I need to make sure  _sbd_watchdog_timeout >= 15s in case of s390(x).

gao-yan:
> @gao-yan Please review this when you have time. Thanks!
> 
> BTW, should we change this logic for s390?
> 
> ```python
>     def _determine_stonith_watchdog_timeout(self):
>         """
>         Determine value of stonith-watchdog-timeout
>         """
>         conf = utils.parse_sysconfig(SYSCONFIG_SBD)
>         res = conf.get("SBD_WATCHDOG_TIMEOUT")
>         if res:
>             self._stonith_watchdog_timeout = -1
>         elif "390" in os.uname().machine:
>             self._stonith_watchdog_timeout = "30s"
> ```

Then as long as this function is always called after sbd sysconfig is settled by _update_configuration() no matter if diskless sbd is used, it should be fine.

crmsh/bootstrap.py
```diff
@@ -490,6 +514,32 @@ def verify_sbd_device(cls):
...
+        """
+        utils.sysconfig_set(SYSCONFIG_SBD, **sbd_config_dict)
+        csync2_update(SYSCONFIG_SBD)
+
+    @staticmethod
+    def calculate_stonith_timeout(sbd_watchdog_timeout):
+        """
+        Calculate stonith timeout
+        """
+        return (sbd_watchdog_timeout * 2 + 20)
```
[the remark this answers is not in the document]
liangxin1300: Changed as:
```
int(sbd_watchdog_timeout * 2 * 1.2)
```

crmsh/bootstrap.py
```diff
@@ -1631,7 +1681,13 @@ def init_qdevice():
         return
     if _context.stage == "qdevice":
         utils.check_all_nodes_reachable()
-        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD)
+        using_diskless_sbd = SBDManager.is_using_diskless_sbd()
+        _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD, using_diskless_sbd)
+        if using_diskless_sbd:
+            sbd_watchdog_timeout = 35
```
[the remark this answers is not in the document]
liangxin1300: Changed as `SBDManager.SBD_WATCHDOG_TIMEOUT_DEFAULT = 35`

crmsh/bootstrap.py
```diff
@@ -1631,7 +1681,13 @@ def init_qdevice():
...
+            stonith_timeout = SBDManager.calculate_stonith_timeout(sbd_watchdog_timeout)
+            SBDManager.update_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout)})
+            invokerc("crm configure property stonith-timeout={}s".format(stonith_timeout))
```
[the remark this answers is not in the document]
liangxin1300: This piece of code will be used when configured diskless SBD then adding qdevice via qdevice stage
So at this time, stonith-watchdog-timeout already be set as "-1"

crmsh/bootstrap.py
liangxin1300: @gao-yan Here `qdevice_sync_timeout` by default is 30s, then `_sbd_watchdog_timeout` is 35, already >= 15s
Do you mean in s390 case, if value of `qdevice_sync_timeout` is small, like 5, we should make sure `sbd_wachdog_timeout` >= 15? Changed like this?
```Python
            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
            if self._is_s390 and self._sbd_watchdog_timeout < 15:
                self._sbd_watchdog_timeout = 15
```

gao-yan: Yes. IIUC, sbd stage can run for an existing cluster? That's also the reason why rather than relying on the default value, we are getting the real qdevice sync_timeout with get_qdevice_sync_timeout() just in case it was configured differently or changed somehow ...

crmsh/bootstrap.py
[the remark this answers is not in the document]
gao-yan: Similarly, qdevice stage can run for an existing cluster? We are not sure if stonith-watchdog-timeout was configured differently or changed somehow?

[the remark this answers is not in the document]
liangxin1300: OK, make sense. It's not wrong to explicitly set it again

crmsh/bootstrap.py
liangxin1300: OK, changed

liangxin1300:
@gao-yan Let me make a summary:

-  When set diskless SBD and qdevice together
```
        stonith-watchdog-timeout=-1 \
        stonith-timeout=84s
```
```
SBD_WATCHDOG_TIMEOUT=35
```
- When adding qdevice on a running cluster also with diskless SBD running
```
        stonith-watchdog-timeout=-1 \
        stonith-timeout=84s
```
```
SBD_WATCHDOG_TIMEOUT=35
```
- When adding diskless SBD on a running cluster also with qdevice runing
```
        stonith-watchdog-timeout=-1 \
        stonith-timeout=84s
```
```
SBD_WATCHDOG_TIMEOUT=35
```

gao-yan:
> * When adding qdevice on a running cluster also with diskless SBD running
> 
> ```
>         stonith-watchdog-timeout=-1 \
>         stonith-timeout=84s
> ```
> 
> ```
> SBD_WATCHDOG_TIMEOUT=35
> ```

Hrm, should we check whether there's already any existing SBD_WATCHDOG_TIMEOUT? And if it's already >= 35s, then we don't touch it?

> 
> * When adding diskless SBD on a running cluster also with qdevice runing
> 
> ```
>         stonith-watchdog-timeout=-1 \
>         stonith-timeout=84s
> ```
> 
> ```
> SBD_WATCHDOG_TIMEOUT=35
> ```

Given the current logic, to be precise, it should be:
```
SBD_WATCHDOG_TIMEOUT= qdevice-sync_timeout + 5  -- (not less than 15s for s390(x) any ways )
stonith-timeout = SBD_WATCHDOG_TIMEOUT * 2 * 1.2
```
, right?

> 
> * Normal case for diskless SBD without qdeivce
> 
> ```
>         stonith-watchdog-timeout=-1 \
>         stonith-timeout=60s
> ```
> 
> ```
> SBD_WATCHDOG_TIMEOUT=5
> ```

IIUC according to the logic, `SBD_WATCHDOG_TIMEOUT` is explicitly set only when qdevice is used?

liangxin1300:
> > * When adding qdevice on a running cluster also with diskless SBD running
> > 
> > ```
> >         stonith-watchdog-timeout=-1 \
> >         stonith-timeout=84s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > SBD_WATCHDOG_TIMEOUT=35
> > ```
> 
> Hrm, should we check whether there's already any existing SBD_WATCHDOG_TIMEOUT? And if it's already >= 35s, then we don't touch it?

I don't think we should do this check, just get the suitable timeout value against current user case provided by current options, and just overwrite the `/etc/sysconfig/sbd` on each time, should no harm or side effect

> 
> > * When adding diskless SBD on a running cluster also with qdevice runing
> > 
> > ```
> >         stonith-watchdog-timeout=-1 \
> >         stonith-timeout=84s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > SBD_WATCHDOG_TIMEOUT=35
> > ```
> 
> Given the current logic, to be precise, it should be:
> 
> ```
> SBD_WATCHDOG_TIMEOUT= qdevice-sync_timeout + 5  -- (not less than 15s for s390(x) any ways )
> stonith-timeout = SBD_WATCHDOG_TIMEOUT * 2 * 1.2
> ```
> 
> , right?

Yes, the logic is
```Python
        # add sbd after qdevice started
        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
            if self._is_s390 and self._sbd_watchdog_timeout < 15:
                self._sbd_watchdog_timeout = 15
            self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
```
> 
> > * Normal case for diskless SBD without qdeivce
> > 
> > ```
> >         stonith-watchdog-timeout=-1 \
> >         stonith-timeout=60s
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > SBD_WATCHDOG_TIMEOUT=5
> > ```
> 
> IIUC according to the logic, `SBD_WATCHDOG_TIMEOUT` is explicitly set only if qdevice is used?

Only if:
- qdevice and diskless SBD are using together
- qdevice configured firstly, then diskless SBD
- diskless SBD configured firstly, then qdevice

else, the `SBD_WATCHDOG_TIMEOUT` will keep unchanged as `5`

liangxin1300:
@gao-yan 
Maybe it will be more direct to see the results on 15sp3?
RPM link with this patch is here: https://build.opensuse.org/package/show/home:XinLiang:branches:crmsh_qdevice_sbd/crmsh

liangxin1300:
> Users set a long SBD_WATCHDOG_TIMEOUT for reasons, for example more tolerance... The thing is this is basically about running `qdevice` stage alone for an existing running cluster. If the current sbd setting doesn't fit the need of running qdevice, it's sensible to change that. But if sbd setting is already good enough, I'd be surprised if `qdevice` stage touched my sbd setting shrinking SBD_WATCHDOG_TIMEOUT.

Followed with this logic, if SBD_WATCHDOG_TIMEOUT is already long enough, we should check whether stonith-timeout is >= SBD_WATCHDOG_TIMEOUT * 2 * 1.2, if yes, keep that value unchanged, too; otherwise, change the stonith-timeout as  `SBD_WATCHDOG_TIMEOUT * 2 * 1.2`, right?

gao-yan:
> > > Users set a long SBD_WATCHDOG_TIMEOUT for reasons, for example more tolerance... The thing is this is basically about running `qdevice` stage alone for an existing running cluster. If the current sbd setting doesn't fit the need of running qdevice, it's sensible to change that. But if sbd setting is already good enough, I'd be surprised if `qdevice` stage touched my sbd setting shrinking SBD_WATCHDOG_TIMEOUT.
> > 
> > 
> > Followed with this logic, if SBD_WATCHDOG_TIMEOUT is already long enough, we should check whether stonith-timeout is >= SBD_WATCHDOG_TIMEOUT * 2 * 1.2, if yes, keep that value unchanged, too; otherwise, change the stonith-timeout as `SBD_WATCHDOG_TIMEOUT * 2 * 1.2`, right?
> 
> Yes, it makes sense to still check that.

And also make sure stonith-watchdog-timeout is -1 or >= SBD_WATCHDOG_TIMEOUT * 2

liangxin1300:
@gao-yan After changing the logic, I should make a summary this time:
- **qdevice and diskless SBD are using together**, that means setup cluster from beginning
  - SBD_WATCHDOG_TIMEOUT has no original value, so set it as `35`
  - stonith-timeout has no original value, either, so set it as `int(sbd_watchdog_timeout * 2 * 1.2)`
  - stonith-watchdog-timeout has no original value, either, so set it as `-1`
 
- **qdevice configured firstly, then setup diskless SBD** via sbd stage
  - SBD_WATCHDOG_TIMEOUT has no original value, set it as `qdevice_sync_timeout + 5`
  - stonith-timeout might has original value, should get the value of property "stonith-timeout", if that property exists and larger than `int(sbd_watchdog_timeout * 2 * 1.2)`, use that larger value
  - stonith-watchdog-timeout has no original value, so set it as `-1`

- **diskless SBD configured firstly, then setup qdevice** via qdevice stage
  - SBD_WATCHDOG_TIMEOUT does have original value, get that value from `/etc/sysconfig/sbd`, compare with the default one, use the larger value
  - stonith-timeout does have original value, should get the value of property "stonith-timeout", if that property exists and larger than `int(sbd_watchdog_timeout * 2 * 1.2)`, use that larger value
  - stonith-watchdog-timeout does have original value, should get the value of property "stonith-watchdog-timeout", if that property exists and larger than `int(sbd_watchdog_timeout) * 2`, use that larger value, else, use -1

crmsh/bootstrap.py
```diff
@@ -331,34 +337,65 @@ def _update_configuration(self):
...
-        if res:
-            self._stonith_watchdog_timeout = -1
-        elif "390" in os.uname().machine:
-            self._stonith_watchdog_timeout = "30s"
+        if sbd_watchdog_timeout:
+            if check_exist:
+                exist_swt = cib_factory.get_property("stonith-watchdog-timeout")
+                if not exist_swt:
+                    return -1
+                stonith_watchdog_timeout = int(exist_swt.strip('s'))
```
gao-yan, line 382: Hrm, it's occurring to me that it actually is a little tricky here. Technically, users could configure such a timing setting with their favorite units, for example s, m, h, ms, us ...

It's kind of difficult to parse the value for comparison. Let's probably make it easy to just override any with `stonith-watchdog-timeout=-1`

crmsh/bootstrap.py
```diff
@@ -491,6 +532,46 @@ def get_sbd_device_from_config(cls):
...
+    @staticmethod
+    def calculate_stonith_timeout(sbd_watchdog_timeout, check_exist=False):
+        """
+        Calculate stonith timeout
+        """
+        stonith_timeout = int(sbd_watchdog_timeout * 2 * 1.2)
+        if check_exist:
+            exist_st = cib_factory.get_property("stonith-timeout")
+            if exist_st and int(exist_st.strip('s')) >= stonith_timeout:
+                stonith_timeout = int(exist_st.strip('s'))
```
gao-yan, line 563: Similarly, users could configure this with their favorite units too. It's kind of difficult to parse the value for comparison. Let's make it easy to just use `sbd_watchdog_timeout * 2 * 1.2`.

crmsh/bootstrap.py
```diff
@@ -205,6 +205,9 @@ class SBDManager(object):
     DISKLESS_SBD_WARNING = """Diskless SBD requires cluster with three or more nodes.
 If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."""
     PARSE_RE = "[; ]"
+    DISKLESS_CRM_CMD = "crm configure property stonith-enabled=true stonith-watchdog-timeout={} stonith-timeout={}"
+    SBD_WATCHDOG_TIMEOUT_DEFAULT = 35
```
gao-yan, line 209: It'd be more clear to use a name like `SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE` :-)

crmsh/bootstrap.py
```diff
@@ -331,34 +337,65 @@ def _update_configuration(self):
...
+                    return -1
+                stonith_watchdog_timeout = int(exist_swt.strip('s'))
+                if stonith_watchdog_timeout >= (int(sbd_watchdog_timeout) * 2):
+                    return "{}s".format(stonith_watchdog_timeout)
+                else:
+                    return -1
+            else:
+                return -1
+        elif is_s390:
+            return "30s"
```
gao-yan, line 390: AFAICT, it would be more straightforward to move this into the initial definition of `STONITH_WATCHDOG_TIMEOUT_DEFAULT`

gao-yan (commented):
Thanks for the changes.

crmsh/bootstrap.py
liangxin1300: OK
Changed

crmsh/bootstrap.py
liangxin1300: OK
Changed

crmsh/bootstrap.py

crmsh/bootstrap.py

gao-yan (approved):
Awesome. Thanks for the nice work, @liangxin1300!

liangxin1300:
> Awesome. Thanks for the nice work, @liangxin1300!

Thank you for your review and great suggestions!

PR #824, opened 2021-06-10
Dev: utils: allow configure link-local ipv6 address

liangxin1300:
Since corosync already support link-local address, we should remove this limitation


zzhou1 (approved):
LGTM

PR #831, opened 2021-06-21
Dev: sbd: enable SBD_DELAY_START in virtualization environment

liangxin1300:


zzhou1 (approved):

liangxin1300:
@gao-yan Any feedback about this PR?
Thanks!

crmsh/sbd.py
```diff
@@ -164,7 +164,7 @@ def _update_configuration(self):
         sbd_config_dict = {
                 "SBD_PACEMAKER": "yes",
                 "SBD_STARTMODE": "always",
-                "SBD_DELAY_START": "no",
+                "SBD_DELAY_START": "yes" if utils.detect_virt() else "no",
```
gao-yan, line 167: So far technically, a setting `SBD_DELAY_START=yes` takes effect only for sbd with disk,  since the delay value is basically `msgwait` retrieved from sbd metadata.

For diskless sbd, we'd have to explicitly specify `SBD_DELAY_START` with a delay value.

As also mentioned in the help text:
https://github.com/ClusterLabs/sbd/blob/master/src/sbd.sysconfig.in#L26-L48

, for sbd with disk, a safe value could be at least:
`corosync token timeout + consensus timeout + pcmk_delay_max + msgwait`

Similarly considering the possible factors, for disk-less sbd, a safe value could be at least:
`corosync token timeout + consensus timeout + stonith-watchdog-timeout`

-- Of course if `stonith-watchdog-timeout` is `-1`, this factor should be `SBD_WATCHDOG_TIMEOUT * 2`.

As we can see, for either case, depending on how many factors are under control/awareness of bootstrap,  it needs to be considered if/how to enable `SBD_DELAY_START`.

Another thing we need to ensure is TimeoutStartSec (defaults to 90s) of sbd.service should be longer that the practical value of start delay.

If it's considered to be too much of change/consideration for the time being, we could either first do `SBD_DELAY_START=yes` for the case with shared disk for now and do improvements in the future.

gao-yan (commented):
Good idea. Indeed it makes sense to enable `SBD_DELAY_START` for virtualization environment as long as pacemaker.service is enabled on system boot.

crmsh/sbd.py
liangxin1300: Thanks for clarifying!
So we can merge this PR now?

gao-yan: I'm only hesitant about the case of diskless sbd. Although `SBD_DELAY_START=yes` doesn't do any harm in that case, but also doesn't take any effect.  Not sure if it's a little confusing to users ...

OTOH, I actually have the idea to improve sbd for that to take effect for diskless sbd.

I'll leave it to you to decide whether to set `SBD_DELAY_START=yes` only for sbd with shared disk for now, or set it for diskless sbd as well and let users wait for it to take effect whenever sbd is updated with the enhancement.

BTW, I think so far `msgwait` set by bootstrap is for sure less than 90s (default `TimeoutStartSec`), right?

gao-yan (approved):

crmsh/sbd.py
liangxin1300: > I'm only hesitant about the case of diskless sbd. Although `SBD_DELAY_START=yes` doesn't do any harm in that case, but also doesn't take any effect. Not sure if it'd be a little confusing to users ...
> 
> OTOH, I actually have the idea to improve sbd for that to take effect for diskless sbd.
> 
> I'll leave it to you to decide whether to set `SBD_DELAY_START=yes` only for sbd with shared disk for now, or set it for diskless sbd as well and let users wait for it to take effect whenever sbd is updated with the enhancement.
> 

I think it should be more clear that set `SBD_DELEY_START=yes` only for sbd with shared disk,
keep `SBD_DELAY_START=no` when using diskless sbd

> BTW, I think so far `msgwait` set by bootstrap is for sure less than 90s (default `TimeoutStartSec`), right?

No, `msgwait` was not set by bootstrap, I'm wondering where this value been set?
In my environment, after bring up sbd, the msgwait is 10, is that good enough?
```
# sbd -d /dev/sda1 dump
==Dumping header on disk /dev/sda1
Header version     : 2.1
UUID               : 486e6ae8-a570-47ce-b351-98d82f6a598d
Number of slots    : 255
Sector size        : 512
Timeout (watchdog) : 5
Timeout (allocate) : 2
Timeout (loop)     : 1
Timeout (msgwait)  : 10
==Header on disk /dev/sda1 is dumped
```

gao-yan: Then it's defaulted on "sbd create". It's fine then.

PR #836, opened 2021-06-25
Fix: resource: make untrace consistent with trace

petrpavlu:
* Split logic in `do_untrace()` into separate methods `_untrace_resource()`, `_untrace_op()` and `_untrace_op_interval()` in the same way how the trace logic is organized.
* When processing `untrace <rsc> <op>`, look up matching operations and remove the trace_ra flag from all of them. This deals with a situation when a stateful resource has two monitor operations, `trace <rsc> monitor` would set the trace_ra flag on both of them but a subsequent call to `untrace <rsc> monitor` would fail to remove these flags.
* When processing `untrace <rsc> probe`, search only for a monitor operation with interval="0". This improves consistency between `trace <rsc> probe` and `untrace <rsc> probe`. The trace variant always sets the trace_ra flag on a monitor operation with interval="0". The untrace variant now only removes the flag from this exact operation while it was previously possible that it could remove it from a normal monitor operation if no probe operation was defined in CIB.
* Add unit tests for the trace+untrace functionality.

Clean ups:
* When removing the trace flag from an operation, do not look for `trace_ra` attributes but only `name=trace_ra` as the former seems no longer needed.
* Expand xpaths `.//op` to `operations/op` and similarly be explicit about accessing `instance_attributes/nvpair`.
* Remove unused variables and imports.

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -644,7 +644,8 @@ class RscMgmt(command.UI):
         return rsc.add_operation(n)
 
     def _trace_resource(self, context, rsc_id, rsc):
-        op_nodes = rsc.node.xpath('.//op')
+        """Enable RA tracing for a specified resource."""
+        op_nodes = rsc.node.xpath('operations/op')
 
         def trace(name):
             for o in op_nodes:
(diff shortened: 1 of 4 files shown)


liangxin1300:
@petrpavlu Thanks!

Looks good to me and great work!
I just confirmed this issue on sle15 and sle12sp5, after applying your commit, the issue has gone!

BTW, I've backported it to branch crmsh-4.3 for sle15 and crmsh-4.1 for sle12sp4+

liangxin1300 (commented):
@petrpavlu Thanks!

Looks good to me and great work!
I just confirmed this issue on sle15 and sle12sp5, after applying your commit, the issue has gone!

BTW, I've backported it to branch crmsh-4.3 for sle15 and crmsh-4.1 for sle12sp4+

liangxin1300 (approved):

PR #840, opened 2021-06-29
Fix: bootstrap: check for missing fields in 'crm_node -l' output (bsc#1182131)

petrpavlu:
The bootstrap code expects that each line in the output from `crm_node -l` is in the following form: `<id> <nodename> <state>`, with all three fields always present. This is the usual situation but it is in some cases possible that `<nodename>` and/or `<state>` can be missing. For instance, a node can be configured in the `nodelist` in `corosync.conf` but until the node has joined the cluster, `crm_node -l` will report only its id. This configuration can even happen as a result of a previous incomplete `crm cluster join` run.

Two places that parse `crm_node -l` output in the bootstrap code are: in `setup_passwordless_with_other_nodes()` when setting up passwordless SSH, in `join_cluster()` when updating nodeids for IPv6. These functions can fail with `ValueError: not enough values to unpack (expected 3, got 1)` or `IndexError: list index out of range` when a nodename is missing in the output.

The proposed patch adds error handling for this situation. The code newly reports a warning if a node state is incomplete and provides information what operation is not possible to perform on the node. This allows bootstrap of the new node to continue but indicates to the user that an update of the old problematic node might be needed if it should rejoin the cluster again in future.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1671,9 +1671,21 @@ def setup_passwordless_with_other_nodes(init_node):
         error("Can't fetch cluster nodes list from {}: {}".format(init_node, err))
     cluster_nodes_list = []
     for line in out.splitlines():
-        _, node, stat = line.split()
-        if stat == "member":
-            cluster_nodes_list.append(node)
+        # Parse line in format: <id> <nodename> <state>, and collect the
+        # nodename.
+        tokens = line.split()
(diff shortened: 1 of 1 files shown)


liangxin1300 (approved):
@petrpavlu 
LGTM!
Thanks for the nice commit!

I've created a backport one for sle15 on #841

PR #842, opened 2021-07-06
Change corosync and sbd parameters based on the profile environment detected

liangxin1300:
## Motivation
In different environment, some corosync and sbd parameters should be adjusted.
It should be nice when bootstrapping if crmsh could detect the environment, then load the different profile data for different platforms.
## Changes
- I add `/etc/crm/profiles.yml` for different platforms
```
# The valid profile names are:
# "microsoft-azure", "google-cloud-platform", "amazon-web-services", "s390", "default"
#
# "default" profile is loaded in the beginning.
#
# Those specific profile will override the corresponding values in "default"
# profile if the specific environment is detected.
#
# Users could customize the "default" profile for their needs, for example,
# those on-premise environments which is not defined yet.
#
# Profiles are only loaded on bootstrap init node.
#
# More details please see man corosync.conf, man sbd

default:
  corosync.totem.crypto_hash: sha1
  corosync.totem.crypto_cipher: aes256
  corosync.totem.token: 5000
  corosync.totem.join: 60
  corosync.totem.max_messages: 20
  corosync.totem.token_retransmits_before_loss_const: 10
  # sbd.msgwait is set to sbd.watchdog_timeout*2 by crmsh
  # or, you can define your own value in profiles.yml
  sbd.watchdog_timeout: 15

microsoft-azure:
  corosync.totem.token: 30000
  sbd.watchdog_timeout: 60
```
- The whole init process in Azure
When starting pacemaker, show the time should wait for sbd(sbd.watchdog_timeout * 2 * 1.2)
```
# crm cluster init -s /dev/sdc1 -y
  Loading "default" profile from /etc/crm/profiles.yml
  Detected "microsoft-azure" platform
  Loading "microsoft-azure" profile from /etc/crm/profiles.yml
  Configuring csync2
  Generating csync2 shared key (this may take a while)...done
  csync2 checking files...done
  Configuring corosync (unicast)
  Initializing SBD......done
  Hawk cluster interface is now running. To see cluster status, open:
    https://10.0.0.4:7630/
  Log in with username 'hacluster'
  Starting pacemaker(waiting for sbd 144s)...done
  Waiting for cluster..............done
  Loading initial cluster configuration
  Done (log saved to /var/log/crmsh/ha-cluster-bootstrap.log)
```
- When starting cluster service, show the time should wait for sbd
```
# crm cluster start
Starting pacemaker(waiting for sbd 144s)...done
INFO: Cluster services started
```
- sbd dump after bootstrapping in Azure
```
# sbd -d /dev/sdb1 dump
==Dumping header on disk /dev/sdb1
Header version     : 2.1
UUID               : c1489a11-02f6-4475-ac98-daa5e119d44e
Number of slots    : 255
Sector size        : 512
Timeout (watchdog) : 60
Timeout (allocate) : 2
Timeout (loop)     : 1
Timeout (msgwait)  : 120
==Header on disk /dev/sdb1 is dumped
```
- Cat /etc/sysconfig/sbd in Azure
```
# cat /etc/sysconfig/sbd|grep SBD_WATCHDOG_TIMEOUT
SBD_WATCHDOG_TIMEOUT=60
```
- show crmsh config in Azure
```
# crm configure show
node 1: crmsh15sp2-1
primitive stonith-sbd stonith:external/sbd \
	params pcmk_delay_max=30s
property cib-bootstrap-options: \
	have-watchdog=true \
	dc-version="2.0.4+20200616.2deceaa3a-3.6.1-2.0.4+20200616.2deceaa3a" \
	cluster-infrastructure=corosync \
	cluster-name=hacluster \
	stonith-enabled=true \
	stonith-timeout=172
rsc_defaults rsc-options: \
	resource-stickiness=1 \
	migration-threshold=3
op_defaults op-options: \
	timeout=600 \
	record-pending=true
```
- corosync.conf in Azure
```
        token: 30000
        max_messages: 20
        token_retransmits_before_loss_const: 10

```


crmsh/sbd.py
```diff
@@ -150,8 +151,12 @@ def _initialize_sbd(self):
         """
         if self.diskless_sbd:
             return
+        opt = ""
+        if self._context.profiles_dict:
+            self._sbd_msgwait = self._context.profiles_dict["sbd_msgwait"]
+            opt = "-4 {}".format(self._sbd_msgwait)
```
liangxin1300, line 157: @gao-yan Do we need use `-1` option to setup watchdog timeout when creating sbd? Or, just write the SBD_WATCHDOG_TIMEOUT in /etc/sysconfig/sbd?
What's different here?

zzhou1: This need something like
`sbd -d $SBD_DEVICE -4 $sbd_msgwait -1 $sbd_watchdog_timeout create`

Keep in mind, watchdog_timeout on disk overrides `/etc/sysconfig/sbd SBD_WATCHDOG_TIMEOUT`

crmsh/sbd.py
```diff
@@ -241,6 +248,28 @@ def _enable_sbd_service(self):
             # in init process
             bootstrap.invoke("systemctl enable sbd.service")
 
+    def _adjust_systemd(self):
+        """
+        Adjust start timeout for sbd when has profiles data
+        """
+        if not self._context.profiles_dict:
```
zzhou1, line 255: The condition has very very weak relationship with `profiles_dict`. And the logic must not be coupled with cloud directly. In general, it might happens to other virtualization environment.

Instead, the condition checks if SBD_DELAY_START configuration is 'no', and the function return if so. And, this is the advanced step after _update_configuration() to enable SBD_DELAY_START in virtualization environment.

zzhou1 (changes requested):

crmsh/sbd.py
liangxin1300: Changed, now the dump output:
```
# sbd -d /dev/sdb1 dump
==Dumping header on disk /dev/sdb1
Header version     : 2.1
UUID               : c1489a11-02f6-4475-ac98-daa5e119d44e
Number of slots    : 255
Sector size        : 512
Timeout (watchdog) : 60
Timeout (allocate) : 2
Timeout (loop)     : 1
Timeout (msgwait)  : 120
==Header on disk /dev/sdb1 is dumped
```

crmsh/sbd.py
liangxin1300: OK, changed when SBD_DELAY_START set

zzhou1 (approved):

zzhou1 (changes requested):
Please re-split the commits: one to accomodate Azure parameters, another one to deal with TimeoutStartUSec alone. Their relationship is not strong enough to mix them together.

crmsh/constants.py
```diff
@@ -490,4 +490,16 @@
 
 
 SSH_OPTION = "-o StrictHostKeyChecking=no"
+
+
+PROFILES_DICT = {
+        "Azure": {
```
arbulu89, line 496: It would be good to have this `Azure` as a constant, or even some kind of enum, as it is used as string in other places, and there are chances of doing some error if the string is changed

crmsh/sbd.py
```diff
@@ -241,6 +250,28 @@ def _enable_sbd_service(self):
...
 
+    def _adjust_systemd(self):
+        """
+        Adjust start timeout for sbd when set SBD_DELAY_START
+        """
+        if not self._delay_start:
+            return
+
+        # TimeoutStartUSec default is 1min 30s, need to parse as seconds
+        cmd = "systemctl show service -p TimeoutStartUSec|\
```
arbulu89, line 261: This looks super anti-python.
Why don't you open the file and apply python regular expressions. Using `grep, sed, awk` looks horrible

arbulu89 (dismissed):
@liangxin1300 ,
I have added a pair of comment I think you should check. 

Besides them, some concerns:
- I don't see any new UT here, and the change looks enough sensitive to be good to have them
- I don't personally like this option to hardcode some provider specific values. If for some reason they change, you will need to change the code and package a new version... And more, knowing that we already have projects that do the same (like the `habootstrap-formula`). Personally, I don't like this way. In the other hand, if you have decided to go this way, it is fine, I'm just talking as a more outsider point of view

liangxin1300:
> @liangxin1300 ,
> I have added a pair of comment I think you should check.
> 
> Besides them, some concerns:
> 
> * I don't see any new UT here, and the change looks enough sensitive to be good to have them

Yeah, I will add UT codes for this, after most of logic archived agreement between us:)

> * I don't personally like this option to hardcode some provider specific values. If for some reason they change, you will need to change the code and package a new version... And more, knowing that we already have projects that do the same (like the `habootstrap-formula`). Personally, I don't like this way. In the other hand, if you have decided to go this way, it is fine, I'm just talking as a more outsider point of view

The issue I concerned is, where can we get the data/values, from only one place, that could be shared with all HA stack projects? I remember we had discussions in confluence page before, Dario suggest we make a specific package which contains all profiles data, and then all projects could use that
@zzhou1 @gao-yan Any suggestions here?

zzhou1:
> > @liangxin1300 ,
> > I have added a pair of comment I think you should check.
> > Besides them, some concerns:
> > 
> > * I don't see any new UT here, and the change looks enough sensitive to be good to have them
> 
> Yeah, I will add UT codes for this, after most of logic archived agreement between us:)
> 
> > * I don't personally like this option to hardcode some provider specific values. If for some reason they change, you will need to change the code and package a new version... And more, knowing that we already have projects that do the same (like the `habootstrap-formula`). Personally, I don't like this way. In the other hand, if you have decided to go this way, it is fine, I'm just talking as a more outsider point of view
> 
> The issue I concerned is, where can we get the data/values, from only one place, that could be shared with all HA stack projects? I remember we had discussions in confluence page before, Dario suggest we make a specific package which contains all profiles data, and then all projects could use that
> @zzhou1 @gao-yan Any suggestions here?

Sounds like, some improvement is possible indeed. How about move the profile dictionary and also the matching string (eg. "microsoft-azure", "google-cloud-platform", "amazon-web-services") into crm.conf. Hence, it can be flexible enough for any one of these CSPs to further customize those values by users, if they don't like our default.

crmsh/sbd.py
liangxin1300: Changed, thanks!

liangxin1300:
@zzhou1 @arbulu89 I put the profile into `/etc/crm/profiles.yml`
What's your think?

crmsh/sbd.py
liangxin1300: @zzhou1 @gao-yan In virtualization environment but not in Azure, what the value should be for sbd msgwait? Keep default or seem with Azure?

zzhou1 (changes requested):
I suggest to tweak the commit header 

from 
"Fix: sbd: adjust sbd parameters when in Azure(bsc#1175896)"

to 
"Fix: sbd: adjust sbd systemd TimeoutStartSec together with SBD_DELAY_START"

It is the genric logic more than just Azure(bsc#1175896).

etc/profiles.yml
```diff
@@ -0,0 +1,7 @@
+microsoft-azure:
+  corosync_token: 30000
+  corosync_consensus: 36000
+  corosync_max_messages: 20
+  corosync_token_retransmits_before_loss_const: 10
+  sbd_watchdog_timeout: 60
+  sbd_msgwait: 120
```
zzhou1, line 7: It probably makes sense to add a small comments in the beginning to describe those valid key definitions, example,  

# The valid profile names are: microsoft-azure, google-cloud-platform", "amazon-web-services.
#

liangxin1300:
@zzhou1 @gao-yan If in virtualization environment but not in Azure, what the value should be for sbd msgwait? Keep default or same with Azure?

crmsh/bootstrap.py
```diff
@@ -186,6 +189,20 @@ def init_sbd_manager(self):
         from .sbd import SBDManager
         self.sbd_manager = SBDManager(self)
 
+    def load_profiles(self):
+        """
+        Load profiles data for different cloud provider
+        """
+        env_name = utils.detect_cloud()
+        if not env_name:
+            return
```
zzhou1, line 198: In stead of directly "return". I suggest to set

env_name = "user_customized"

This makes the log more generic and gives users the opportunity to define the "user_customized" profile in "profiles.yaml", for example, when their environment is on-premise.

zzhou1 (changes requested):
It makes sense to tweak the commit header from 
"Fix: bootstrap: bootstrap parameters better cope with Azure for corosync(bsc#1175896)"

to something like,
"Fix: bootstrap: corosync and sbd parameters better cope with the environment profile"

Since the code is generic to both CSP and on-premise.

zzhou1:
> @zzhou1 @gao-yan If in virtualization environment but not in Azure, what the value should be for sbd msgwait? Keep default or same with Azure?

Aha, then, you brought up three folds in my view.

First of all, don't use Azure's default sbd parameters for others.

Secondly, I think you touch the part I call it "user_customized" in profiles.yaml, in particular, those ones not in CSP.

Thirdly, there are several hardcoded defaults values in the code currently. That's fine. They are the last resort if no "profiles.yaml" exisits. However, I feel it worth to bring them up front and write them into "user_customized" explicitly. These will slightly improve the user experience than the block magic only known by reading the code. Well, this is not a high priority improvment, I feel.

crmsh/bootstrap.py
```diff
@@ -1164,6 +1182,18 @@ def gen_mcastaddr():
...
+    """
+    Adjust corosync's parameters according profiles
+    """
+    if not _context.profiles_dict:
+        return
+    corosync.set_value("totem.token", _context.profiles_dict["corosync_token"])
+    corosync.set_value("totem.consensus", _context.profiles_dict["corosync_consensus"])
+    corosync.set_value("totem.max_messages", _context.profiles_dict["corosync_max_messages"])
+    corosync.set_value("totem.token_retransmits_before_loss_const", _context.profiles_dict["corosync_token_retransmits_before_loss_const"])
+
```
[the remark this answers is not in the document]
liangxin1300: Changed profile as
```
user_customized:
  totem.token: 5000
  totem.consensus: 6000
  totem.join: 60
  totem.max_messages: 20
  totem.token_retransmits_before_loss_const: 10
  sbd.watchdog_timeout: 5
  sbd.msgwait: 10

microsoft-azure:
  totem.token: 30000
  totem.consensus: 36000
  totem.max_messages: 20
  totem.token_retransmits_before_loss_const: 10
  sbd.watchdog_timeout: 60
  sbd.msgwait: 120
```

Then if any key starts with "totem.", we can set this key's value

crmsh/bootstrap.py

liangxin1300:
> > @zzhou1 @gao-yan If in virtualization environment but not in Azure, what the value should be for sbd msgwait? Keep default or same with Azure?
> 
> Aha, then, you brought up three folds in my view.
> 
> First of all, don't use Azure's default sbd parameters for others.
> 
> Secondly, I think you touch the part I call it "user_customized" in profiles.yaml, in particular, those ones not in CSP.
> 
> Thirdly, there are several hardcoded defaults values in the code currently. That's fine. They are the last resort if no "profiles.yaml" exisits. However, I feel it worth to bring them up front and write them into "user_customized" explicitly. These will slightly improve the user experience than the black magic only known by reading the code. Well, this is not a high priority improvment, I feel.

Changed as
```
user_customized:
  totem.token: 5000
  totem.consensus: 6000
  totem.join: 60
  totem.max_messages: 20
  totem.token_retransmits_before_loss_const: 10
  sbd.watchdog_timeout: 5
  sbd.msgwait: 10

microsoft-azure:
  totem.token: 30000
  totem.consensus: 36000
  totem.max_messages: 20
  totem.token_retransmits_before_loss_const: 10
  sbd.watchdog_timeout: 60
  sbd.msgwait: 120
```

crmsh/bootstrap.py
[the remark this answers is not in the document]
zzhou1: In theory, the user experience can be more robust if the code could validate all keys in profiles.yml. The error message should be given if any typo, for example, a missing "s" character as below,

eg. totem.consensu: 6000
eg. amazon-web-service:

etc/profiles.yml
```diff
@@ -0,0 +1,19 @@
+# The valid profile names are:
+# "microsoft-azure", "google-cloud-platform", "amazon-web-services, "user_customized"
```
zzhou1, line 2: a very small typo of the missing " mark here.

`# "microsoft-azure", "google-cloud-platform", "amazon-web-services, "user_customized"`

How about 

```
# The valid profile names are: microsoft-azure, google-cloud-platform, amazon-web-services, user_customized
# 
# NOTE: The `user_customized` profile will be loaded in the beginning as the default values, and parameters will be overriden by the correspondig profile environment if detected.

```

crmsh/bootstrap.py
```diff
@@ -186,6 +190,19 @@ def init_sbd_manager(self):
...
 
+    def load_profiles(self):
+        """
+        Load profiles data for different cloud provider
+        """
+        if not os.path.exists(PROFILES_FILE):
+            return
+        env_name = utils.detect_cloud()
+        if not env_name:
+            env_name = USER_CUSTOMIZED
```
zzhou1, line 201: revisit this today, I think the correct and better logic is to load  the USER_CUSTOMIZED profile as the defaults and let it overriden by the correspondig profile environment if detected.

etc/profiles.yml
```diff
@@ -0,0 +1,19 @@
...
+
+user_customized:
+  totem.token: 5000
+  totem.consensus: 6000
+  totem.join: 60
+  totem.max_messages: 20
+  totem.token_retransmits_before_loss_const: 10
+  sbd.watchdog_timeout: 5
+  sbd.msgwait: 10
```
zzhou1, line 11: Maybe move other hard-coded parameters here too, for example,

    secauth:    on
    crypto_hash:    sha1
    crypto_cipher:  aes256
    cluster_name:   %(clustername)s
    clear_node_high_bit: yes

zzhou1 (changes requested):

crmsh/bootstrap.py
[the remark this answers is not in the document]
zzhou1: BTW, about the commit header, how about somethign like,
```
Fix: bootstrap: adjust corosync and sbd parameters according to the profile environment detected (bsc#1175896)
…

File: crmsh/bootstrap.py (hunk 9 above)
Comment (liangxin1300) [reply crmsh/bootstrap.py:201]: It's the same here
If detected one of three could providers, then use that profiles data; else, use the default one

File: crmsh/bootstrap.py [hunk 11]
```diff
@@ -186,6 +190,20 @@ def init_sbd_manager(self):
...
+        with open(PROFILES_FILE) as f:
+            data = yaml.load(f, Loader=yaml.SafeLoader)
+        if env_name in data:
```
liangxin1300, line 204: Since we don't have profiles of GCP and AWS, here should check if the key in the dict

crmsh/bootstrap.py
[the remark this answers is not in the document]
liangxin1300: > BTW, about the commit header, how about somethign like,
> 
> Fix: bootstrap: adjust corosync and sbd parameters according to the profile environment detected (bsc#1175896)

Changed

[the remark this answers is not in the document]
liangxin1300: > In theory, the user experience can be more robust if the code could validate all keys in profiles.yml. The error message should be given if any typo, for example, a missing "s" character as below,
> 
> eg. totem.consensu: 6000
> eg. amazon-web-service:

This is not easy or might not worth to do
Currently, in the worst case, like profiles.yml not exist, crmsh can make sure bootstrap process could finished(token and sbd all using default parameters)

etc/profiles.yml
liangxin1300: Changed as
```
# The valid profile names are:
# "microsoft-azure", "google-cloud-platform", "amazon-web-services", "user_customized"
# Default is "user_customized" if no cloud environment detected
```

etc/profiles.yml
liangxin1300: Not suggest since these are rarely changed

crmsh/bootstrap.py
zzhou1: Okay, fine, it can work. It depends on how we interpret the usage of profiles.yml indeed.

etc/profiles.yml
zzhou1: The point to add them here is trying to define a "full" list of parameters. What's is concern me a little is we don't clearly provide the full list. Therefore users are blind on this.

I agree, often, only the subset is needed. 

For example, we can see some AWS users doesn't use crypto. Well, probably the majority stay with crypto.
https://docs.aws.amazon.com/sap/latest/sap-hana/sap-hana-on-aws-cluster-configuration.html

crmsh/bootstrap.py
zzhou1: make sense

crmsh/bootstrap.py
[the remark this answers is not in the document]
zzhou1: Fine. I don't see the priority to justify the effort too.

etc/profiles.yml
```diff
@@ -0,0 +1,25 @@
+# The valid profile names are:
+# "microsoft-azure", "google-cloud-platform", "amazon-web-services", "default"
+# Default is "default" if no cloud environment detected
+# More details please see `man corosync.conf`
+
+default:
+  totem.crypto_hash: sha1
+  totem.crypto_cipher: aes256
+  totem.token: 5000
+  totem.consensus: 6000
```
gao-yan, line 10: Let's default ` totem.consensus`, so that corosync automatically calculate a proper value at runtime.

etc/profiles.yml
```diff
@@ -0,0 +1,25 @@
...
+  totem.max_messages: 20
+  totem.token_retransmits_before_loss_const: 10
+  sbd.watchdog_timeout: 5
+  sbd.msgwait: 10
+
+microsoft-azure:
+  totem.crypto_hash: sha1
+  totem.crypto_cipher: aes256
+  totem.token: 30000
+  totem.consensus: 36000
```
gao-yan, line 21: Similar here.

etc/profiles.yml
gao-yan: Let's comment out sbd settings here, so that `sbd create` chooses the proper defaults itself, which will be especially sensible for s390(x).

crmsh/sbd.py
```diff
@@ -161,10 +168,12 @@ def _update_configuration(self):
         """
         shutil.copyfile(self.SYSCONFIG_SBD_TEMPLATE, SYSCONFIG_SBD)
         self._determine_sbd_watchdog_timeout()
+        if utils.detect_virt() and self._sbd_devices:
+            self._delay_start = True
```
gao-yan, line 172: BTW, given that Emil's PR for enabling `SBD_DELAY_START=yes` for the case of diskless sbd has been merged upstream:
https://github.com/ClusterLabs/sbd/pull/133

, so we could enable it here for diskless sbd as well. Probably it should belong to a separate PR/commit.

crmsh/sbd.py
```diff
@@ -241,6 +250,30 @@ def _enable_sbd_service(self):
...
 
+    def _adjust_systemd(self):
+        """
+        Adjust start timeout for sbd when set SBD_DELAY_START
+        """
+        if not self._delay_start:
+            return
+
+        # TimeoutStartUSec default is 1min 30s, need to parse as seconds
+        cmd = "systemctl show service -p TimeoutStartUSec"
```
gao-yan, line 261: It should be more sensible to just check this for `sbd.service`.

etc/profiles.yml
zzhou1: > Let's default ` totem.consensus`, so that corosync automatically calculates a proper value based on `totem.token` and so on at runtime.

Agree. Well, I do see the "pain" from the history for about , almost every

liangxin1300: Do you mean not write `consensus` in corosync.conf?
Agree

crmsh/sbd.py
liangxin1300: OK, I will create another PR for this

crmsh/sbd.py
liangxin1300: `systemctl show sbd.service -p TimeoutStartUSec`?

zzhou1: > It should be more sensible to directly check this for `sbd.service`. And parsing could be easier by supplying `--value` option here.

aha, yes, make sense, `systemctl show -p TimeoutStartUSec sbd --value`

We have hard time to find if systemd can generate "90s" instead of "1m 30sec", but systemd don't provide, sigh

etc/profiles.yml
liangxin1300: OK

crmsh/sbd.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, context):
         self._sbd_watchdog_timeout = 0
         self._is_s390 = "390" in os.uname().machine
         self._context = context
+        self._sbd_msgwait = 10
```
gao-yan, line 50: Be aware correspondingly, if msgwait is defaulted on `sbd create`, sbd uses 30 for s390(x) otherwise 10, basically 2 * default_watchdog_timeout.

etc/profiles.yml
zzhou1: Instead of delete sbd in "default", I prefer to add "s390" profile explicitly to override the default values. This avoid the hard-coded value for s390 and users just not aware them. This fits well for the concept to load profiles based on the environment.

crmsh/sbd.py
liangxin1300: @gao-yan I added this profile, and make sure this will be loaded when in s390:
```
s390:
  sbd.watchdog_timeout: 15
  sbd.msgwait: 30
```

etc/profiles.yml
gao-yan: > Instead of delete sbd in "default", I prefer to add "s390" profile explicitly to override the default values. This avoid the hard-coded value for s390 and users just not aware them. This fits well for the concept to load profiles based on the environment.

Profiles are basically configurable. Any settings here could be changed or even unset by users... It's more or less another abuse of the term "default" in here :-) I mean the term still could be used in the context of profiles, but settings here shouldn't be considered by us as the "real internal defaults".

zzhou1: You are right. "default" is not intended to be the "real internal defaults". It's naming is tricky. It's not perfect, we do agree.  Initially, we give it name "user_customized". Later on we name it to "default" for users to read the execution log more intuitive, 

eg. Loading form "default" profile from /etc/crmsh/profiles.yml. 

This "default" profile indeed let users to customize their own need, particular for those on-premise environments which is our of our pre-defined profiles in the yaml file.

gao-yan: > This avoid the hard-coded value for s390 and users just not aware them.

And I think this is something that should be documented better in sbd manpage. Despite how it should be like in the shipped profiles, code-wise crmsh should anyway be able to handle the case where the settings are truly defaulted here.

crmsh/sbd.py
gao-yan: Like mentioned, in case of s390, if an user unset this setting from profile s390 or even deleted the profile s390, it'd be a wrong value to be used.

OTOH, frankly I don't see much value of explicitly duplicating the internal default values in the shipped profiles. We anyway could not rely on them to be as internal defaults, since they could be changed or unset by users.

We should make it somehow work properly in cases of real "default".

* Any settings are unset in profiles
* Any specific profiles are missing
* Probably even the whole profiles.yml is missing

 by defaulting

etc/profiles.yml
zzhou1: BTW, sounds make sense to have the sbd values in "default" profile for the gernal use according to bsc#1122604. 

sbd msgwait timeout 30s
watchdog timeout 15s

gao-yan: Yes, definitely sensible to list them if they are different from the internal defaults.

crmsh/sbd.py
```diff
@@ -278,6 +310,7 @@ def sbd_init(self):
             self._update_configuration()
         self._determine_stonith_watchdog_timeout()
         self._enable_sbd_service()
+        self._adjust_systemd()
```
zzhou1, line 313: BUG: _adjust_systemd() is needed at the join node too.

crmsh/sbd.py
zzhou1: There are prirority levels. The intuitive order would be,

specific profile if detected > "default" profile > crmsh defaults if any > internal defaults of sbd/corosync

These priority logic is simple and consisence enough in my view. Even, it can be evolve further. For example, if any users has demand, we can implement the logic further to allow users to manage the sperate "profile-01-xxxx.yml", "profile-02-xxxx.yml", if they want. These file will have the highest priority to override all others.

zzhou1 (approved):

liangxin1300:
@arbulu89 Any suggestions? Thanks!

crmsh/sbd.py
```diff
@@ -212,6 +211,30 @@ def _get_sbd_device_from_config(self):
...
+        if not res:
+            raise ValueError("Cannot get the value of SBD_WATCHDOG_TIMEOUT")
+        return int(res)
+
+    @staticmethod
+    def get_sbd_systemd_timeout():
+        """
+        Systemd start timeout for sbd.service
+        """
+        return int(SBDManager.get_sbd_watchdog_timeout() * 2 * 1.2)
```
zzhou1, line 236: For more robustness and correctness, how about,
```
if self.diskless_sbd:
    return int(SBDManager.get_sbd_watchdog_timeout() * 2 * 1.2)
else:
    return int(SBDManager.get_sbd_msgwait() * 1.2)
```

zzhou1 (changes requested):

crmsh/sbd.py
liangxin1300: Changed
Thanks!

crmsh/bootstrap.py
```diff
@@ -186,6 +195,61 @@ def init_sbd_manager(self):
...
+    def determine_profile_type(self):
+        """
+        Determine profile type
+        """
+        if self.is_s390:
+            profile_type = self.S390_PROFILE_NAME
+        elif self.cloud_type:
+            profile_type = self.cloud_type
+        else:
+            return None
```
arbulu89, line 215: This looks a typical place where an Exception is raised, but this `None` option can be good in any case

crmsh/bootstrap.py
```diff
@@ -186,6 +195,61 @@ def init_sbd_manager(self):
...
+        """
+        profile_dict = {}
+        if not profile_type:
+            return profile_dict
+
+        if profile_type in self.profiles_data:
+            status("Loading \"{}\" profile from {}".format(profile_type, self.PROFILES_FILE))
+            profile_dict = self.profiles_data[profile_type]
+        else:
+            status("\"{}\" profile is not exist in {}".format(profile_type, self.PROFILES_FILE))
```
arbulu89, line 231: Change the msg by `profile does not exist`

crmsh/bootstrap.py
```diff
@@ -1164,26 +1241,32 @@ def gen_mcastaddr():
     csync2_update(corosync.conf())
 
 
+def profiles_adjust_for_corosync():
```
arbulu89, line 1244: I don't understand this method. Why do we call it `adjust_for_corosync` and we only check the `totem` entry?
What happens with everything else.
If we only want to modify the totem, maybe we should rename the method `to adjust_corosync_totem`

arbulu89 (commented):
@liangxin1300 Added some minor comments.
I don't have any major change to request.

zzhou1 (approved):

crmsh/bootstrap.py
liangxin1300: Changed, thank you!

crmsh/bootstrap.py
liangxin1300: Changed as a more flexible resolution

crmsh/sbd.py
```diff
@@ -150,8 +154,13 @@ def _initialize_sbd(self):
         """
         if self.diskless_sbd:
             return
+        opt = ""
+        if self._context.profiles_dict:
+            sbd_msgwait = self._context.profiles_dict["sbd.msgwait"]
+            self._sbd_watchdog_timeout = self._context.profiles_dict["sbd.watchdog_timeout"]
+            opt = "-4 {} -1 {}".format(sbd_msgwait, self._sbd_watchdog_timeout)
```
gao-yan, line 161: what if no msgwait is explicitly set in the profiles?

etc/profiles.yml
```diff
@@ -0,0 +1,35 @@
...
+microsoft-azure:
+  corosync.totem.token: 30000
+  corosync.totem.max_messages: 20
+  corosync.totem.token_retransmits_before_loss_const: 10
+  sbd.watchdog_timeout: 60
+  sbd.msgwait: 120
+
+s390:
+  sbd.watchdog_timeout: 15
+  sbd.msgwait: 30
```
gao-yan, line 35: Wonder why we need to duplicate them for s390 given that they are the same as the default.

crmsh/sbd.py
liangxin1300: Changed the logic, and removed the sbd.msgwait from profiles.yml
only detected if "sbd.watchdog_timeout" in profiles.yml
then msgwait will be calculated

crmsh/bootstrap.py
```diff
@@ -696,10 +761,22 @@ def init_cluster_local():
...
+    start_pacemaker()
     wait_for_cluster()
 
 
+def start_pacemaker():
+    """
+    Start pacemaker service with wait time for sbd
+    """
+    pacemaker_start_msg = "Starting pacemaker"
+    if utils.package_is_installed("sbd") and utils.service_is_enabled("sbd.service"):
```
zzhou1, line 773: BUG: start_pacemaker() need check self._delay_start, something like,
```
    if utils.package_is_installed("sbd") and utils.service_is_enabled("sbd.service") and self._delay_start:
```

zzhou1 (changes requested):

crmsh/bootstrap.py

liangxin1300:
@gao-yan I added this commit https://github.com/ClusterLabs/crmsh/pull/842/commits/bbfefa050b08ffc4739f8379d4307bc26284cf66
to make sure diskless sbd could load sbd_watchdog_value from profiles.yml if that value exists

My question is, should I `plus` self._sbd_watchdog_timeout in function _determine_sbd_watchdog_timeout?

crmsh/sbd.py
```diff
@@ -199,7 +215,7 @@ def _determine_stonith_watchdog_timeout(self):
         res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
         if res:
             self._stonith_watchdog_timeout = -1
-        elif self._is_s390:
+        elif self._context.is_s390:
             self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT_S390
```
zzhou1, line 219: By default `stonith_watchdog_timeout` is 0, aka. disable. We only need to turn it on for diskless sbd. 
`-1` means paceamker will auto-calculate it based on SBD_WATCHDOG_TIMEOUT. 
`-1` is good enough. No need additional lines to explicitly deal with s390 since now profiles.yml concept deals with SBD_WATCHDOG_TIMEOUT flexibly.

I'm thinking to change
```
    def _determine_stonith_watchdog_timeout(self):
```

to
```
    def _enable_stonith_watchdog_timeout_for_diskless_sbd(self):
        if not self.diskless_sbd:
            return
        res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
        if res:
            self._stonith_watchdog_timeout = -1
```

Good to have @gao-yan 's double check here

crmsh/sbd.py
```diff
@@ -184,7 +200,7 @@ def _determine_sbd_watchdog_timeout(self):
         if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
             qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
             self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
-            if self._is_s390 and self._sbd_watchdog_timeout < 15:
+            if self._context.is_s390 and self._sbd_watchdog_timeout < 15:
                 self._sbd_watchdog_timeout = 15
             self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
         # add sbd and qdevice together from beginning
```
zzhou1, line 206: The main point of the current _determine_sbd_watchdog_timeout() is trying to update _sbd_watchdog_timeout based on qdevice_sync_timeout. It worth to send a message to user sbd_watchdog_timeout is changed because of this.

I think this is also valid to disk-based-sbd too. Loop in @gao-yan to double check, TIA!

zzhou1 (changes requested):

crmsh/sbd.py
gao-yan: > The main point of the current _determine_sbd_watchdog_timeout() is trying to update _sbd_watchdog_timeout based on qdevice_sync_timeout. It worth to send a message to user sbd_watchdog_timeout is changed because of this.

Makes sense.

> 
> I think this is also valid to disk-based-sbd too. Loop in @gao-yan to double check, TIA!

I don't think it's needed for sbd with disk.

For the specific situation:
https://github.com/corosync/corosync-qdevice/issues/21#issuecomment-842282435

, in case of diskless sbd, it purely relies on update/loss of quorum to get itself reset to achieve fencing. But for the case sbd with disk, because of the different mechanism, the failing node will eat the poison pill and already get fenced before sync timeout is popped...

crmsh/sbd.py
gao-yan: I like the idea of calculating msgwait. But it doesn't seem to correctly handle the case where  `sbd.msgwait`  is set in profile but `sbd.watchdog_timeout` doesn't exist ...

I guess this could be a good place to reference predefined `SBD_WATCHDOG_TIMEOUT_DEFAULT` and `SBD_WATCHDOG_TIMEOUT_DEFAULT_s390`

crmsh/sbd.py
gao-yan: `stonith_watchdog_timeout=-1` relies on `SBD_WATCHDOG_TIMEOUT` to be explicitly set in sysconfig.  This is to handle the case where ` SBD_WATCHDOG_TIMEOUT` is unset in sysconfig.

But rather than handling it in here, it could be better to be initialized for s390 case also in above together with the normal case:
`self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT`

zzhou1: Yeah, to add the "else" clause does make sense,
```
        if res:
            self._stonith_watchdog_timeout = -1
        else:
            self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
```

crmsh/sbd.py
```diff
@@ -148,10 +152,20 @@ def _initialize_sbd(self):
         """
         Initialize SBD device
         """
+        opt = ""
+        if "sbd.watchdog_timeout" in self._context.profiles_dict:
+            self._sbd_watchdog_timeout = self._context.profiles_dict["sbd.watchdog_timeout"]
+            sbd_msgwait = int(self._sbd_watchdog_timeout) * 2
+            opt = "-4 {} -1 {}".format(sbd_msgwait, self._sbd_watchdog_timeout)
         if self.diskless_sbd:
             return
```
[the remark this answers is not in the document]
zzhou1: I think it's fine when we "redifine" the docstring to something like,
```
Initialize SBD parameters according to profiles.yml, or the crmsh defined defaulst as the last resort. 
This covers both disk-based-sbd, and diskless-sbd scenarios.
```

That's my thoughts to suggest to move the s3900 part here to consolidate all the related profile concept in one place.
Anyway, Xin might has his perference to organize the code appropriatly.

crmsh/sbd.py
```diff
@@ -45,26 +44,31 @@ def __init__(self, context):
         self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
         self._stonith_timeout = 60
         self._sbd_watchdog_timeout = 0
```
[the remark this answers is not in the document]
liangxin1300: Nice suggestions, thank you!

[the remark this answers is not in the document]
zzhou1: I think I agree with Yan to keep some crmsh hard-coded defaults. Though, Xin tried to avoid the deplication between `crmsh hard-code` and `profiles.yml` ;)

liangxin1300:
@zzhou1 @gao-yan 
Latest changes:
- Set some default value in `SBDManager` class
```
    SBD_WATCHDOG_TIMEOUT_DEFAULT = 5
    SBD_WATCHDOG_TIMEOUT_DEFAULT_S390 = 15
    STONITH_WATCHDOG_TIMEOUT_DEFAULT = SBD_WATCHDOG_TIMEOUT_DEFAULT * 2
    STONITH_WATCHDOG_TIMEOUT_DEFAULT_S390 = SBD_WATCHDOG_TIMEOUT_DEFAULT_S390 * 2
```
- Set above default value when init class
```Python
        if context.is_s390:
            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390
            self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT_S390
        else:
            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT
            self._stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
```
- In _initialize_sbd function, make sure diskless SBD also has the sbd_watchdog_timeout value from profile
```
        if "sbd.watchdog_timeout" in self._context.profiles_dict:
            self._sbd_watchdog_timeout = self._context.profiles_dict["sbd.watchdog_timeout"]
            self._correct_sbd_watchdog_timeout_for_s390()
        if self.diskless_sbd:
            return
```
- Make sure msgwait always 2*sbd_watchdog_timeout
```Python
        sbd_msgwait_default = int(self._sbd_watchdog_timeout) * 2
        sbd_msgwait = sbd_msgwait_default
        if "sbd.msgwait" in self._context.profiles_dict:
            sbd_msgwait = self._context.profiles_dict["sbd.msgwait"]
            if int(sbd_msgwait) < sbd_msgwait_default:
                bootstrap.warn("sbd msgwait is set to {}, it was {}".format(sbd_msgwait_default, sbd_msgwait))
                sbd_msgwait = sbd_msgwait_default
```
- Make sure sbd_watchdog_timeout never less than default on s390 platform
```Python
    def _correct_sbd_watchdog_timeout_for_s390(self):
        """
        Corrent watchdog timeout if less than s390 default
        """
        if self._context.is_s390 and self._sbd_watchdog_timeout < self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390:
            bootstrap.warn("sbd_watchdog_timeout is set to {} for s390, it was {}".format(self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390, self._sbd_watchdog_timeout))
            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390
```

crmsh/sbd.py
```diff
@@ -184,7 +200,7 @@ def _determine_sbd_watchdog_timeout(self):
         if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
             qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
             self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
```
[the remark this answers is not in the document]
liangxin1300: @gao-yan So here my question is, should we `plus`?
```Python
self._sbd_watchdog_timeout += qdevice_sync_timeout + 5
```
Or
```Python
if self._sbd_watchdog_timeout < qdevice_sync_timeout:
    self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
```

crmsh/sbd.py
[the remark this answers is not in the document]
gao-yan: This is why I'd just drop s390 profile to prevent unnecessary duplication there, given that it's consistent with internal default, and we have to have the internal default anyway.

Profile is a place where users configure things that are different from the defaults rather than where users get to know the defaults.

crmsh/sbd.py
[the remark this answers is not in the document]
gao-yan: The latter makes sense. And let's provide a "notice" messages so that users know the fact and the reason that a different value from their definition is used.

crmsh/sbd.py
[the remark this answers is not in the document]
zzhou1: You may want to tweak the docstring of the function.

crmsh/sbd.py
zzhou1: Thanks Yan, I have much better understanding of the point here now. Though, I think I will need more time to understand the complete picture in related to qdevice-sync-timeout. 

It looks things get simplified quite a bit in sbd-1.5.0 + corosync-3, the part with qdevice-sync-timeout will be deleted  then, though I'm 100% sure. If so, could be better to add a inline TODO to revisit this in sbd-1-5.0 + corosync-3 context.

crmsh/sbd.py
```diff
@@ -184,8 +218,7 @@ def _determine_sbd_watchdog_timeout(self):
...
-            if self._is_s390 and self._sbd_watchdog_timeout < 15:
-                self._sbd_watchdog_timeout = 15
+            self._correct_sbd_watchdog_timeout_for_s390()
             self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
         # add sbd and qdevice together from beginning
         elif self._context.qdevice_inst:
```
zzhou1, line 224: BUG: my reading is the following wrongly reset `self._sbd_watchdog_timeout` to crmsh's hard-coded defaults. It could have been loaded from profiles.yml in _initialize_sbd(), earlier.

crmsh/sbd.py
```diff
@@ -144,14 +155,35 @@ def _get_sbd_device(self):
             dev_list = self._get_sbd_device_interactive()
         self._sbd_devices = dev_list
 
+    def _correct_sbd_watchdog_timeout_for_s390(self):
```
zzhou1, line 158: Cosmetic feedback, ideally, you may consider to the better function names, for example, the same prefix "_adjust" or
"_update" for the following,

def _adjust_systemd()

def _update_configuration()
def update_configuration()  # these two are too close to tell the difference just from their names.

def _correct_sbd_watchdog_timeout_for_s390()
def _determine_stonith_watchdog_timeout()
def _determine_sbd_watchdog_timeout()

zzhou1 (changes requested):
Good progress, though still some changes needed.

crmsh/sbd.py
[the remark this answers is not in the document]
liangxin1300: Changed as
```Python
-            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
+            if self._sbd_watchdog_timeout <= qdevice_sync_timeout:
+                watchdog_timeout_with_qdevice = qdevice_sync_timeout + 5
+                bootstrap.warn("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(watchdog_timeout_with_qdevice, self._sbd_watchdog_timeout))
+                self._sbd_watchdog_timeout = watchdog_timeout_with_qdevice
```

crmsh/sbd.py
liangxin1300: Changed as
```Python
-            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+            if self._sbd_watchdog_timeout < self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE:
+                bootstrap.warn("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE, self._sbd_watchdog_timeout))
+                self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
```

crmsh/sbd.py
```diff
@@ -174,33 +208,35 @@ def _update_configuration(self):
...
         if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
             qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
-            self._sbd_watchdog_timeout = qdevice_sync_timeout + 5
-            if self._is_s390 and self._sbd_watchdog_timeout < 15:
-                self._sbd_watchdog_timeout = 15
+            if self._sbd_watchdog_timeout <= qdevice_sync_timeout:
+                watchdog_timeout_with_qdevice = qdevice_sync_timeout + 5
+                bootstrap.warn("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(watchdog_timeout_with_qdevice, self._sbd_watchdog_timeout))
+                self._sbd_watchdog_timeout = watchdog_timeout_with_qdevice
+            self._adjust_sbd_watchdog_timeout_for_s390()
```
zzhou1, line 224: Improvment to simplify code:

1. No need to call _adjust_sbd_watchdog_timeout_for_s390() here. The reason is crmsh hard-coded s390 default is initialized in __init__(), also the above function is called when load profiles.yml in _initialize_sbd(). No other place could have _sbd_watchdog_timeout smaller than the s390 default.

2. Furthermore, this function has duplicate code and can be simplified further

liangxin1300: OK, remove it from here

crmsh/sbd.py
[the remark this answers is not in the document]
liangxin1300: Changed as
```
        Initialize SBD parameters according to profiles.yml, or the crmsh defined defaulst as the last resort.
        This covers both disk-based-sbd, and diskless-sbd scenarios.
        For diskless-sbd, set _sbd_watchdog_timeout then return;
        For disk-based-sbd, also calculate the msgwait value, then initialize the SBD device.

```

crmsh/sbd.py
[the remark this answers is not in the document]
liangxin1300: OK, dropped "s390" part

zzhou1 (approved):
Great! Now, it covers diskless-sbd and qdevice too !

crmsh/sbd.py
```diff
@@ -241,6 +308,30 @@ def _enable_sbd_service(self):
...
+
+        # TimeoutStartUSec default is 1min 30s, need to parse as seconds
+        cmd = "systemctl show -p TimeoutStartUSec sbd --value"
+        out = utils.get_stdout_or_raise_error(cmd)
+        res_seconds = re.search("(\d+)s", out)
+        default_start_timeout = int(res_seconds.group(1)) if res_seconds else 0
+        res_min = re.search("(\d+)min", out)
+        default_start_timeout += 60 * int(res_min.group(1)) if res_min else 0
+        sbd_watchdog_timeout = SBDManager.get_sbd_watchdog_timeout()
+        if default_start_timeout >= (sbd_watchdog_timeout * 2):
```
gao-yan, line 326: Technically we should check it against `msgwait` for the case with disk.

liangxin1300: Changed to compare `if default_start_timeout >= self.get_sbd_start_timeout_threshold():`
That will make code more reusable

Thanks!

crmsh/sbd.py
```diff
@@ -174,34 +208,27 @@ def _update_configuration(self):
...
+                watchdog_timeout_with_qdevice = qdevice_sync_timeout + 5
+                bootstrap.warn("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(watchdog_timeout_with_qdevice, self._sbd_watchdog_timeout))
+                self._sbd_watchdog_timeout = watchdog_timeout_with_qdevice
             self._stonith_timeout = self.calculate_stonith_timeout(self._sbd_watchdog_timeout)
         # add sbd and qdevice together from beginning
         elif self._context.qdevice_inst:
-            self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+            if self._sbd_watchdog_timeout < self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE:
+                bootstrap.warn("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE, self._sbd_watchdog_timeout))
+                self._sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
```
gao-yan, line 229: It occurred to me,  in case that `corosync.quorum.device.timeout` could be configured in profiles, of course it should check against that rather than the default.  But no urgency. Feel free to do the improvement now or some time in the future.

crmsh/sbd.py
```diff
@@ -212,6 +239,44 @@ def _get_sbd_device_from_config(self):
         else:
             return None
 
+    @staticmethod
+    def is_delay_start():
+        """
+        Check if SBD_DELAY_START is yes
+        """
+        res = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        return True if res and res == "yes" else False
```
gao-yan, line 248: Technically sbd determines this based on `crm_is_true()` function and whether it's explicitly set with a delay value:

https://github.com/ClusterLabs/sbd/blob/master/src/sbd-inquisitor.c#L989-L999

https://github.com/ClusterLabs/pacemaker/blob/master/lib/common/strings.c#L414-L444

Feel free to decide whether we need to make it cover all the cases of `True` here :-)

gao-yan (approved):
Looks much neater. Thanks for the awesome work, @liangxin1300 !

Only a couple of nitpickings here.

crmsh/sbd.py
liangxin1300: Any suggest value for this timeout on different platform? If not clear yet, we could add it in the future

crmsh/sbd.py
liangxin1300: Changed as using `return utils.is_boolean_true(res)`
Thanks!

Issue #843, opened 2021-07-07
bootstrap ocfs2 option and stage should apply `no-quorum-policy=freeze`


zzhou1:
Besides the title, it makes sense to give a warning message to the end user for such change as well. This is to align with the documentatin of SUSE Linux Enterprise High Availability Extenstion in regarding to no-quorum-policy [1]

[1] https://documentation.suse.com/sle-ha/15-SP3/html/SLE-HA-all/cha-ha-config-basics.html#sec-ha-config-basics-global-quorum

PR #846, opened 2021-07-13
Dev: ocfs2: set no-quorum-policy as freeze when configuring OCFS2

liangxin1300:
Implement #843

### DIFF
--- a/crmsh/ocfs2.py
+++ b/crmsh/ocfs2.py
@@ -299,6 +299,11 @@ e.g. crm cluster init ocfs2 -o <ocfs2_device>
             self._config_resource_stack_ocfs2_along()
         bootstrap.status("  OCFS2 device {} mounted on {}".format(self.target_device, self.mount_point))
 
+        res = utils.get_stdout_or_raise_error("crm configure get_property no-quorum-policy")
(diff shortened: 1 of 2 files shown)


crmsh/ocfs2.py
```diff
@@ -299,6 +299,10 @@ def init_ocfs2(self):
             self._config_resource_stack_ocfs2_along()
         bootstrap.status("  OCFS2 device {} mounted on {}".format(self.target_device, self.mount_point))
 
+        res = utils.get_stdout_or_raise_error("crm configure get_property no-quorum-policy")
+        if res != "freeze":
+            utils.get_stdout_or_raise_error("crm configure property no-quorum-policy=freeze")
+
```
liangxin1300, line 305: @zzhou1 
> This setting is recommended for clusters where certain resources depend on communication with other nodes

What if these resources been deleted? We cannot change no-quorum-policy as stop back automatically

zzhou1: I think it is fine. But, it worth to print a explict message to user the default value is changed to "freeze" here.

zzhou1 (approved):

Issue #459, opened 2019-07-24
slow load for configure


dmuhamedagic:
Every crm cf invocation takes almost two seconds:

```
root@alpaca1:~# time crm cf  <> /dev/null

real	0m1.709s
user	0m0.310s
sys	0m0.175s
```

It turns out that most of the time is used by `crm_resource --list-standards` (for whatever reason is that one taking so long, but that's another matter).
This is cached for the completion, but the completion is not used for one-shot commands such as `crm cf show blah`.

dmuhamedagic:
No comments? Don't you find this an important usability issue?

dmuhamedagic:
As for `crm_resource --list-standards`, the culprit is actually systemd and I see the same behaviour with `systemctl --list-unit-files`. It makes around 300k filesystem related system calls (for instance, readlinkat was invoked 119684 times). This is on a Debian Buster.

gao-yan:
Hi Dejan,

It's definitely an usability issue if it's so slow. I gave it a quick check with my different releases of SUSE Linux. It looks much better though.

`crm_resource --list-standards` takes around 0.06s for SLE12SP4,  0.25s for opensuse 15.1.

`time crm cf  <> /dev/null` takes around 0.2s for SLE12 SP4, 0.45s for opensuse 15.1.

`systemctl list-unit-files >/dev/null` takes around 0.05s for SLE12 SP4,  0.25s for opensuse 15.1.

A more similar command `crm_resource --list-agents systemd` takes  0.06s for SLE12 SP4,  0.23s for opensuse15.1.

The performance is acceptable AFAICT, although there are kind of big differences between the releases. They are both VMs BTW.  I assume the results could quite depend on different systems, environments, and/or probably versions of systemd ...  --  Debian Buster seems like a very new release. Not sure if there's performance regression related to systemd though...

dmuhamedagic:
Hey Yan,
Adding --state=enabled to the list-unit-files command makes it run twice as long (3.2s on an Intel Atom). That's really dreadful. 
With Debian Stretch (systemd 232) I get similar results as you with SLE12SP4 (0.05s), so there's definitely a regression between that version and systemd 241 of Debian Buster. I also have one system running a version in between (237) and its performance is also in between (~0.5s). Opensuse 15.1 runs version 234 shows a 5x regression in terms of performance. So, apparently the regression is cumulative :-D
[...]

PR #851, opened 2021-07-26
Medium: ra: performance/usability improvement (avoid systemd)

dmuhamedagic:
Use --list-ocf-providers to check the crm_resource compatibility.
The --list-standards invokes systemd which takes quite some time.
Unfortunately, it doesn't seem like that will be fixed and, at any rate,
it's better to depend less on other system components.

This is no change in functionality, but a major improvement in
user experience. See #459 for more information.

### DIFF
--- a/crmsh/ra.py
+++ b/crmsh/ra.py
@@ -65,7 +65,7 @@ def can_use_lrmadmin():
 
 @utils.memoize
 def can_use_crm_resource():
-    _rc, s = get_stdout("crm_resource --list-standards", stderr_on=False)
+    _rc, s = get_stdout("crm_resource --list-ocf-providers", stderr_on=False)
     return s != ""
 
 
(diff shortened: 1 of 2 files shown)


liangxin1300:
@dmuhamedagic Thanks for this commit, nice work!

BTW, you might need to append another commit to pass the `crmsh CI / original_regression_test`

Thank you!

dmuhamedagic:
Ah, yes, will take a look at the regression tests, thanks for reminding me.
The test which failed now is unrelated:
docker.io/shap/continuous_deliver:latest
OBS_USER or OBS_PASS not set...
Error: Process completed with exit code 1.

dmuhamedagic:
There is no output change with this commit, so the regression tests don't need modifications.

dmuhamedagic:
The regression tests use -R :(

liangxin1300 (approved):
Thank you @dmuhamedagic !

PR #855, opened 2021-08-04
Fix: doc: Note that resource tracing is only supported by OCF RAs

liangxin1300:


doc/crm.8.adoc
```diff
@@ -2006,6 +2006,10 @@ To trace the probe operation which exists for all resources, either
 set a trace for `monitor` with interval `0`, or use `probe` as the
 operation name.
 
+Note: RA tracing is only supported by OCF resource agent;
```
gao-yan, line 2009: s/agent/agents/

gao-yan (approved):
Thanks. Only a nitpicking.

zzhou1 (approved):

PR #859, opened 2021-08-06
Dev: ui_resource: Add warning if detect maintenance-mode is true while start/stop/add RA

liangxin1300:


zzhou1 (changes requested):
The code is incomplete since the `maintenance` mode has 3 different levels: cluster/node/resource.

crmsh/xmlutil.py
```diff
@@ -243,28 +303,33 @@ def rsc_clone(self, ident):
...
+        if NodeState().are_all_nodes_in_maintenance():
+            return False, "all nodes are in maintenance"
+        # then check if node running this resource is in maintenance
+        if NodeState().is_node_running_resource_in_maintenance(ident):
+            return False, "node which running \"{}\" is in maintenance".format(ident)
         rsc_meta_node = get_rsc_meta_node(rsc_node)
+        # then check the rsc maintenance meta attribute
+        attr = get_attr_value(rsc_meta_node, "maintenance")
+        if attr and is_xs_boolean_true(attr):
+            return False, "resource \"{}\" is in maintenance".format(ident)
```
zzhou1, line 323: The above "maintenance" checks are not the scope of is_managed(). Since, "maintenance" and "is-managed" are two different terminologies and not fit into the current function name itself all together. I'm thinking their function names something like,

is_managed(rsc)
is_maintenance(rsc)
is_managed_or_maintenance(rsc)

crmsh/ui_resource.py
```diff
@@ -293,6 +293,9 @@ def _commit_meta_attrs(self, context, resources, name, value):
 
         rc = True
         for rsc in resources:
+            rc, reason = xmlutil.RscState().is_managed(rsc)
+            if not rc:
+                context.error("Resource {} is unmanaged ({})".format(rsc, reason))
```
zzhou1, line 298: Confusing terminology here, since "unmanaged" does not match to the fact when the "maintenance" mode is true.

zzhou1 (changes requested):
please tweak the title, something like,

Dev: ui_resource: reject start/stop/restart if "is-managed" or "maintenance" get detected

crmsh/ui_resource.py
liangxin1300: @zzhou1 For the display of `crm_mon`, both maintenance and is-manage=false situation, the resource status always show `unmanaged`

crmsh/xmlutil.py
```diff
@@ -173,7 +173,71 @@ def get_top_cib_nodes(node, nodes_l):
...
+        attr = get_attr_value(attr_entry, "maintenance")
+        return is_xs_boolean_true(attr) if attr else False
+
+    def are_all_nodes_in_maintenance(self):
+        """
+        Check if all nodes are in maintenance
+        """
+        return all([self.is_node_in_maintenance(node) for node in self.list_nodes()])
+
+    def is_node_running_resource_in_maintenance(self, rsc_id):
```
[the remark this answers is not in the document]
liangxin1300: Changed

crmsh/xmlutil.py
liangxin1300: From display of `crm_mon`, pacemaker show both situations as `unmanaged` for the resource
How about change the function name as `is_managed_or_maintenance`?

PR #875, opened 2021-09-18
Dev: log: Rotate crmsh.log as 1M and backup count as 10

liangxin1300:
Use RotatingFileHandler class to rotated crmsh.log, archived when reach 1M size, keep latest 10 archived log

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -121,10 +121,12 @@ LOGGING_CFG = {
             "flushLevel": logging.CRITICAL,
         },
         "file": {
-            "class": "logging.FileHandler",
+            "class": "logging.handlers.RotatingFileHandler",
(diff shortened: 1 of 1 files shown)


zzhou1 (commented):
Does it make sense to have them configurable in crm.conf?

If the log file size is never big enough to bring the attention to the end users. Then, the current hard-code is fine. Anyway, this need the measurement in the practice to make the decision accordingly.

liangxin1300:
> Does it make sense to have them configurable in crm.conf?
> 
> If the log file size is never big enough to bring the attention to the end users. Then, the current hard-code is fine. Anyway, this need the measurement in the practice to make the decision accordingly.

Consider the log from bootstrap and normal operations command line, the 1M size is big enough

PR #876, opened 2021-09-26
Dev: sbd: Adjust pcmk_delay_max for stonith:external/sbd agent

liangxin1300:
Enable this when two-nodes cluster
Disble this when non-two-nodes cluster

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -34,6 +34,9 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
     SBD_WATCHDOG_TIMEOUT_DEFAULT = 5
     SBD_WATCHDOG_TIMEOUT_DEFAULT_S390 = 15
     SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE = 35
+    SBD_RA_TYPE = "stonith:external/sbd"
+    PCMK_DELAY_MAX = "pcmk_delay_max"
+    PCMK_DELAY_MAX_VALUE = 30
 
     def __init__(self, context):
         """
@@ -378,11 +381,12 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         """
         if not utils.package_is_installed("sbd") or \
                 not utils.service_is_enabled("sbd.service") or \
-                utils.has_resource_configured("stonith:external/sbd"):
+                utils.has_resource_configured(self.SBD_RA_TYPE):
             return
 
         if self._get_sbd_device_from_config():
-            if not bootstrap.invokerc("crm configure primitive stonith-sbd stonith:external/sbd pcmk_delay_max=30s"):
+            pcmk_delay_max_str = " {}={}s".format(self.PCMK_DELAY_MAX, self.PCMK_DELAY_MAX_VALUE) if utils.is_two_nodes_cluster() else ""
+            if not bootstrap.invokerc("crm configure primitive stonith-sbd {}{}".format(self.SBD_RA_TYPE, pcmk_delay_max_str)):
                 utils.fatal("Can't create stonith-sbd primitive")
             if not bootstrap.invokerc("crm configure property stonith-enabled=true"):
                 utils.fatal("Can't enable STONITH for SBD")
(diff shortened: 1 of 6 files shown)


crmsh/utils.py
```diff
@@ -2929,4 +2946,11 @@ def fatal(error_msg):
     handled by Context.run in ui_context.py
     """
     raise ValueError(error_msg)
+
+
+def is_two_nodes_cluster():
+    """
+    Check if current cluster has two nodes
+    """
+    return len(list_cluster_nodes()) == 2
```
zzhou1, line 2955: Please rename 'is_two_nodes_cluster()' to something like `is_2node_cluster_without_qdevice()`, and also add the qdevice check accordingly.

zzhou1 (changes requested):

crmsh/utils.py
liangxin1300: So that means if there are 2 nodes cluster + qdevice running, it's the safe case which will not have the fencing race issue, and don't need `pcmk_delay_max`?

crmsh/sbd.py
```diff
@@ -378,11 +381,12 @@ def configure_sbd_resource(self):
...
         if not utils.package_is_installed("sbd") or \
                 not utils.service_is_enabled("sbd.service") or \
-                utils.has_resource_configured("stonith:external/sbd"):
+                utils.has_resource_configured(self.SBD_RA_TYPE):
             return
 
         if self._get_sbd_device_from_config():
-            if not bootstrap.invokerc("crm configure primitive stonith-sbd stonith:external/sbd pcmk_delay_max=30s"):
+            pcmk_delay_max_str = " {}={}s".format(self.PCMK_DELAY_MAX, self.PCMK_DELAY_MAX_VALUE) if utils.is_two_nodes_cluster() else ""
+            if not bootstrap.invokerc("crm configure primitive stonith-sbd {}{}".format(self.SBD_RA_TYPE, pcmk_delay_max_str)):
```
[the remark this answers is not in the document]
zzhou1: I do prefer to let adjust_pcmk_delay_max() be idempotent, then it can be called any time and do the adjustment accordingly, not matter stage, join, or remove. 

For 3->2, sorry, I didn't describe completely. My brain extended to the rare use case when users manually configure pcmk_delay_max on their own before 'remove' node. I agree this is the rare situation.

crmsh/sbd.py
```diff
@@ -469,3 +473,19 @@ def get_sbd_value_from_config(key):
...
+    def adjust_pcmk_delay_max(self):
+        """
+        """
+        if not utils.has_resource_configured(self.SBD_RA_TYPE):
+            return
+        cmd = None
+        sbd_ra_id = utils.get_primitive_id(self.SBD_RA_TYPE)
+        has_pcmk_delay_max = utils.has_parameter_configured(sbd_ra_id, self.PCMK_DELAY_MAX)
+        if utils.is_two_nodes_cluster():
+            if not has_pcmk_delay_max:
```
zzhou1, line 486: It makes sense to me to remove this line above, " if not has_pcmk_delay_max: "

crmsh/utils.py
zzhou1: You are right.

crmsh/sbd.py
```diff
@@ -469,3 +473,19 @@ def get_sbd_value_from_config(key):
         conf = utils.parse_sysconfig(SYSCONFIG_SBD)
         res = conf.get(key)
         return res
+
+    def adjust_pcmk_delay_max(self):
```
zzhou1, line 477: Currently, adjust_pcmk_delay_max() only deal with disk-based sbd and diskless sbd. 

In theory it applies to other non-sbd stonith device types too, for 2-node cluster without qdevice. for example. 

crm configure show type:primitive | grep -e stonith:

And, adjust_pcmk_delay_max() can be called anywhere, even outside of bootstrap. With that, I prefer to move this function to cibconfig.py, 

Moreover, there are more complex situations with multiple stonith devices in the system. Maybe we only start with  the scope for the cluster with one stonith device, for the time being. And, this is good to ask Yan's advice too.

zzhou1 (changes requested):

Issue #879, opened 2021-10-13
`crm cluster health` uses platform.linux_distribution, dropped in python 3.8


athos-ribeiro:
`scripts/health/collect.py` relies on a call to `platform.linux_distribution`, which has been deprecated since Python 3.5 and was removed on python 3.8 [1].

This results in errors for `crm cluster health` calls on systems with python >= 3.8.

Until Python 3.7, when `platform.linux_distribution` was still present and deprecated, the Python docs suggested using distro [3] as an alternative, which could be a solution for this issue. However, distro itself requires python >= 3.6, which could imply losing support for python 3.4 and 3.5 or implementing a fallback logic to use `platform.linux_distribution` when available.

[1] https://bugs.python.org/issue28167
[2] https://docs.python.org/3.7/library/platform.html#platform.linux_distribution
[3] https://pypi.org/project/distro/

liangxin1300:
@athos-ribeiro Thanks for pointing this!
To be honest, the content in `scripts/health/collect.py` seems duplicated a little, comparing with codes in `crmsh/report` 
So here I want to reuse the function `get_distro_info` from `crmsh/report/utillib.py`, see https://github.com/ClusterLabs/crmsh/pull/883/commits/c90a4e6e871a941ef65d9735852b3889d10ba51e

PR #883, opened 2021-11-01
Replace/rename hb_report as crm report

liangxin1300:
#### Dev: crm_report: Move hb_report directory to crmsh/report
- Replace `hb_report` word as `crm report`
- Move hb_report directory into crmsh/report
- Rename hb_report/hb_report.in as crmsh/report/core.py
- Change default report result name starts with `crm_report`
- Remove crmsh/ui_report.py
#### Dev: crm_report: Integrate report log into crmsh logging
- To keep consistent logging style and reuse/inherite defined logging in crmsh/log.py
#### Dev: crm report: Get distribution info correctly and reuse it
- #879


crmsh/config.py
```diff
@@ -294,7 +294,8 @@ def get(self, value):
         'collect_extra_logs': opt_string('/var/log/messages /var/log/pacemaker/pacemaker.log /var/log/pacemaker.log /var/log/crmsh/crmsh.log'),
```
zzhou1, line 294: Would be useful to collect `crm.conf` and `profiles.yml` from all nodes too

liangxin1300: Added

zzhou1 (approved):

PR #890, opened 2021-11-23
Dev: sbd: Adjust timeout related values

liangxin1300:
## Changes
- Consolidate sbd timeout related methods/constants/formulas into class SBDTimeout
    
- Adjust stonith-timeout value, formulas are:
```
  stonith-timeout >= 1.2 * (pcmk_delay_max + msgwait)  # for disk-based sbd
  stonith-timeout >= 1.2 * max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)  # for disk-less sbd
  stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for ALL situations
```
 
- Adjust SBD_DELAY_START value, formulas are:
```
  SBD_DELAY_START = no # for non virtualization environment, which is the system default
  SBD_DELAY_START >= (token + consensus + pcmk_delay_max + msgwait)  # for disk-based sbd
  SBD_DELAY_START >= (token + consensus + 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
``` 
- Adjust pcmk_delay_max:
```
  pcmk_delay_max=30 # only for the single stonith device in the 2-node cluster without qdevice
  pcmk_delay_max deletion # only for the single stonith device, not in the 2-node cluster without qdevice
```
- Dev: behave: Add functional test for previous changes and cases

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -33,7 +33,7 @@ from . import corosync
 from . import tmpfiles
 from . import lock
 from . import userdir
-from .constants import SSH_OPTION, QDEVICE_HELP_INFO, CRM_MON_ONE_SHOT
+from .constants import SSH_OPTION, QDEVICE_HELP_INFO, CRM_MON_ONE_SHOT, STONITH_TIMEOUT_DEFAULT
 from . import ocfs2
 from . import qdevice
 from . import log
@@ -63,12 +63,13 @@ WATCHDOG_CFG = "/etc/modules-load.d/watchdog.conf"
 BOOTH_DIR = "/etc/booth"
 BOOTH_CFG = "/etc/booth/booth.conf"
 BOOTH_AUTH = "/etc/booth/authkey"
+SBD_SYSTEMD_DELAY_START_DIR = "/etc/systemd/system/sbd.service.d"
 FILES_TO_SYNC = (BOOTH_DIR, corosync.conf(), COROSYNC_AUTH, CSYNC2_CFG, CSYNC2_KEY, "/etc/ctdb/nodes",
         "/etc/drbd.conf", "/etc/drbd.d", "/etc/ha.d/ldirectord.cf", "/etc/lvm/lvm.conf", "/etc/multipath.conf",
         "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK, SYSCONFIG_SBD, PCMK_REMOTE_AUTH, WATCHDOG_CFG,
-        PROFILES_FILE, CRM_CFG)
-
+        PROFILES_FILE, CRM_CFG, SBD_SYSTEMD_DELAY_START_DIR)
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "sbd", "cluster", "ocfs2", "admin", "qdevice")
+
 
 class QdevicePolicy(Enum):
     QDEVICE_RELOAD = 0
@@ -698,12 +699,14 @@ def start_pacemaker(node_list=[]):
     Start pacemaker service with wait time for sbd
     When node_list set, start pacemaker service in parallel
     """
-    from .sbd import SBDManager
+    from .sbd import SBDTimeout
     pacemaker_start_msg = "Starting pacemaker"
-    if utils.package_is_installed("sbd") and \
+    # not _context means not in init or join process
+    if not _context and \
+            utils.package_is_installed("sbd") and \
             utils.service_is_enabled("sbd.service") and \
-            SBDManager.is_delay_start():
-        pacemaker_start_msg += "(waiting for sbd {}s)".format(SBDManager.get_suitable_sbd_systemd_timeout())
+            SBDTimeout.is_sbd_delay_start():
+        pacemaker_start_msg += "(delaying start of sbd for {}s)".format(SBDTimeout.get_sbd_delay_start_sec_from_sysconfig())
     with logger_utils.status_long(pacemaker_start_msg):
         utils.start_service("pacemaker.service", enable=True, node_list=node_list)
 
@@ -1237,7 +1240,7 @@ op_defaults op-options: timeout=600 record-pending=true
 rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
 
-    _context.sbd_manager.configure_sbd_resource()
+    _context.sbd_manager.configure_sbd_resource_and_properties()
 
 
 def init_admin():
@@ -1334,20 +1337,17 @@ def init_qdevice():
         utils.disable_service("corosync-qdevice.service")
         return
     if _context.stage == "qdevice":
-        from .sbd import SBDManager
+        from .sbd import SBDManager, SBDTimeout
         utils.check_all_nodes_reachable()
         using_diskless_sbd = SBDManager.is_using_diskless_sbd()
         _context.qdevice_reload_policy = evaluate_qdevice_quorum_effect(QDEVICE_ADD, using_diskless_sbd)
         # add qdevice after diskless sbd started
         if using_diskless_sbd:
             res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
-            if res:
-                sbd_watchdog_timeout = max(int(res), SBDManager.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE)
-            else:
-                sbd_watchdog_timeout = SBDManager.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
-            stonith_timeout = SBDManager.calculate_stonith_timeout(sbd_watchdog_timeout)
-            SBDManager.update_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout)})
-            invokerc("crm configure property stonith-watchdog-timeout=-1 stonith-timeout={}s".format(stonith_timeout))
+            if not res or int(res) < SBDTimeout.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE:
+                sbd_watchdog_timeout_qdevice = SBDTimeout.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+                SBDManager.update_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout_qdevice)})
+                utils.set_property(stonith_timeout=SBDTimeout.get_stonith_timeout())
 
     logger.info("""Configure Qdevice/Qnetd:""")
     qdevice_inst = _context.qdevice_inst
@@ -1822,6 +1822,14 @@ def join_cluster(seed_host):
     # attempt to join the cluster failed)
     init_cluster_local()
 
+    if utils.service_is_active("sbd.service"):
+        from .sbd import SBDTimeout
+        SBDTimeout.adjust_sbd_timeout_related_cluster_configuration()
+    else:
+        value = get_stonith_timeout_generally_expected()
+        if value:
+            utils.set_property_conditionally("stonith-timeout", value)
+
     with logger_utils.status_long("Reloading cluster configuration"):
 
         if ipv6_flag and not is_unicast:
@@ -1927,6 +1935,10 @@ def remove_node_from_cluster():
     """
     Remove node from running cluster and the corosync / pacemaker configuration.
     """
+    if utils.service_is_active("sbd.service"):
+        from .sbd import SBDTimeout
+        SBDTimeout.adjust_sbd_timeout_related_cluster_configuration(removing=True)
+
     node = _context.cluster_node
     set_cluster_node_ip()
 
@@ -2392,4 +2404,17 @@ def bootstrap_arbitrator(context):
     logger.info("Enabling and starting the booth arbitrator service")
     utils.start_service("booth@booth", enable=True)
 
+
+def get_stonith_timeout_generally_expected():
+    """
+    Adjust stonith-timeout for all scenarios, formula is:
+
+    stonith-timeout = STONITH_TIMEOUT_DEFAULT + token + consensus
+    """
+    stonith_enabled = utils.get_property("stonith-enabled")
+    # When stonith disabled, return
+    if utils.is_boolean_false(stonith_enabled):
+        return None
+
+    return STONITH_TIMEOUT_DEFAULT + corosync.token_and_consensus_timeout()
 # EOF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -17,6 +17,9 @@ from . import log
 logger = log.setup_logger(__name__)
 
 
+COROSYNC_TOKEN_DEFAULT = 1000  # in ms units
+
+
 def conf():
     return os.getenv('COROSYNC_MAIN_CONFIG_FILE', '/etc/corosync/corosync.conf')
 
@@ -762,3 +765,39 @@ def create_configuration(clustername="hacluster",
                               _COROSYNC_CONF_TEMPLATE_RING_ALL + \
                               _COROSYNC_CONF_TEMPLATE_TAIL
     utils.str2file(_COROSYNC_CONF_TEMPLATE % config_common, conf())
+
+
+def get_corosync_value(key):
+    """
+    Get corosync configuration value from corosync-cmapctl or corosync.conf
+    """
+    try:
+        out = utils.get_stdout_or_raise_error("corosync-cmapctl {}".format(key))
+        res = re.search(r'{}\s+.*=\s+(.*)'.format(key), out)
+        return res.group(1) if res else None
+    except ValueError:
+        out = get_value(key)
+        return out
+
+
+def get_corosync_value_dict():
+    """
+    Get corosync value, then return these values as dict
+    """
+    value_dict = {}
+
+    token = get_corosync_value("totem.token")
+    value_dict["token"] = int(int(token)/1000) if token else int(COROSYNC_TOKEN_DEFAULT/1000)
+
+    consensus = get_corosync_value("totem.consensus")
+    value_dict["consensus"] = int(int(consensus)/1000) if consensus else int(value_dict["token"]*1.2)
+
+    return value_dict
+
+
+def token_and_consensus_timeout():
+    """
+    Get corosync token plus consensus timeout
+    """
+    _dict = get_corosync_value_dict()
+    return _dict["token"] + _dict["consensus"]
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -3,12 +3,259 @@ import re
 import shutil
 from . import utils
 from . import bootstrap
-from .bootstrap import SYSCONFIG_SBD
+from .bootstrap import SYSCONFIG_SBD, SBD_SYSTEMD_DELAY_START_DIR
 from . import log
+from . import constants
+from . import corosync
 
 
 logger = log.setup_logger(__name__)
 logger_utils = log.LoggerUtils(logger)
+
+
+class SBDTimeout(object):
+    """
+    Consolidate sbd related timeout methods and constants
+    """
+    STONITH_WATCHDOG_TIMEOUT_DEFAULT = -1
+    SBD_WATCHDOG_TIMEOUT_DEFAULT = 5
+    SBD_WATCHDOG_TIMEOUT_DEFAULT_S390 = 15
+    SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE = 35
+    QDEVICE_SYNC_TIMEOUT_MARGIN = 5
+
+    def __init__(self, context=None, removing=False):
+        """
+        Init function
+        """
+        self.context = context
+        self.sbd_msgwait = None
+        self.stonith_timeout = None
+        self.sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT
+        self.stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
+        self.sbd_delay_start = None
+        self.removing = removing
+        self.two_node_without_qdevice = False
+
+    def set_sbd_watchdog_timeout(self):
+        """
+        Set sbd_watchdog_timeout from profiles.yml if exists
+        Then adjust it if in s390 environment
+        """
+        if "sbd.watchdog_timeout" in self.context.profiles_dict:
+            self.sbd_watchdog_timeout = int(self.context.profiles_dict["sbd.watchdog_timeout"])
+        if self.context.is_s390 and self.sbd_watchdog_timeout < self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390:
+            logger.warning("sbd_watchdog_timeout is set to %d for s390, it was %d", self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390, self.sbd_watchdog_timeout)
+            self.sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_S390
+
+    def set_sbd_msgwait(self):
+        """
+        Set sbd msgwait from profiles.yml if exists
+        Default is 2 * sbd_watchdog_timeout
+        """
+        sbd_msgwait_default = 2 * self.sbd_watchdog_timeout
+        sbd_msgwait = sbd_msgwait_default
+        if "sbd.msgwait" in self.context.profiles_dict:
+            sbd_msgwait = int(self.context.profiles_dict["sbd.msgwait"])
+            if sbd_msgwait < sbd_msgwait_default:
+                logger.warning("sbd msgwait is set to %d, it was %d", sbd_msgwait_default, sbd_msgwait)
+                sbd_msgwait = sbd_msgwait_default
+        self.sbd_msgwait = sbd_msgwait
+
+    def adjust_sbd_watchdog_timeout_with_diskless_and_qdevice(self):
+        """
+        When using diskless SBD with Qdevice, adjust value of sbd_watchdog_timeout
+        """
+        # add sbd after qdevice started
+        if utils.is_qdevice_configured() and utils.service_is_active("corosync-qdevice.service"):
+            qdevice_sync_timeout = utils.get_qdevice_sync_timeout()
+            if self.sbd_watchdog_timeout <= qdevice_sync_timeout:
+                watchdog_timeout_with_qdevice = qdevice_sync_timeout + self.QDEVICE_SYNC_TIMEOUT_MARGIN
+                logger.warning("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(watchdog_timeout_with_qdevice, self.sbd_watchdog_timeout))
+                self.sbd_watchdog_timeout = watchdog_timeout_with_qdevice
+        # add sbd and qdevice together from beginning
+        elif self.context.qdevice_inst:
+            if self.sbd_watchdog_timeout < self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE:
+                logger.warning("sbd_watchdog_timeout is set to {} for qdevice, it was {}".format(self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE, self.sbd_watchdog_timeout))
+                self.sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
+
+    @staticmethod
+    def get_sbd_msgwait(dev):
+        """
+        Get msgwait for sbd device
+        """
+        out = utils.get_stdout_or_raise_error("sbd -d {} dump".format(dev))
+        # Format like "Timeout (msgwait)  : 30"
+        res = re.search("\(msgwait\)\s+:\s+(\d+)", out)
+        if not res:
+            raise ValueError("Cannot get sbd msgwait for {}".format(dev))
+        return int(res.group(1))
+
+    @staticmethod
+    def get_sbd_watchdog_timeout():
+        """
+        Get SBD_WATCHDOG_TIMEOUT from /etc/sysconfig/sbd
+        """
+        res = SBDManager.get_sbd_value_from_config("SBD_WATCHDOG_TIMEOUT")
+        if not res:
+            raise ValueError("Cannot get the value of SBD_WATCHDOG_TIMEOUT")
+        return int(res)
+
+    @staticmethod
+    def get_stonith_watchdog_timeout():
+        """
+        For non-bootstrap case, get stonith-watchdog-timeout value from cluster property
+        """
+        default = SBDTimeout.STONITH_WATCHDOG_TIMEOUT_DEFAULT
+        if not utils.service_is_active("pacemaker.service"):
+            return default
+        value = utils.get_property("stonith-watchdog-timeout")
+        return int(value.strip('s')) if value else default
+
+    def _load_configurations(self):
+        """
+        Load necessary configurations for both disk-based/disk-less sbd
+        """
+        self.two_node_without_qdevice = utils.is_2node_cluster_without_qdevice(self.removing)
+
+        dev_list = SBDManager.get_sbd_device_from_config()
+        if dev_list:  # disk-based
+            self.disk_based = True
+            self.msgwait = SBDTimeout.get_sbd_msgwait(dev_list[0])
+            self.pcmk_delay_max = utils.get_pcmk_delay_max(self.two_node_without_qdevice)
+        else:  # disk-less
+            self.disk_based = False
+            self.sbd_watchdog_timeout = SBDTimeout.get_sbd_watchdog_timeout()
+            self.stonith_watchdog_timeout = SBDTimeout.get_stonith_watchdog_timeout()
+        self.sbd_delay_start_value_expected = self.get_sbd_delay_start_expected() if utils.detect_virt() else "no"
+        self.sbd_delay_start_value_from_config = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+
+        logger.debug("Inspect SBDTimeout: %s", vars(self))
+
+    def get_stonith_timeout_expected(self):
+        """
+        Get stonith-timeout value for sbd cases, formulas are:
+
+        value_from_sbd = 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
+        value_from_sbd = 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
+
+        stonith_timeout = max(value_from_sbd, constants.STONITH_TIMEOUT_DEFAULT) + token + consensus
+        """
+        if self.disk_based:
+            value_from_sbd = int(1.2*(self.pcmk_delay_max + self.msgwait))
+        else:
+            value_from_sbd = int(1.2*max(self.stonith_watchdog_timeout, 2*self.sbd_watchdog_timeout))
+
+        value = max(value_from_sbd, constants.STONITH_TIMEOUT_DEFAULT) + corosync.token_and_consensus_timeout()
+        logger.debug("Result of SBDTimeout.get_stonith_timeout_expected %d", value)
+        return value
+
+    @classmethod
+    def get_stonith_timeout(cls):
+        cls_inst = cls()
+        cls_inst._load_configurations()
+        return cls_inst.get_stonith_timeout_expected()
+
+    def get_sbd_delay_start_expected(self):
+        """
+        Get the value for SBD_DELAY_START, formulas are:
+
+        SBD_DELAY_START = (token + consensus + pcmk_delay_max + msgwait)  # for disk-based sbd
+        SBD_DELAY_START = (token + consensus + 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
+        """
+        token_and_consensus_timeout = corosync.token_and_consensus_timeout()
+        if self.disk_based:
+            value = token_and_consensus_timeout + self.pcmk_delay_max + self.msgwait
+        else:
+            value = token_and_consensus_timeout + 2*self.sbd_watchdog_timeout
+        return value
+
+    @staticmethod
+    def get_sbd_delay_start_sec_from_sysconfig():
+        """
+        Get suitable systemd start timeout for sbd.service
+        """
+        # TODO 5ms, 5us, 5s, 5m, 5h are also valid for sbd sysconfig
+        value = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        if utils.is_boolean_true(value):
+            return 2*SBDTimeout.get_sbd_watchdog_timeout()
+        return int(value)
+
+    @staticmethod
+    def is_sbd_delay_start():
+        """
+        Check if SBD_DELAY_START is not no or not set
+        """
+        res = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        return res and res != "no"
+
+    def adjust_systemd_start_timeout(self):
+        """
+        Adjust start timeout for sbd when set SBD_DELAY_START
+        """
+        sbd_delay_start_value = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        if sbd_delay_start_value == "no":
+            return
+
+        cmd = "systemctl show -p TimeoutStartUSec sbd --value"
+        out = utils.get_stdout_or_raise_error(cmd)
+        start_timeout = utils.get_systemd_timeout_start_in_sec(out)
+        if start_timeout > int(sbd_delay_start_value):
+            return
+
+        utils.mkdirp(SBD_SYSTEMD_DELAY_START_DIR)
+        sbd_delay_start_file = "{}/sbd_delay_start.conf".format(SBD_SYSTEMD_DELAY_START_DIR)
+        utils.str2file("[Service]\nTimeoutSec={}".format(int(1.2*int(sbd_delay_start_value))), sbd_delay_start_file)
+        bootstrap.csync2_update(SBD_SYSTEMD_DELAY_START_DIR)
+        utils.cluster_run_cmd("systemctl daemon-reload")
+
+    def adjust_stonith_timeout(self):
+        """
+        Adjust stonith-timeout property
+        """
+        utils.set_property_conditionally("stonith-timeout", self.get_stonith_timeout_expected())
+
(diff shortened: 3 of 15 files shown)


zzhou1:
> stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations

It is not "for other situations". It is "for ALL situations". eg. there could be (token+consensus) > (pcmk_delay_max + msgwait) 

Also enforce ">=" to leave the room for any manual change outside of crmsh. That says, don't use "=" to set the value directly, that likely override some intentional configuration by sysadmin.

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+
+        value_from_configuration = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        if re.search(r'\d+', value_from_configuration):
+            adjust_value = max(int(value_from_configuration), value_from_calculation)
+        else:
+            adjust_value = value_from_calculation
+        return adjust_value
+
+    @staticmethod
+    def get_suitable_sbd_systemd_timeout():
```
zzhou1, line 175: This function is only called in start_pacemaker() to prepare pacemaker_start_msg. With that, it makes sense to me 
1. to rename to something like, get_sbd_delay_start_sec_from_sysconfig()
2. no need (1.2* )
3. to convert either 1 or yes to 2*SBD_WATCHDOG_TIMEOUT
4. to take the exception of int() to 0, eg. for "no/NO/false/False/xxx"

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+        bootstrap.csync2_update(SBD_SYSTEMD_DELAY_START_DIR)
+        utils.cluster_run_cmd("systemctl daemon-reload")
+
+    def adjust_stonith_timeout(self):
+        """
+        Adjust stonith-timeout property
+        """
+        utils.set_property(stonith_timeout=self.get_stonith_timeout())
+
+    def adjust_pcmk_delay_max(self):
```
zzhou1, line 221: TODO is needed, maybe in the future, and outside of sbd.py. This function applies to any fence device, not just sbd only.

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+
+    def adjust_delay_start(self):
+        """
+        Adjust SBD_DELAY_START in /etc/sysconfig/sbd
+        """
+        value = str(self.sbd_delay_start_value) if self.two_node_without_qdevice else "no"
+        SBDManager.update_configuration({"SBD_DELAY_START": value})
+
+    @classmethod
+    def adjust_sbd_timeout_related(cls, removing=False):
```
zzhou1, line 242: Maybe more clear by the little detailed name, say, adjust_sbd_timeout_related_cluster_configuration()

zzhou1 (changes requested):

crmsh/corosync.py
```diff
@@ -762,3 +765,39 @@ def create_configuration(clustername="hacluster",
...
+    token = get_corosync_value("totem.token")
+    value_dict["token"] = int(int(token)/1000) if token else int(COROSYNC_TOKEN_DEFAULT/1000)
+
+    consensus = get_corosync_value("totem.consensus")
+    value_dict["consensus"] = int(int(consensus)/1000) if consensus else int(value_dict["token"]*1.2)
+
+    return value_dict
+
+
+def token_consensus_margin():
```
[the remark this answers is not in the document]
liangxin1300: Changed as `token_and_consensus_timeout`

crmsh/corosync.py
```diff
@@ -17,6 +17,9 @@
 logger = log.setup_logger(__name__)
 
 
+COROSYNC_TOKEN_DEFAULT = 1000  # in ms units
```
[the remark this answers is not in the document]
liangxin1300: If user use whatever suse's tools to setup cluster, the value will be 5000ms, but what if user write corosync.conf manually without token?

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+        self.context = context
+        self.sbd_msgwait = None
+        self.stonith_timeout = None
+        self.sbd_watchdog_timeout = self.SBD_WATCHDOG_TIMEOUT_DEFAULT
+        self.stonith_watchdog_timeout = self.STONITH_WATCHDOG_TIMEOUT_DEFAULT
+        self.sbd_delay_start = None
+        self.removing = removing
+        self.two_node_without_qdevice = False
+
+    def set_sbd_watchdog_timeout(self):
```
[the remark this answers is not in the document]
liangxin1300: I prefer not to change the name, since the sbd_watchdog_timeout be set firstly when init sbd, not adjust value after set, and loading from profiles is just part of logic, not all logics

crmsh/bootstrap.py
```diff
@@ -2392,4 +2399,19 @@ def bootstrap_arbitrator(context):
     logger.info("Enabling and starting the booth arbitrator service")
     utils.start_service("booth@booth", enable=True)
 
+
+def adjust_stonith_timeout_for_normal():
+    """
+    Adjust stonith-timeout for non-sbd scenarios, formula is:
+
+    stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)
```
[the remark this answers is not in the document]
liangxin1300: Done

crmsh/bootstrap.py
```diff
@@ -2392,4 +2399,19 @@ def bootstrap_arbitrator(context):
...
+    """
+    stonith_enabled = utils.get_property("stonith-enabled")
+    # When stonith disabled, return
+    if utils.is_boolean_false(stonith_enabled):
+        return
+
+    if not utils.service_is_active("sbd.service"):
```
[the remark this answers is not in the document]
liangxin1300: Changed as
```
    _value = utils.get_property("stonith-timeout")
    stonith_timeout_from_cib = int(_value.strip('s')) if _value else 0
    stonith_timeout_from_calculation = max(STONITH_TIMEOUT_DEFAULT, corosync.token_and_consensus_timeout())
    if stonith_timeout_from_cib < stonith_timeout_from_calculation:
        utils.set_property(stonith_timeout=stonith_timeout_from_calculation)
```

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+        sbd_delay_start_file = "{}/sbd_delay_start.conf".format(SBD_SYSTEMD_DELAY_START_DIR)
+        utils.str2file("[Service]\nTimeoutSec={}".format(int(1.2*self.sbd_delay_start_value)), sbd_delay_start_file)
+        bootstrap.csync2_update(SBD_SYSTEMD_DELAY_START_DIR)
+        utils.cluster_run_cmd("systemctl daemon-reload")
+
+    def adjust_stonith_timeout(self):
+        """
+        Adjust stonith-timeout property
+        """
+        utils.set_property(stonith_timeout=self.get_stonith_timeout())
```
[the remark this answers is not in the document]
liangxin1300: I used to be like this.
But what if in 2-node cluster, the stonith-timeout will be large;
Then the third node joining, then should we `decrease` the stonith-timeout value, or just keep it since origin is larger?

crmsh/sbd.py
liangxin1300: Already marked with TODO

crmsh/sbd.py
liangxin1300: I can't get your means here
> to convert either 1 or yes to 2*SBD_WATCHDOG_TIMEOUT

Whether this function's return value is sbd delay start, or 2*SBD_WATCHDOG_TIMEOUT?

> to take the exception of int() to 0, eg. for "no/NO/false/False/xxx"

I don't understand what does this mean?

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+        if dev_list:  # disk-based
+            self.disk_based = True
+            self.msgwait = SBDTimeout.get_sbd_msgwait(dev_list[0])
+            self.pcmk_delay_max = utils.get_pcmk_delay_max(self.two_node_without_qdevice)
+        else:  # disk-less
+            self.disk_based = False
+            self.sbd_watchdog_timeout = SBDTimeout.get_sbd_watchdog_timeout()
+            self.stonith_watchdog_timeout = SBDTimeout.get_stonith_watchdog_timeout()
+
+        self.sbd_delay_start_value = self.get_sbd_delay_start_value()
```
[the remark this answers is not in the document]
liangxin1300: Done

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+        cls_inst = cls()
+        cls_inst._load_configurations()
+        return cls_inst.get_stonith_timeout()
+
+    def get_sbd_delay_start_value(self):
+        """
+        Get the value for SBD_DELAY_START, formulas are:
+
+        SBD_DELAY_START = (token + consensus + pcmk_delay_max + msgwait)  # for disk-based sbd
+        SBD_DELAY_START = (token + consensus + 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
```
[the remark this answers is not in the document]
liangxin1300: Done

crmsh/sbd.py
```diff
@@ -3,14 +3,263 @@
...
+
+        self.sbd_delay_start_value = self.get_sbd_delay_start_value()
+        logger.debug("Inspect SBDTimeout: %s", vars(self))
+
+    def get_stonith_timeout(self):
+        """
+        Get stonith-timeout value for sbd cases, formulas are:
+
+        stonith-timeout = 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
+        stonith-timeout = 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
```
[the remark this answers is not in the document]
liangxin1300: Done

crmsh/sbd.py
liangxin1300: Done

crmsh/sbd.py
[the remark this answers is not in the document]
zzhou1: Good argument. 

Normally, the difference  between 2-node and 3-node is pcmk_delay_max, which is usually 30. No big deal to me. 
The debatable question here is whether to protect any larger stonith_timeout defined outside of crmsh. I personally would like to give the priority to the user, as the consistent policy if possible.

zzhou1 (changes requested):
One major change is needed for the conversation around get_sbd_delay_start()

zzhou1:
`crm cluster remove tw3` works nicely so far.

Well, I expect the following alternative will do the same job.
`crm node delete tw3`

liangxin1300:
> `crm cluster remove tw3` works nicely so far.
> 
> Well, I expect the following alternative will do the same job. `crm node delete tw3`

Do you mean `crm cluster remove <node>` and `crm node delete` share the same code functions and execute the the process?

crmsh/sbd.py
```diff
@@ -3,14 +3,265 @@
...
+            cmd = "crm resource param {} set pcmk_delay_max {}s".format(SBDManager.SBD_RA_ID, self.pcmk_delay_max)
+        else:
+            cmd = "crm resource param {} delete pcmk_delay_max".format(SBDManager.SBD_RA_ID)
+        utils.get_stdout_or_raise_error(cmd)
+
+    def adjust_delay_start(self):
+        """
+        Adjust SBD_DELAY_START in /etc/sysconfig/sbd
+        """
+        value = str(self.sbd_delay_start_value) if self.two_node_without_qdevice else "no"
```
zzhou1, line 240: Sorry, SBD_DELAY_START is only matters to virtualization environment no matter diskless or disk-based, no matter the cluster size or qdevice.

The better name is adjust_sbd_delay_start() to make it consistent to get_sbd_delay_start()

crmsh/sbd.py
```diff
@@ -3,14 +3,265 @@
...
+        cls_inst._load_configurations()
+
+        message = "Adjusting sbd related timeout values for 2-node cluster"
+        if not cls_inst.two_node_without_qdevice:
+            message = "Reverting sbd related timeout values for non 2-node cluster"
+
+        with logger_utils.status_long(message):
+            cls_inst.adjust_delay_start()
+            cls_inst.adjust_pcmk_delay_max()
+            cls_inst.adjust_stonith_timeout()
```
zzhou1, line 261: adjust_stonith_timeout()
adjust_pcmk_delay_max()

These two has nothing related to the virtualization environment, must run before detect_virt().

And they doesn't belong to sbd.py

zzhou1 (changes requested):

crmsh/sbd.py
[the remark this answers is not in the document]
liangxin1300: Changed, keep unchanged when existing "stonith-timeout" is larger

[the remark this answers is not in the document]
liangxin1300: Changed, keep unchanged when existing "stonith-timeout" is larger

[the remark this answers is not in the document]
liangxin1300: I created this reusable function:
```
def set_property_conditionally(property_name, value_from_calculation):
    """
    Set cluster property if calculated value is larger then current cib value
    """
    _value = get_property(property_name)
    value_from_cib = int(_value.strip('s')) if _value else 0
    if value_from_cib < value_from_calculation:
        cmd = "crm configure property {}={}".format(property_name, value_from_calculation)
        get_stdout_or_raise_error(cmd)
```

crmsh/sbd.py
liangxin1300: Changed as:
```
def adjust_sbd_delay_start(self):
        """
        Adjust SBD_DELAY_START in /etc/sysconfig/sbd
        """
        value_from_configuration = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
        value_from_calculation = str(self.sbd_delay_start_value) if utils.detect_virt() else "no"
        if value_from_configuration != value_from_calculation:
            SBDManager.update_configuration({"SBD_DELAY_START": value_from_calculation})
```

crmsh/sbd.py
liangxin1300: Changed as 
```
    def get_sbd_delay_start_sec_from_sysconfig():
        """
        Get suitable systemd start timeout for sbd.service
        """
        value = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
        if value in ["yes", "1"]:
            return 2*SBDTimeout.get_sbd_watchdog_timeout()
        return int(value)
```
This function will not be called in `start_pacemaker` if SBD_DELAY_START set to "no"

crmsh/sbd.py
liangxin1300: `adjust_stonith_timeout` in SBDTimeout class is for sbd, it get runtime stonith timeout value from msgwait or SBD_WATCHDOG_TIMEOUT, then compare with the value form cib property, at last use the larger one

`bootstrap.adjust_stonith_timeout_for_all_situations` will be called after above, still compare with the value from cib proerty, at last use the larger one

zzhou1: Okay, detect_virt() finds it appropriate places elsewhere. 

Agree with you, adjust_stonith_timeout() do have the direct relationship with SBD.

I agree to close this conversation and defer the TODO to adjust_pcmk_delay_max() function

crmsh/sbd.py
zzhou1: For get_sbd_delay_start_sec_from_sysconfig(), please at least add a TODO to parse "5ms, 5us, 5s, 5m, 5h" in the future.

zzhou1 (changes requested):
Minor changes are expected to resolve conversations.

zzhou1:
> > `crm cluster remove tw3` works nicely so far.
> > Well, I expect the following alternative will do the same job. `crm node delete tw3`
> 
> Do you mean `crm cluster remove <node>` and `crm node delete` share the same code functions and execute the the process? Then we need another PR to do this

If the code can be reused directly, that will be great. From the behavior wise, `crm node delete <node>` should adjust_pcmk_delay_max here. Maybe, most of other behavior should be the same too. Anyway, worth an issue to track this.

crmsh/sbd.py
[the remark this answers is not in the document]
zzhou1: I can understand. Whether "set" or "adjust" is kind of slightly different language context. It's fine if you stick with your "set". I personally like "adjust"

Even so, I would suggest a little more detailed function name, eg.
set_sbd_watchdog_timeout_with_profiles_yml()

zzhou1 (approved):

gao-yan:
> > stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations
> 
> It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)

I'm a little lost here...

Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with time that will be spent on fencing?

wenningerk:
> > 
> > It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)
> 
> I'm a little lost here...
> 
> Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with the time that will be spent on fencing?

That corresponds to the final conclusion Jan Friesse and me came to.
And is the reason why I finally didn't implement that test in sbd itself - as done with sync-timeout.
https://bugzilla.redhat.com/show_bug.cgi?id=1877434 was closed a while ago.
Let me copy here a statement from the private part of that bug.

```
After reconsideration it looks as if there are no relevant cases that would
require a relationship token timeout and sbd watchdog timeout (and from
that derived stonith-watchdog-timeout).

All fencing actions are determined from the current CIB state. Due to
corosync's virtual synchrony, there should be no situations where the
target node is in the membership, and the DC sees a CIB change that
requires fencing but the target node does not also see that change.
Therefore, the DC and sbd on the target node should always be in sync
as far as whether fencing is required, and it should happen within the
sbd timeout.
```

Sry for not interfering here earlier.

gao-yan:
> That corresponds to the final conclusion Jan Friesse and me came to. And is the reason why I finally didn't implement that test in sbd itself - as done with sync-timeout. https://bugzilla.redhat.com/show_bug.cgi?id=1877434 was closed a while ago. Let me copy here a statement from the private part of that bug.
> 
> ```
> After reconsideration it looks as if there are no relevant cases that would
> require a relationship token timeout and sbd watchdog timeout (and from
> that derived stonith-watchdog-timeout).
> 
> All fencing actions are determined from the current CIB state. Due to
> corosync's virtual synchrony, there should be no situations where the
> target node is in the membership, and the DC sees a CIB change that
> requires fencing but the target node does not also see that change.
> Therefore, the DC and sbd on the target node should always be in sync
> as far as whether fencing is required, and it should happen within the
> sbd timeout.
> ```

Thanks for the information, @wenningerk!

IIUC it means, for the case of diskless sbd, the only requirement in regard of SBD_WATCHDOG_TIMEOUT remains -- A value of SBD_WATCHDOG_TIMEOUT is safe as long as it's longer than qdevice sync timeout if qdevice is used, right?

wenningerk:
> 
> Thanks for the information, @wenningerk!
> 
> IIUC it means, for the case of diskless sbd, the only requirement in regard of SBD_WATCHDOG_TIMEOUT remains -- A value of SBD_WATCHDOG_TIMEOUT is safe as long as it's longer than qdevice sync timeout if qdevice is used, right?

yes, as of my current understanding at least ;-)

gao-yan:
> > > stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations
> > 
> > 
> > It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)
> 
> I'm a little lost here...
> 
> Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with the time that will be spent on fencing?

@liangxin1300 @zzhou1 What's confusing me here is, you seem to be talking about determination of `stonith-timeout` for non-sbd cases? But "token + consensus" is not relevant even for the cases of sbd with or without shared disk.

I mean, yes, "token + consensus" should be added into `SBD_DELAY_START`,  but I don't see them as a factor of the generic `stonith-timeout`.

zzhou1:
> > > > stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations
> > > 
> > > 
> > > It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)
> > 
> > 
> > I'm a little lost here...
> > Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with the time that will be spent on fencing?
> 
> @liangxin1300 @zzhou1 What's confusing me here is, you seem to be talking about determination of `stonith-timeout` for non-sbd/generic cases? But `token + consensus` is not relevant even for the cases of sbd with or without shared disk.
> 
> I mean, yes, `token + consensus` should be added into `SBD_DELAY_START`, but I don't see them as a factor of the generic `stonith-timeout`.

@gao-yan  it is a generic situation. I observed this and found the detailed discussion here. 
https://bugzilla.redhat.com/show_bug.cgi?id=1941108

gao-yan:
> > > > > stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations
> > > > 
> > > > 
> > > > It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)
> > > 
> > > 
> > > I'm a little lost here...
> > > Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with the time that will be spent on fencing?
> > 
> > 
> > @liangxin1300 @zzhou1 What's confusing me here is, you seem to be talking about determination of `stonith-timeout` for non-sbd/generic cases? But `token + consensus` is not relevant even for the cases of sbd with or without shared disk.
> > I mean, yes, `token + consensus` should be added into `SBD_DELAY_START`, but I don't see them as a factor of the generic `stonith-timeout`.
> 
> @gao-yan I treat is a generic situation. I observed this in my lab and referred to the detailed discussion here. https://bugzilla.redhat.com/show_bug.cgi?id=1941108

Ah, now I recall it's about the case where the fencing target is still technically in the membership when fencing action returns, and how to prevent the unnecessary second fencing...

If to address this case, "token + consensus" needs to be generally *added* into stonith-timeout.

> 
> Well, this triggers me to correlate to "Pacemaker Explain" document: "It(stonith-timeout) has been replaced by the pcmk_reboot_timeout and pcmk_off_timeout properties. " I might over look stonith-timeout here. Maybe, stonith-timeout is not used without SBD at all? Thanks to clarify on this.

I think this is about the parameters of fencing resource rather than the global stonith-timeout.

zzhou1:
> > Well, this triggers me to correlate to "Pacemaker Explain" document: "It(stonith-timeout) has been replaced by the pcmk_reboot_timeout and pcmk_off_timeout properties. " I might over look stonith-timeout here. Maybe, stonith-timeout is not used without SBD at all? Thanks to clarify on this.
> 
> I think this is about the parameters of fencing resource rather than the global stonith-timeout.
> 
> And of course any such parameter configured for specific fencing resource takes precedence over the global one.

Super! Thanks to clear my mind.

crmsh/sbd.py
```diff
@@ -26,14 +275,11 @@ class SBDManager(object):
   specify here will be destroyed.
 """
     SBD_WARNING = "Not configuring SBD - STONITH will be disabled."
-    DISKLESS_SBD_WARNING = """Diskless SBD requires cluster with three or more nodes.
-If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."""
+    DISKLESS_SBD_WARNING = "Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for two-nodes cluster, should be combined with QDevice."
```
gao-yan, line 278: Nitpicking: two-node or 2-node rather than two-nodes :-)

gao-yan (commented):
Thanks for the good work, @liangxin1300 !

liangxin1300:
@gao-yan @zzhou1 Thanks for your nice suggestions!


> Basically token + consensus needs to be generally added into the value for that.

I changed `get_stonith_timeout_runtime` as `get_stonith_timeout_expected`, and compare the value from sbd with max(STONITH_TIMEOUT_DEFAULT, token+consensus), return the larger one
```
    def get_stonith_timeout_expected(self):
        """
        Get stonith-timeout value for sbd cases, formulas are:

        stonith-timeout >= 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
        stonith-timeout >= 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd

        And above value will compare with max(STONITH_TIMEOUT_DEFAULT, token+consensus),
        then return the larger one
        """
        if self.disk_based:
            value_from_sbd = int(1.2*(self.pcmk_delay_max + self.msgwait))
        else:
            value_from_sbd = int(1.2*max(self.stonith_watchdog_timeout, 2*self.sbd_watchdog_timeout))

        value = max(value_from_sbd, bootstrap.get_stonith_timeout_generally_expected())
        logger.debug("Result of SBDTimeout.get_stonith_timeout_expected %d", value)
        return value
```

gao-yan:
> @gao-yan @zzhou1 Thanks for your nice suggestions!
> 
> > Basically token + consensus needs to be generally added into the value for that.
> 
> I changed `get_stonith_timeout_runtime` as `get_stonith_timeout_expected`, and compare the value from sbd with max(STONITH_TIMEOUT_DEFAULT, token+consensus), return the larger one
> 
> ```
>     def get_stonith_timeout_expected(self):
>         """
>         Get stonith-timeout value for sbd cases, formulas are:
> 
>         stonith-timeout >= 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
>         stonith-timeout >= 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
> 
>         And above value will compare with max(STONITH_TIMEOUT_DEFAULT, token+consensus),
>         then return the larger one
>         """
>         if self.disk_based:
>             value_from_sbd = int(1.2*(self.pcmk_delay_max + self.msgwait))
>         else:
>             value_from_sbd = int(1.2*max(self.stonith_watchdog_timeout, 2*self.sbd_watchdog_timeout))
> 
>         value = max(value_from_sbd, bootstrap.get_stonith_timeout_generally_expected())
>         logger.debug("Result of SBDTimeout.get_stonith_timeout_expected %d", value)
>         return value
> ```

If to address:
https://bugzilla.redhat.com/show_bug.cgi?id=1941108

, `token+consensus` needs to be additionally added onto the time that fencing itself potentially will need, meaning for the default:

`stonith-timeout = STONITH_TIMEOUT_DEFAULT + (token+consensus)`

For the calculated:
`stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) + (token+consensus)`

gao-yan:
> > > > > stonith-timeout >= max(STONITH_TIMEOUT_DEFAULT, token+consensus)  # for other situations
> > > > 
> > > > 
> > > > It is not "for other situations". It is "for ALL situations". eg. there could be the situation (token+consensus) > (pcmk_delay_max + msgwait) eg. there could be the situation (token+consensus) > max(stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT)
> > > 
> > > 
> > > I'm a little lost here...
> > > Fencing starts, meaning stonith timer starts only when a new membership has been reformed, which is after "token + consensus" have timed out. Why does "token + consensus" matter for determining the value of stonith-timeout or need to be compared with the time that will be spent on fencing?
> > 
> > 
> > @liangxin1300 @zzhou1 What's confusing me here is, you seem to be talking about determination of `stonith-timeout` for non-sbd/generic cases? But `token + consensus` is not relevant even for the cases of sbd with or without shared disk.
> > I mean, yes, `token + consensus` should be added into `SBD_DELAY_START`, but I don't see them as a factor of the generic `stonith-timeout`.
> 
> @gao-yan I treat is a generic situation. I observed this in my lab and referred to the detailed discussion here. https://bugzilla.redhat.com/show_bug.cgi?id=1941108
> 
> Well, this triggers me to correlate to "Pacemaker Explain" document: "It(stonith-timeout) has been replaced by the pcmk_reboot_timeout and pcmk_off_timeout properties. " I might over look stonith-timeout here. Maybe, stonith-timeout is not used without SBD at all? Thanks to clarify on this.

zzhou1:
@gao-yan 

See my view down below,

> If to address: https://bugzilla.redhat.com/show_bug.cgi?id=1941108
> 

Over there, actually, I didn't figure out how to illustrate the precise rationale for "pcmk_reboot_timeout to a value more than (token) + (consensus) + (how long they expect the reboot to take)". At least I couldn't prove that addition part for how_long_they_expect_the_reboot_to_take. Noted that, the description of brc#1941108#c0, node2 won't start the cluster stack at all. With that, how_long_they_expect_the_reboot_to_take can be a infinite number in theory ;)

My understanding and expectation is that as long as pacemaker fencer op returns "OK" successfully, then this fence failure can avoid, hence no double fencing. And, I observe pacemaker-fenced do ack 'reboot' ok once the new corosync membership formed at the time of token+timeout. That says, as long as  stonith-timeout > token+timeout, there won't be fence failure, hence no double fencing. BTW, sbd poison pill is special and could avoid double fencing after all. This is true. But other normal fence devices is not. I can prove this double fencing with fence_virsh in my lab with the disabled cluster service . 

My wild guess is whether brc#1941108 might pull in the similar concept of SBD_DELAY_START, and mess two different concepts?!

> , `token+consensus` needs to be generally additionally added onto the time that fencing itself potentially will need, meaning for the default:
> 
> `stonith-timeout = STONITH_TIMEOUT_DEFAULT + (token+consensus)`
> 
> For the calculated: `stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) + (token+consensus)`

All in all, with my above narrative, in theory, I don't see the need to add them together. Though, practically, the bigger number surely add enough more buffer.

gao-yan:
Considering the current default/calculated `stonith-timeout` alone, it's already the time that is expected to possibly take for a fencing to return. A fencing target can be safely assumed being down only when a fencing has returned. And only until then we can assume the node starts leaving corosync membership, and it'll then take (token+consensus) for us get a new membership reformed.

As you can see from the bug entry,  although the fencing succeeded very soon, but it could not be acked until the new membership is reformed. So as long as stonith-timeout timer is popped before that, it will anyway cause fencing failure/timeout and result into another fencing.

Fencing itself only took a couple of seconds to return, which might make you think it's insignificant. But it consumed stonith-timeout. It'd be more obvious with sbd of course.

> 
> > , `token+consensus` needs to be generally additionally added onto the time that fencing itself potentially will need, meaning for the default:
> > `stonith-timeout = STONITH_TIMEOUT_DEFAULT + (token+consensus)`
> > For the calculated: `stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) + (token+consensus)`
> 
> All in all, with my above narrative, in theory, I stick with `stonith-timeout >= max(value_from_sbd, STONITH_TIMEOUT_DEFAULT, token+consensus )` instead of adding them together as the above. Though, practically, the bigger number surely add enough more buffer.

I don't really get what this is for :-)

To me, it's like either not to address the case with:

`stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) 

, or address it with:

`stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT)  + (token+consensus)`

Frankly I don't think having a longer stonith-timeout could do harm as long as we have the good reason.

zzhou1:
> Considering the current default/calculated `stonith-timeout` alone, it's already the time that is expected to be possibly taken for a fencing to return. A fencing target can be safely assumed being down only when a fencing has returned. And only until then we can assume the node starts leaving corosync membership, and it'll then take (token+consensus) for us get a new membership reformed.

From my experience to read the log sbd and fence_virsh, I could see  two places pacemaker-fenced try to return "OK" or "error"
1. once reach stonith-timeout, 
2. once reach token+consensus

> 
> As you can see from the bug entry, although the fencing action itself succeeded very soon, but it could not be acked until the new membership is reformed. So as long as stonith-timeout timer is popped before that, it will anyway cause fencing failure/timeout and result into another fencing.

Yes, I think I understand this part. Also because of this it leads me to stick with stonith-timeout > token+consensus.

> 
> Fencing itself only took a couple of seconds to return, which might make you think it's insignificant. But it consumed stonith-timeout. It'd be more obvious with sbd of course.
> 
> > > , `token+consensus` needs to be generally additionally added onto the time that fencing itself potentially will need, meaning for the default:
> > > `stonith-timeout = STONITH_TIMEOUT_DEFAULT + (token+consensus)`
> > > For the calculated: `stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) + (token+consensus)`
> > 
> > 
> > All in all, with my above narrative, in theory, I stick with `stonith-timeout >= max(value_from_sbd, STONITH_TIMEOUT_DEFAULT, token+consensus )` instead of adding them together as the above. Though, practically, the bigger number surely add enough more buffer.
> 
> I don't really get what this is for :-)
> 
> To me, it's like either not to address the case with:
> 
> `stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) `
> 
> , or address it with:
> 
> `stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT) + (token+consensus)`
> 

Let me put this way by using the use cases

a) sbd use cases:     ​
   stonith-timeout = max(value_from_sbd, STONITH_TIMEOUT_DEFAULT)

   ​This is safe even not take "token+consensus" into account. The ugly
   situation is acceptable when it is less than (token+consensus). Because of
   SBD_DELAY_START, two poison pills makes no difference as one poison pill.

b) fence_virsh use cases, as the example for non-sbd situation:
   ​stonith-timeout = max(STONITH_TIMEOUT_DEFAULT, (token+consensus))

   There is no influence from value_from_sbd at all.

> Frankly I don't think having a longer stonith-timeout could do harm as long as we have the good reason.

I'm kind of agree , because of no obviously harm. However, pacemaker code internally will multiply 1.2 to the user-end stonith-timeout, and it does gives the enough buffer. I don't have to insist to add too much buffer.

Another point from my technical debate is I couldn't find the real example to illustrates to add all above values together.

That says, 
`stonith-timeout >= max(value_from_sbd, STONITH_TIMEOUT_DEFAULT, token+consensus)` 
it covers both sbd and non-sbd situation as the generic formula.

BTW, here is my test configuration with fence_virsh

    - Bad configuration cause fencing failure (aka. double fencing)
      totem.token (u32) = 30000  => token+consensus will be 66
      stonith-timeout=50  => internally pacameker will get 60, which < 66

    - Good configuration has only one single fencing
      totem.token (u32) = 30000  => token+consensus will be 66
      stonith-timeout=70  => internally pacameker will get 84, which > 66

gao-yan:
> > Considering the current default/calculated `stonith-timeout` alone, it's already the time that is expected to be possibly taken for a fencing to return. A fencing target can be safely assumed being down only when a fencing has returned. And only until then we can assume the node starts leaving corosync membership, and it'll then take (token+consensus) for us get a new membership reformed.
> 
> From my experience to read the log sbd and fence_virsh, I could see two places pacemaker-fenced try to return "OK" or "error"
> 
> 1. once reach stonith-timeout,
> 2. once reach token+consensus

The point is, we wouldn't want to unnecessarily pop stonith-timeout only because we are still in reforming of new membership hence DC is not ready to acknowledge a fencing which was even successful ...

> b) fence_virsh use cases, as the example for non-sbd situation:​stonith-timeout = max(STONITH_TIMEOUT_DEFAULT, (token+consensus))

Things might not be always as easy as with fence_virsh for example in public clouds purely with their own fencing mechanisms...

> 
> There is no influence from value_from_sbd at all.
> 
> > Frankly I don't think having a longer stonith-timeout could do harm as long as we have the good reason.
> 
> I'm kind of agree , because of no obviously harm. However, pacemaker code internally will multiply 1.2 to the user-end stonith-timeout, and it does gives the enough buffer. I don't have to insist to add too much buffer.

Bear in mind, the margin is meant for the internal overheads, broadcasting/processing/acknowledging of fencing requests/replies among cluster nodes, which could be significant depending on the environment/situation ...

Just because we know there's the internal margin and usually it's not consumed up shouldn't encourage us to consider it for other purpose.

> 
> Another point from my technical debate is I couldn't find the real example to illustrates to add all above values together.

Well, because it's based on the experience/assumptions with normal/good situations :-)

A lot of fencing mechanisms usually take less than 1 second to finish fencing, but rather just giving it only 1 second stonith-timeout assuming it can always do the job with a  single quick shot, we give it longer time for possible bad situations and for it to even retry ...

It's not that I think the case must or must not be addressed, it just doesn't make sense to me to go with a halfway solution, otherwise we could get a conversation like:

Q: Why is my stonith-timeout bootrapped being equal to "token + consensus"?
A: To prevent unnecessary second fencing
Q: So it does prevent that?
A: Well, mostly probably ...

zzhou1:
> The point is, we wouldn't want to unnecessarily pop stonith-timeout only because we are still in reforming of new membership hence DC is not ready to acknowledge a fencing which was even successful ...

Agree.

> 
> > b) fence_virsh use cases, as the example for non-sbd situation:​stonith-timeout = max(STONITH_TIMEOUT_DEFAULT, (token+consensus))
> 
> Things might not be always as easy as with fence_virsh for example in public clouds purely with their own fencing mechanisms...

Thanks, yes, the public cloud and openstack fence mechanism on their own are the good example. They need very long stonith-timeout. This re-enforce us to use ">=" in the code, instead of "=".

[...]

> 
> A lot of fencing mechanisms usually take less than 1 second to finish fencing, but rather just giving it only 1 second stonith-timeout assuming it can always do the job with a single quick shot, we give it longer time for possible bad situations and for it to even retry ...
> 

That's a valid angle in the sense of the bad situation. I would correlate this to public cloud fence agents again. The final appropriate stonith-timeout should leverage the knowledge of sysadmin or the external orchestration software, for example. That's beyond the scope of crmsh so far.

> It's not that I think the case must or must not be addressed, it just doesn't make sense to me to go with a halfway solution, otherwise we could get a conversation like:
> 
> Q: Why is my stonith-timeout bootrapped being equal to "token + consensus"? A: To prevent unnecessary second fencing Q: So it does prevent that? A: Well, mostly probably ...
 
Now, I can see your point, and where I stand for. My point is based on ">=" which tries to stretch stonith-timeout to the reasonable minimum, no necessarily to blindly force "=" to cover all situation. 

Anyway, toward closure of this debate with stonith-timeout, I defer the final decision to you and/or Xin in the code ;)

Thanks, I enjoy this kind of debate. Have a nice weekend!

liangxin1300:
Thanks for the nice discussion and suggestions! I've updated the code about stonith-timeout, the formula:
```
      value_from_sbd = 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
      value_from_sbd = 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
      stonith_timeout >= max(value_from_sbd, constants.STONITH_TIMEOUT_DEFAULT) + token + consensus
      stonith-timeout >= STONITH_TIMEOUT_DEFAULT+token+consensus  # for all situations
```

crmsh/sbd.py
```diff
@@ -3,14 +3,261 @@
...
+        logger.debug("Inspect SBDTimeout: %s", vars(self))
+
+    def get_stonith_timeout_expected(self):
+        """
+        Get stonith-timeout value for sbd cases, formulas are:
+
+        value_from_sbd = 1.2 * (pcmk_delay_max + msgwait) # for disk-based sbd
+        value_from_sbd = 1.2 * max (stonith_watchdog_timeout, 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
+
+        stonith_timeout >= max(value_from_sbd, constants.STONITH_TIMEOUT_DEFAULT) + token + consensus
```
gao-yan, line 141: Similarly about ">=" in the comment.

crmsh/sbd.py
```diff
@@ -3,14 +3,261 @@
...
+        cls_inst = cls()
+        cls_inst._load_configurations()
+        return cls_inst.get_stonith_timeout_expected()
+
+    def get_sbd_delay_start_expected(self):
+        """
+        Get the value for SBD_DELAY_START, formulas are:
+
+        SBD_DELAY_START >= (token + consensus + pcmk_delay_max + msgwait)  # for disk-based sbd
+        SBD_DELAY_START >= (token + consensus + 2*SBD_WATCHDOG_TIMEOUT) # for disk-less sbd
```
gao-yan, line 163: Similarly.

crmsh/sbd.py
```diff
@@ -3,14 +3,261 @@
...
+        Adjust start timeout for sbd when set SBD_DELAY_START
+        """
+        sbd_delay_start_value = SBDManager.get_sbd_value_from_config("SBD_DELAY_START")
+        if sbd_delay_start_value == "no":
+            return
+
+        cmd = "systemctl show -p TimeoutStartUSec sbd --value"
+        out = utils.get_stdout_or_raise_error(cmd)
+        start_timeout = utils.get_systemd_timeout_start_in_sec(out)
+        if start_timeout >= int(sbd_delay_start_value):
```
gao-yan, line 202: Allowing "==" would be a little tricky... Probably better require ">".

gao-yan (approved):
Generally looks good to me.

Besides a few more nitpickings here, there are still unresolved conversations including the ones in is_sbd_delay_start() and adjust_systemd_start_timeout() and is_2node_cluster_without_qdevice(). Feel free to resolve or improve in the future.

liangxin1300:
@zzhou1 @gao-yan Thanks for reviewing!

PR #892, opened 2021-11-28
Fix: crm: Avoid possible Permission denied error under other users

liangxin1300:


crmsh/log.py
```diff
@@ -398,11 +398,15 @@ def setup_directory_for_logfile():
     os.makedirs(_dir, exist_ok=True)
 
 
-def setup_logging():
+def setup_logging(only_help=False):
     """
     Setup log directory and loadding logging config dict
     """
-    setup_directory_for_logfile()
+    # To avoid possible Permission denied error under other users
```
zzhou1, line 405: It is boo# not bsc# in the commit header, and probably also make sense to append to the code comment too.

# To avoid the potential "permission denied" error under other users (boc#1192754)

liangxin1300: Changed
Thanks!

zzhou1 (approved):

PR #893, opened 2021-11-30
Dev: ui_node: Improve node standby/online methods

liangxin1300:
In #797, standby/online part, there were issues:
- Can not handle and replace `false/no/n/0` value for standby
- standby nvpair could not contain node_id, then replace will not matched and failed

To avoid above issues:
- Not use re.sub to replace whole cib string
- Use xml query then `set` the value directly

Other than that:
- Avoid leading with number for xml ID
- Drop already standby/online check
- When doing standby, remove another type of lifetime standby

### DIFF
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -3,7 +3,6 @@
 # See COPYING for license information.
 
 import re
-from lxml import etree
 from . import config
 from . import command
 from . import completers as compl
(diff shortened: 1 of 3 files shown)


crmsh/ui_node.py
```diff
@@ -332,18 +332,24 @@ def do_standby(self, context, *args):
...
+    def _do_standby_or_online_action(self, cib, node_list, action):
+        """
+        Common part of standby and online
+        """
         # Consider both "nodes" and "status" section might contain different standby value at the same time
-        # Should replace all possible "off" values here
         for node in node_list:
             node_id = utils.get_nodeid_from_name(node)
-            cib_str = re.sub(constants.STANDBY_NV_RE.format(node_id=node_id, value="off"), r'\1value="on"\2', cib_str)
+            for query_path in [constants.XML_NODE_QUERY_STANDBY_PATH, constants.XML_STATUS_QUERY_STANDBY_PATH]:
```
gao-yan, line 344: According to the newly expected `lifetime`, to prevent potential interference,  we should probably delete any possibly existing setting with a different `lifetime`?

liangxin1300: Action `standby` only accept two kinds of lifetime: "reboot" or "forever", and these would make cib.xml no different with no lifetime set

gao-yan: For sure without explicitly specifying `lifetime`, it goes into the configuration meaning it's `forever`.

What I meant:

If an user wants to have a "reboot" standby, while there's an existing `standby` setting in the configuration no matter it's on or off. They won't actually get the "reboot" standby because the code here basically sets standby in both configuration and status.

That's the reason why I suggest:

1) If setting standby in configuration, drop any from status.
2) if setting standby in status, drop any from configuration.

liangxin1300: @gao-yan  Thanks for pointing these!
Changed

gao-yan (approved):
Looks good to me. Thanks for the nice work!

PR #895, opened 2021-12-01
Dev: ui_cluster: Check service is available before enable/disable qdevice

liangxin1300:


crmsh/ui_cluster.py
```diff
@@ -208,7 +208,7 @@ def _enable_disable_common(self, context, *args):
         node_list = parse_option_for_nodes(context, *args)
         action = context.get_command_name()
         utils.cluster_run_cmd("systemctl {} pacemaker.service".format(action), node_list)
-        if utils.is_qdevice_configured():
+        if utils.service_is_available("corosync-qdevice.service"):
```
zzhou1, line 211: Blindly disable qdevice is okay. But, enable it even without the qdevice configuration won't be intuitive I think.

zzhou1 (changes requested):

zzhou1 (approved):

PR #900, opened 2021-12-07
Dev: ui_cluster: Improvement for stop process

liangxin1300:
- Revert "Dev: ui_cluster: Make sure node is online when stop service"
  Since to wait for node online when stopping cluster seems no make sense

- Dev: log: In status_long function, add a blank line when exception
  To make the error message has more clear format

- Dev: ui_cluster: check dlm controld ra is running when stop cluster
  When dlm controld is stopped, this time trying to stop cluster will cause error when trying to fetch dlm options

- Dev: ui_cluster: Exit stop process when there is no DC

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -836,9 +836,11 @@ def append_file(dest, src):
         return False
 
 
-def get_dc():
+def get_dc(timeout=None):
     cmd = "crmadmin -D"
-    rc, s = get_stdout(add_sudo(cmd))
+    if timeout:
+        cmd += " -t {}".format(timeout)
+    rc, s, _ = get_stdout_stderr(add_sudo(cmd))
     if rc != 0:
         return None
     if not s.startswith("Designated"):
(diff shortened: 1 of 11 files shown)


crmsh/utils.py
```diff
@@ -3046,4 +3046,12 @@ def get_systemd_timeout_start_in_sec(time_res):
     res_min = re.search("(\d+)min", time_res)
     start_timeout += 60 * int(res_min.group(1)) if res_min else 0
     return start_timeout
+
+
+def any_unclean_nodes():
+    """
+    Get the UNCLEAN node list if any
+    """
+    out = get_stdout_or_raise_error("crm_mon -1")
```
zzhou1, line 3055: I wonder  it might be easy to convert the follow to pythonic crmsh code. It more elegant than grep the human text.
`crm_mon --output-as=xml | xmllint --xpath '//nodes/node[@unclean="true"]' - `

zhaohem (approved):
looks good to me.

crmsh/utils.py
zzhou1: It deserves a dedicate fix in general. Resolve this by,
https://github.com/ClusterLabs/crmsh/issues/902

PR #909, opened 2021-12-22
Dev: bootstrap: Avoid duplicated setting for rsc_defaults

liangxin1300:


crmsh/bootstrap.py
```diff
@@ -1215,8 +1215,10 @@ def init_cluster():
 
     crm_configure_load("update", """property cib-bootstrap-options: stonith-enabled=false
 op_defaults op-options: timeout=600 record-pending=true
-rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
+    rsc_defaults_id = xmlutil.RscState().get_rsc_default_id(default="rsc-options")
```
zzhou1, line 1219: 1. pacemaker will set the default of `resource-stickiness=1` for  build-resource-defaults` attribute during the build time on the purpose to setup the default stickiness value for the distribution.

2. almost no value for bootstrap to set `resource-stickiness=1` once again, when rsc_defaults_id detect the first attribute `build-resource-defaults` which already has `resource-stickiness=1`

3. The concept is more naturally  to let `migration-threshold=3` belong to `rsc-options`(better to be `rsc_defaults-options`) attribute instead of `build-resource-defaults` attribute

From my view, the whole commit would be as simple as,

- rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
+rsc_defaults rsc_defaults-options: migration-threshold=3 

NOTE: I do prefer to reuse the same attribute name `rsc_defaults-options` of pacemaker. [REF to: pcmk explained]

zzhou1 (approved):

PR #923, opened 2022-01-14
Fix: bootstrap: Don't change pacemaker.service bootup preference

liangxin1300:
Don't change pacemaker.service bootup preference when running `crm cluster start`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -669,11 +669,11 @@ def init_cluster_local():
     if pass_msg:
         logger.warning("You should change the hacluster password to something more secure!")
 
-    start_pacemaker()
+    start_pacemaker(enable_flag=True)
     wait_for_cluster()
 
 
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -669,11 +669,11 @@ def init_cluster_local():
     if pass_msg:
         logger.warning("You should change the hacluster password to something more secure!")
 
-    start_pacemaker()
+    start_pacemaker(enable=True)
     wait_for_cluster()
 
 
-def start_pacemaker(node_list=[]):
+def start_pacemaker(node_list=[], enable=False):
```
zzhou1, line 676: I tend to remove `enable` for the argument, and also remove it when calling utils.start_service().
To make the strong position that `start_pacemaker()` has nothing related to systemd `enable`.

zzhou1 (changes requested):

zzhou1 (approved):

PR #925, opened 2022-01-14
Fix: adapt report to pacemaker daemon name change

athos-ribeiro:


liangxin1300 (approved):
Thank you for your nice commit!

PR #929, opened 2022-01-19
Fix: sbd: not overwrite SYSCONFIG_SBD and sbd-disk-metadata if input 'n'(bsc#1194870)

liangxin1300:
backport from #930

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -294,6 +294,7 @@ class SBDManager(object):
         self._context = context
         self._delay_start = False
         self.timeout_inst = None
+        self.no_overwrite = False
 
     @staticmethod
     def _get_device_uuid(dev, node=None):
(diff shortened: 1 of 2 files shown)


crmsh/sbd.py
```diff
@@ -345,6 +349,7 @@ def _get_sbd_device_interactive(self):
 
         configured_dev_list = self._get_sbd_device_from_config()
         if configured_dev_list and not bootstrap.confirm("SBD is already configured to use {} - overwrite?".format(';'.join(configured_dev_list))):
```
gao-yan, line 351: The message here is not really clear about what's already configured and what's going to be overwritten. Could be sysconfig, sbd on-disk metadata, and/or sbd resource and the cluster property settings in pacemaker?

zzhou1: Recap the commit title, how about,

[Fix: sbd: not overwrite SYSCONFIG_SBD and sbd-disk-metadata if input 'n'(bsc#1194870)]

INFO: SBD is already configured to use xxxxxx - overwrite (y/n)? 

If 'n', the behavior will be
- bootstrap reuses SYSCONFIG_SBD and tries to reuse sbd-disk-metadata if get initialized already.
- That says, bootstrap skips refreshing a new SYSCONFIG_SBD, and skips 'sbd create' to initialize the sbd disk metadata if exist.

zzhou1 (changes requested):

zzhou1 (approved):

PR #930, opened 2022-01-20
[crmsh-4.3] Fix: sbd: not overwrite SYSCONFIG_SBD and sbd-disk-metadata if input 'n'(bsc#1194870)

liangxin1300:
On interactive mode, when detect sbd already configured and input 'n', the sbd configure file shouldn't be overwritten

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -225,6 +225,7 @@ If you want to use diskless SBD for two-nodes cluster, should be combined with Q
         self._stonith_timeout = 60
         self._sbd_watchdog_timeout = 0
         self._is_s390 = "390" in os.uname().machine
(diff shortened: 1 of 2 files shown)


zzhou1 (approved):

crmsh/bootstrap.py
```diff
@@ -283,6 +284,7 @@ def _get_sbd_device_interactive(self):
 
         configured_dev_list = self._get_sbd_device_from_config()
         if configured_dev_list and not confirm("SBD is already configured to use {} - overwrite?".format(';'.join(configured_dev_list))):
```
gao-yan, line 286: Same here.

gao-yan (approved):

Issue #910, opened 2021-12-22
crm node online failed


liangxin1300:
Using cibadmin to replace raise error, said the cib is order than existing
Can't repoduce it each time, I doubt the reason it's the way to dump cib
`cibadmin -Ql` then changed and replace might have race condtion issue?
How about using `cibadmin -Qs` or only -Q?

PR #932, opened 2022-01-24
Dev: ui_node: Use diff and patch instead of replace cib

liangxin1300:
To fix issue #910, avoid possible race condition


gao-yan (approved):
Looks good to me. Nice work.

Issue #860, opened 2021-08-16
enhancement: let `crm cluster join` wait the peer node to be up instead of exit with an error


zzhou1:
Minor potential change for the following error message,

sles15sp2-n2:~ # crm cluster join -yc asd-sles15sp2-n1
ERROR: cluster.join: Cluster is inactive on asd-sles15sp2-n1

How about to improve it with the retry logic,
sles15sp2-n2:~ # crm cluster join -yc asd-sles15sp2-n1
cluster.join: Cluster is inactive on asd-sles15sp2-n1. Retry in 10 seconds ...

PR #934, opened 2022-01-27
Dev: bootstrap: the joining node retries an active cluster

liangxin1300:
see #860

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -33,7 +33,7 @@ from . import corosync
 from . import tmpfiles
 from . import lock
 from . import userdir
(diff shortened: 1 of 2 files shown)


zzhou1 (changes requested):
Sounds good! Though a small comment, the commit title can be improved, how about, 
"Dev: bootstrap: the joining node retries an active cluster"

liangxin1300:
> Sounds good! Though a small comment, the commit title can be improved, how about, "Dev: bootstrap: the joining node retries an active cluster"

Thanks
Changed

zzhou1 (approved):

PR #935, opened 2022-01-27
Parametrize the log dir

aleksei-burlakov:
By default all logs are stored in /var/lib/heartbeat/trace_ra
This patch adds a new resource attribute trace_dir to the cib.file
The trace_dir should come alongside the trace_ra,
whereas trace_ra signals that the resource is traced,
the trace_dir defines the path.
IMPORTANT: This patch changes the syntax of the $ crm resource trace
command. Now, the action should be prepanded with the -a key
and the interval with the -i key and the directory with the -d command.
Those keys are required because they are optional and can be umbiguous.

### DIFF
--- a/crmsh/constants.py
+++ b/crmsh/constants.py
@@ -230,6 +230,7 @@ alert_meta_attributes = (
     "timeout", "timestamp-format"
 )
 trace_ra_attr = "trace_ra"
+trace_dir_attr = "trace_dir"
 score_types = {'advisory': '0', 'mandatory': 'INFINITY'}
 boolean_ops = ('or', 'and')
 binary_ops = ('lt', 'gt', 'lte', 'gte', 'eq', 'ne')
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -635,15 +635,17 @@ class RscMgmt(command.UI):
             return None
         return rsc
 
-    def _add_trace_op(self, rsc, op, interval):
+    def _add_trace_op(self, rsc, op, interval, dir):
         from lxml import etree
         n = etree.Element('op')
         n.set('name', op)
         n.set('interval', interval)
         n.set(constants.trace_ra_attr, '1')
+        if dir is not None:
+            n.set(constants.trace_dir_attr, dir)
         return rsc.add_operation(n)
 
-    def _trace_resource(self, context, rsc_id, rsc):
+    def _trace_resource(self, context, rsc_id, rsc, dir):
         """Enable RA tracing for a specified resource."""
         op_nodes = rsc.node.xpath('operations/op')
 
@@ -651,7 +653,7 @@ class RscMgmt(command.UI):
             for o in op_nodes:
                 if o.get('name') == name:
                     return
-            if not self._add_trace_op(rsc, name, '0'):
+            if not self._add_trace_op(rsc, name, '0', dir):
                 context.fatal_error("Failed to add trace for %s:%s" % (rsc_id, name))
         trace('start')
         trace('stop')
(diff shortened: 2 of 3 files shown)


aleksei-burlakov:
This patch adresses the PM-2641. It stores the log-dir in the cib.xml. Later resource agents should take the log-dir from there. This patch changes the syntax. Now instead
```
$ crm resource trace dummy1 monitor 50
$ crm resource untrace dummy1 monitor 50
```
the user would write
```
$ crm resource trace dummy1 -a monitor -i 50 -d /var/lib/blah/blah
$ crm resource trace dummy1 -a monitor -i 50
```
It's possible to make it w/o changing the syntax, in a separate command like
`$ crm resource logdir /var/lib/blah`
, but that would be ugly imho.

@gao-yan @ajaeger  what do you think?

crmsh/constants.py
```diff
@@ -230,6 +230,8 @@
     "timeout", "timestamp-format"
 )
 trace_ra_attr = "trace_ra"
+trace_dir_attr = "trace_dir"
+trace_dir_default = "/var/lib/heartbeat/trace_ra/Dummy/"
```
gao-yan, line 234: I think we'd still like resource-agents to decide the default. So we could basically omit it in the cib if it's not explicitly specified by the user.

crmsh/ui_resource.py
```diff
@@ -635,76 +639,110 @@ def _get_trace_rsc(self, rsc_id):
...
         else:
-            rsc.set_op_attr(op_node, constants.trace_ra_attr, "1")
+            op_node = rsc.set_op_attr(op_node, constants.trace_ra_attr, "1")
+            rsc.set_op_attr(op_node, constants.trace_dir_attr, dir)
 
     @command.completers(compl.primitives, _raoperations)
-    def do_trace(self, context, rsc_id, op=None, interval=None):
-        'usage: trace <rsc> [<op>] [<interval>]'
+    def do_trace(self, context, rsc_id, *args):
+        'usage: trace <rsc> [-a=<op>] [-i=<interval>] [-d=<log-dir>]'
```
gao-yan, line 696: AFAICS, by introducing these short arguments, it'd kind of break the convention of crmsh commands.

I understand it's for the precise parsing of these optional options and avoiding any confusions. But it seems to me it's still possible for the parsing code to intelligently differentiate a `dir` from a `op` and `interval`, since it won't be a known type of `op` or a timing value, and it should start with `/`.

gao-yan (commented):
Thanks for this, @aleksei-burlakov.

Let's not merge it until the corresponding feature has been implemented and merged from resource-agents side and possibly also pacemaker.

aleksei-burlakov:
> Does this need changes in the man page as well or is the manual auto generated?

It's only a draft. The final PR will also include the documentation.

liangxin1300:
Hi @aleksei-burlakov ,
Thanks for this commit!
Since you're using argparse, to make the content consistent between `man crm` and `crm resource help trace|untrace`,
Here is the reference:
https://github.com/ClusterLabs/crmsh/pull/797/commits/923330c29daedd39aa851b16ee412d941126f34c

aleksei-burlakov:
@gao-yan @liangxin1300 I have removed the ArgParser and the short-keys.
`But it seems to me it's still possible for the parsing code to intelligently differentiate a dir from a op and interval, since it won't be a known type of op or a timing value, and it should start with /.`
That's how it's distinguishing between the arguments now. I have tested it, it works. The PR is only missing the documentation.

aleksei-burlakov:
This draft PR is superseded by #939

PR #939, opened 2022-02-08
Dev: Parametrize the log dir

aleksei-burlakov:
By default all logs are stored in `/var/lib/heartbeat/trace_ra`
This patch adds a new resource attribute trace_dir to the `cib.file`
The `trace_dir` should come alongside the `trace_ra`,
whereas `trace_ra` signals that the resource is traced,
the `trace_dir` defines the path.

### DIFF
--- a/crmsh/constants.py
+++ b/crmsh/constants.py
@@ -230,6 +230,7 @@ alert_meta_attributes = (
     "timeout", "timestamp-format"
 )
 trace_ra_attr = "trace_ra"
+trace_dir_attr = "trace_dir"
 score_types = {'advisory': '0', 'mandatory': 'INFINITY'}
 boolean_ops = ('or', 'and')
 binary_ops = ('lt', 'gt', 'lte', 'gte', 'eq', 'ne')
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -635,15 +635,17 @@ class RscMgmt(command.UI):
             return None
         return rsc
 
-    def _add_trace_op(self, rsc, op, interval):
+    def _add_trace_op(self, rsc, op, interval, dir):
         from lxml import etree
         n = etree.Element('op')
         n.set('name', op)
         n.set('interval', interval)
         n.set(constants.trace_ra_attr, '1')
+        if dir is not None:
+            n.set(constants.trace_dir_attr, dir)
         return rsc.add_operation(n)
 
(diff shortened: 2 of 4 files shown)


jpbuecken:
Nice, as a user, this is what we needed (as you know, a cluster is used to mitigate downtimes, and if we need to enable tracing in production a full /var is not very good to avoid downtimes :-) ).

Question: Will trace_dir be available as cluster wide default option so we can add it to our default setup and only switch trace_ra per resource if needed?

e.g
```
rsc_defaults rsc-options: \
        resource-stickiness=100 \
        is-managed=true \
        trace_dir="/extra/trace/fs"
```

aleksei-burlakov:
Oops, I forgot to put the reference to the draft PR https://github.com/ClusterLabs/crmsh/pull/935

@jpbuecken , exactly, the cib file will be changed like (assumung there is a `Dummy` resource `dummy1`)

```
     <resources>                                                                                                                                                                             
       <primitive id="dummy1" class="ocf" provider="pacemaker" type="Dummy">                                                                                                                 
         <operations>                                                                                                                                                                                                    
          <op name="monitor" interval="15" id="dummy1-monitor-15"/>
           <op name="start" interval="0" id="dummy1-start-0">
             <instance_attributes id="dummy1-start-0-instance_attributes">
                <nvpair name="trace_ra" value="1" id="dummy1-start-0-instance_attributes-trace_ra"/>
+               <nvpair name="trace_dir" value="/foo/bar" id="dummy1-start-0-instance_attributes-trace_dir"/>
              </instance_attributes>                                                           
           </op>                                                                              
         </operations>                                                                        
       </primitive>                                                                           
     </resources>
```

gao-yan:
> Question: Will trace_dir be available as cluster wide default option so we can add it to our default setup and only switch trace_ra per resource if needed?

It's an op parameter, so it's likely to work with op_defaults like:
```
op_defaults trace_dir="/extra/trace/fs"
```

aleksei-burlakov:
Hi @liangxin1300 , can you please approve my PR?

liangxin1300:
> Hi @liangxin1300 , can you please approve my PR?

Sorry for late, I will take a look today

Thanks!

liangxin1300:
Other part of codes looks good to me.
Nice work! @aleksei-burlakov 
After changing the doc part, I will merge and submit the code into 15SP4 in next week

liangxin1300 (approved):

PR #949, opened 2022-03-04
Fix: utils: Only raise exception when return code of systemctl command over ssh larger than 4

liangxin1300:
When using systemctl to check service status over ssh, such return codes  are normal:
```
0 | program is running or service is OK
1 | program is dead and /var/run pid file exists
2 | program is dead and /var/lock lock file exists
3 | program is not running
4 | program or service status is unknown
```
So non zero return code with some no harm stderr output from ssh(such as with ssh banner configured), will trigger the condition to raise exception


zzhou1 (approved):
Good fix!  Approved. 
It's up to you if you take one minor comment: I prefer the inline code comment refers to "EXIT STATUS" in `man systemctl`, instead of refers to LSB directly.

PR #950, opened 2022-03-04
[crmsh-4.3] Fix: utils: Only raise exception when return code of systemctl command overssh larger than 4

liangxin1300:
backport #949


zzhou1 (approved):
Good fix! Approved.
It's up to you if you take one minor comment: I prefer the inline code comment refers to "EXIT STATUS" in man systemctl, instead of refers to LSB directly.

PR #951, opened 2022-03-15
Improve for "one qnetd for multi cluster" scenario

liangxin1300:
## Problems
* Race condition might happen when multi cluster setup qdevice at the same time, include:
  - init database on qnetd
  ```
  Certificate database (/etc/corosync/qnetd/nssdb) already exists. Delete it to initialize new db
  ```
  - qnetd sign the crq file while other cluster copy the new same name crq
* Not allow multi cluster with the same cluster name
* The crq and crt files on qnetd should be removed while removing qdevice

## Solutions
**Not allow multi cluster with the same cluster name:**
- During configuring qdevice process on this cluster, after setup passwordless between qnetd, special lock directory with cluster name should be created (like "/run/.crmsh_qdevice_lock_for_<cluster_name>")
- Raise error when finding cluster name directory exists on qnetd (like "/run/.crmsh_qdevice_lock_for_<cluster_name>")
```
ERROR: cluster.init: Duplicated cluster name "cluster1"!
```

- Raise error when the same cluster-name exists on qnetd, at the beginning fo configuring qdevice (using "corosync-qnetd-tool -l -c <cluster name>" to check)
```
ERROR: cluster.init: This cluster's name "cluster1" already exists on qnetd server!
```
- During configuring qdevice process on this cluster, after setup passwordless between qnetd, special lock directory with cluster name should be created (like "/run/.crmsh_qdevice_lock_for_<cluster_name>")

**Allow multi cluster with different cluster name:**

- On step "Initialize database on QNetd server", add lock to protect init database failed when in parallel, other clusters who failed to claim the lock should wait
- On step "sing crq on qnetd", to avoid possible race condition, the crq file should add suffix with cluster name
- Remember to clean up cluster name specific crq and crt file on qnetd node when remove qdevice from cluster

## Changes
* Dev: qdevice: Add cluster name suffix to qdevice-net-node.crq on qnet
  And remove crt and crq files on qnetd when remove qdevice
* Dev: lock: Make lock.py more generic
* Dev: qdevice: Consolidate qdevice related separate functions into qdevice.QDevice class
* Dev: qdevice: Add lock to protect init_db_on_qnetd function
* Dev: qdevice: Add lock to protect for the same cluster name
* Dev: qdevice: remove askpass since it's unnecessary after passwordless configured on qnetd
* Dev: qdevice: Improve remove qdevice process

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -2160,9 +2160,11 @@ def remove_qdevice():
 
     with logger_utils.status_long("Removing QDevice configuration from cluster"):
         qnetd_host = corosync.get_value('quorum.device.net.host')
-        qdevice_inst = qdevice.QDevice(qnetd_host)
+        cluster_name = corosync.get_value('totem.cluster_name')
+        qdevice_inst = qdevice.QDevice(qnetd_host, cluster_name=cluster_name)
         qdevice_inst.remove_qdevice_config()
         qdevice_inst.remove_qdevice_db()
+        qdevice_inst.remove_certification_files_on_qnetd()
         update_expected_votes()
     if _context.qdevice_reload_policy == QdevicePolicy.QDEVICE_RELOAD:
         invoke("crm cluster run 'crm corosync reload'")
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1829,6 +1829,7 @@ def remove_node_from_cluster():
     set_cluster_node_ip()
 
     stop_services(SERVICES_STOP_LIST, remote_addr=node)
+    qdevice.QDevice.remove_qdevice_db([node])
 
     # delete configuration files from the node to be removed
     rc, _, err = invoke('ssh {} root@{} "bash -c \\\"rm -f {}\\\""'.format(SSH_OPTION, node, " ".join(_context.rm_list)))
@@ -2063,12 +2064,9 @@ def remove_qdevice():
         invoke("crm cluster run 'systemctl stop corosync-qdevice'")
 
     with logger_utils.status_long("Removing QDevice configuration from cluster"):
-        qnetd_host = corosync.get_value('quorum.device.net.host')
-        cluster_name = corosync.get_value('totem.cluster_name')
-        qdevice_inst = qdevice.QDevice(qnetd_host, cluster_name=cluster_name)
-        qdevice_inst.remove_qdevice_config()
-        qdevice_inst.remove_qdevice_db()
-        qdevice_inst.remove_certification_files_on_qnetd()
+        qdevice.QDevice.remove_certification_files_on_qnetd()
+        qdevice.QDevice.remove_qdevice_config()
+        qdevice.QDevice.remove_qdevice_db()
         update_expected_votes()
     if qdevice_reload_policy == qdevice.QdevicePolicy.QDEVICE_RELOAD:
         invoke("crm cluster run 'crm corosync reload'")
@@ -2142,6 +2140,8 @@ def remove_self():
     else:
         # disable and stop cluster
         stop_services(SERVICES_STOP_LIST)
+        qdevice.QDevice.remove_certification_files_on_qnetd()
+        qdevice.QDevice.remove_qdevice_db([utils.this_node()])
         # remove all trace of cluster from this node
         # delete configuration files from the node to be removed
         if not invokerc('bash -c "rm -f {}"'.format(" ".join(_context.rm_list))):
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -23,7 +23,6 @@ import yaml
 from string import Template
 from lxml import etree
 from pathlib import Path
-from enum import Enum
 from contextlib import contextmanager
 from . import config
 from . import utils
@@ -57,8 +56,6 @@ PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
 SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service"]
 USER_LIST = ["root", "hacluster"]
-QDEVICE_ADD = "add"
-QDEVICE_REMOVE = "remove"
 WATCHDOG_CFG = "/etc/modules-load.d/watchdog.conf"
 BOOTH_DIR = "/etc/booth"
 BOOTH_CFG = "/etc/booth/booth.conf"
@@ -69,12 +66,6 @@ FILES_TO_SYNC = (BOOTH_DIR, corosync.conf(), COROSYNC_AUTH, CSYNC2_CFG, CSYNC2_K
         "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK, SYSCONFIG_SBD, PCMK_REMOTE_AUTH, WATCHDOG_CFG,
         PROFILES_FILE, CRM_CFG, SBD_SYSTEMD_DELAY_START_DIR)
 INIT_STAGES = ("ssh", "ssh_remote", "csync2", "csync2_remote", "corosync", "sbd", "cluster", "ocfs2", "admin", "qdevice")
-
-
-class QdevicePolicy(Enum):
-    QDEVICE_RELOAD = 0
-    QDEVICE_RESTART = 1
-    QDEVICE_RESTART_LATER = 2
 
 
 class Context(object):
@@ -110,7 +101,6 @@ class Context(object):
         self.qdevice_heuristics = None
         self.qdevice_heuristics_mode = None
         self.qdevice_rm_flag = None
-        self.qdevice_reload_policy = QdevicePolicy.QDEVICE_RESTART
         self.ocfs2_devices = []
         self.use_cluster_lvm2 = None
         self.mount_point = None
@@ -160,7 +150,8 @@ class Context(object):
                 tie_breaker=self.qdevice_tie_breaker,
                 tls=self.qdevice_tls,
                 cmds=self.qdevice_heuristics,
-                mode=self.qdevice_heuristics_mode)
+                mode=self.qdevice_heuristics_mode,
+                is_stage=self.stage == "qdevice")
 
     def _validate_sbd_option(self):
         """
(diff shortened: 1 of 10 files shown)


crmsh/bootstrap.py
```diff
@@ -1342,60 +1293,13 @@ def init_qdevice():
             utils.fatal("Failed to copy ssh key: {}".format(err))
     # Start qdevice service if qdevice already configured
     if utils.is_qdevice_configured() and not confirm("Qdevice is already configured - overwrite?"):
-        start_qdevice_service()
+        qdevice_inst.start_qdevice_service()
         return
-
+    qdevice_inst.set_cluster_name()
     # Validate qnetd node
     qdevice_inst.valid_qnetd()
```
zzhou1, line 1300: Ideally, calling qdevice_inst.valid_qnetd() here without locking will have the potential race condition, just in case there are multiple clusters with the same name. It should be protected by locking, hence better move it into  config_and_start_qdevice().

zzhou1 (commented):
Nice implementation

crmsh/qdevice.py
```diff
@@ -221,9 +297,7 @@ def init_db_on_qnetd(self):
         cmd = "corosync-qnetd-certutil -i"
         desc = "Step 1: Initialize database on {}".format(self.qnetd_addr)
         logger_utils.log_only_to_file(desc)
-        if self.askpass:
-            print(desc)
-        parallax.parallax_call([self.qnetd_addr], cmd, self.askpass)
+        parallax.parallax_call([self.qnetd_addr], cmd)
```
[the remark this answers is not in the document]
liangxin1300: I don't understand why we need to restart qnetd service here?

crmsh/bootstrap.py
liangxin1300: Is that harmful if have race condition here for the `valid_qnetd` function?

crmsh/qdevice.py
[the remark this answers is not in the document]
zzhou1: Yes, the initial point is "start" to pickup the new nssdb.  

To extend it to "restart" is for the use case that I remove the nssdb directory manually while corosync-qnetd is still running.

crmsh/bootstrap.py
zzhou1: First of all, this is a typical scenario to put "compare and change" logic under the locking context.

Now, given two fresh clusters with the same name, they will likely  go over config_and_start_qdevice() twice. One overwrite another.

I agree it is a very rare corner case, and not likely happen. But, it is better to make the code fits well with the theory, that will help code maintenance.

crmsh/qdevice.py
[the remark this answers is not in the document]
liangxin1300: Still can not get your point here, why need to remove nssdb while qnetd is still running ?

crmsh/bootstrap.py
liangxin1300:
```
    @qnetd_lock_for_same_cluster_name
    def config_and_start_qdevice(self):
```
This function was locked by `qnetd_lock_for_same_cluster_name` before enter `config_and_start_qdevice`, wouldn't go over twice

crmsh/qdevice.py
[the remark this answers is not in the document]
zzhou1: Just for advanced users/hackers/develpers' convenience sometimes. To remove the whole nssdb could simply clean up the whole nssdb, instead of the lengthy crmsh approach, or others.

crmsh/bootstrap.py
zzhou1: aha, you are right. Then, no race condition! That's a little tricky here since there are two different locks. Good for me now!

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
...
         elif not utils.package_is_installed("corosync-qnetd", self.qnetd_addr):
-            exception_msg = "Package \"corosync-qnetd\" not installed on {}".format(self.qnetd_addr)
+            exception_msg = "Package \"corosync-qnetd\" not installed on {}!".format(self.qnetd_addr)
             suggest = "install \"corosync-qnetd\" on {}".format(self.qnetd_addr)
-
+        if utils.service_is_active("corosync-qnetd", self.qnetd_addr):
+            cmd = "corosync-qnetd-tool -l -c {}".format(self.cluster_name)
+            if utils.get_stdout_or_raise_error(cmd, remote=self.qnetd_addr):
+                exception_msg = "This cluster's name \"{}\" already exists on qnetd server!".format(self.cluster_name)
+                suggest = "consider to use the different cluster-name property"
```
zzhou1, line 260: Revisit this. I feel the information of ERROR is good enough, even better to delete above line 
```suggestion
```

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
...
-            exception_msg = "host for qnetd must be a non-cluster node"
+            exception_msg = "Host for qnetd must be a non-cluster node!"
             suggest = "change to another host or stop cluster service on {}".format(self.qnetd_addr)
         elif not utils.package_is_installed("corosync-qnetd", self.qnetd_addr):
-            exception_msg = "Package \"corosync-qnetd\" not installed on {}".format(self.qnetd_addr)
+            exception_msg = "Package \"corosync-qnetd\" not installed on {}!".format(self.qnetd_addr)
             suggest = "install \"corosync-qnetd\" on {}".format(self.qnetd_addr)
-
+        if utils.service_is_active("corosync-qnetd", self.qnetd_addr):
+            cmd = "corosync-qnetd-tool -l -c {}".format(self.cluster_name)
```
zzhou1, line 257: This adds assumption qnetd server must be running. This might be a little bit too strong requirement?!

One use case comes to my mind. It is to populate certifications without a running qnetd server. This is sounds a valid. To do that, keeping cluster-$ClusterName.crt on qnetd server seems a reasonable approach to replace dynamic detection above.

liangxin1300: Here the dynamic checking by corosync-qnetd-tool  command, I think it's enough for more generic situation, which all related nodes and servers are running during setup process

I already removed the crt file after finished certification process since I think it's might not safe

And even we keep that file to do the static checking, there always has corner case

liangxin1300: BTW, instead of consider all the corner cases, I suggest we make a check when running crm cluster start
That means, when need to start qdevice, after the starting action finished, we check:
If qdevice process started
If qdevice give the vote(by checking corosync-quorumtool) , I found that when have duplicated cluster name, later cluster cannot get the qdevice vote
If that vote in qdevice is 0 or qdevice process not started but qdevice configured, crm cluster start  give warning to user

crmsh/qdevice.py
liangxin1300: It's consistent style for the previous checking information, like:
```
            exception_msg = "Host for qnetd must be a non-cluster node!"
            suggest = "change to another host or stop cluster service on {}".format(self.qnetd_addr)
```
```
            exception_msg = "Host for qnetd must be a non-cluster node!"
            suggest = "change to another host or stop cluster service on {}".format(self.qnetd_addr)
```
The output will be:
```
ERROR: cluster.init: Host for qnetd must be a non-cluster node!
Cluster service already successfully started on this node except qdevice service.
If you still want to use qdevice, change to another host or stop cluster service on hanode2.
Then run command "crm cluster init" with "qdevice" stage, like:
        crm cluster init qdevice qdevice_related_options
That command will setup qdevice separately.
```
```
ERROR: cluster.init: Package "corosync-qnetd" not installed on hanode2!
Cluster service already successfully started on this node except qdevice service.
If you still want to use qdevice, install "corosync-qnetd" on hanode2.
Then run command "crm cluster init" with "qdevice" stage, like:
        crm cluster init qdevice qdevice_related_options
That command will setup qdevice separately.
```

Do you think the suggest information is also unnecessary for previous cases? Just like:
```
ERROR: cluster.init: Host for qnetd must be a non-cluster node!
```
```
ERROR: cluster.init: Package "corosync-qnetd" not installed on hanode2!
```
are good enough?

zzhou1: I like the pattern of "msg/suggestion" very much, especially the suggestion part !  Just in some simple case, msg part is self-explained, then no necessarily to provide the suggestion. 

For the additional two errors:

> ERROR: cluster.init: Package "corosync-qnetd" not installed on hanode2!
The corresponding "suggestion" does make sense to me.

> ERROR: cluster.init: Host for qnetd must be a non-cluster node!
However, this ERROR is too strong in the bigger picture. It does not fit the situation to add HA capability to the qnetd server itself. With that in mind, I do suggest to delete this check, the corresponding ERROR and suggestion.

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
         """
         Validate on qnetd node
         """
-        if utils.check_ssh_passwd_need(self.qnetd_addr):
-            self.askpass = True
-
         exception_msg = ""
         suggest = ""
         if utils.service_is_active("pacemaker", self.qnetd_addr):
```
zzhou1, line 250:
```suggestion
```

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
...
-            exception_msg = "host for qnetd must be a non-cluster node"
+            exception_msg = "Host for qnetd must be a non-cluster node!"
```
zzhou1, line 251:
```suggestion
```

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
...
             suggest = "change to another host or stop cluster service on {}".format(self.qnetd_addr)
```
zzhou1, line 252:
```suggestion
```

crmsh/qdevice.py
```diff
@@ -173,20 +245,24 @@ def valid_qnetd(self):
...
         elif not utils.package_is_installed("corosync-qnetd", self.qnetd_addr):
```
zzhou1, line 253:
```suggestion
        if not utils.package_is_installed("corosync-qnetd", self.qnetd_addr):
```

crmsh/qdevice.py
liangxin1300: Changed

crmsh/qdevice.py
liangxin1300: changed

crmsh/qdevice.py
liangxin1300: changed

crmsh/qdevice.py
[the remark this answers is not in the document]
liangxin1300: Suggest to keep this logic unchanged. Seems no harm here

zzhou1 (approved):

PR #952, opened 2022-03-18
utils: update detect_cloud pattern for aws

yeoldegrove:

PR #953, opened 2022-03-18
[crmsh-4.3] medium: utils: update detect_cloud pattern for aws

liangxin1300:
backport #952

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2017,8 +2017,13 @@ def detect_aws():
     """
     Detect if in AWS
     """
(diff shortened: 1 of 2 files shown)


liangxin1300:
To make it accepted in 15SP2, maybe we need fire a new [L3]bug? @zzhou1

zzhou1:
> To make it accepted in 15SP2, maybe we need fire a new [L3]bug? @zzhou1
We can not create L3 bug by ourselves. However, we can push this to crmsh-4.3 branch, and push to the product when we receive a L3 if the customer explicit ask for it.

zzhou1 (approved):

PR #957, opened 2022-03-28
Improve behave functional test, add /test/run-functional-tests

liangxin1300:
## Motivation
During developing, when specific GH CI case failed, it's not easy to reproduce it locally:

- setup N nodes cluster, and setup passwordless between nodes
- install python3-behave package on one node
- copy crmsh/test/features/  to one node
- run behave <specific_test_file.feature>

It will be more convenient to have an `one-liner` tool to do all above steps, in one step.

## Features of `/test/run-functional-tests`
- Setup N nodes cluster quickly
- Setup N nodes without cluster configured and started
- List existing test cases
- Run specific test case
- Cleanup what already deployed

#### Help info
```
# ./test/run-functional-tests -h
Usage: run-functional-tests [OPTIONS]|[TESTCASE INDEX]
run-functional-tests is a tool for developers to setup the cluster in containers to run functional tests.
The container image is based on Tumbleweed with preinstalled packages of the cluster stack include pacemaker/corosync/crmsh and many others.
Users can make the code change under crmsh.git including test cases. This tool will pick up the code change and "make install" to all running containers.

OPTIONS:
  -h, --help		Show this help message and exit
  -l            	List existing functional test cases and exit
  -n NUM        	Only setup a cluster with NUM nodes(containers)
  -x			Don't config corosync on containers(with -n option)
  -d            	Cleanup the cluster containers

EXAMPLES:
To launch 2 nodes with the running cluster with the very basic corosync.conf
# crmsh.git/test/run-functional-tests -n 2

To launch 2 nodes without the cluster stack running to play with "crm cluster init/join"
# crmsh.git/run-functional-tests -n 2 -x

To list the existing test cases. Users could add his own new test cases.
# crmsh.git/test/run-functional-tests -l

To run a single or a number of functional test cases
# crmsh.git/test/run-functional-tests 1
# crmsh.git/test/run-functional-tests 1 2 3

To clean up the all containers which are generated by this tool
# crmsh.git/test/run-functional-tests -d
```

#### Setup N nodes cluster quickly
```
# ./test/run-functional-tests -n 2
INFO: Loading docker image liangxin1300/hatbw...
INFO: Create ha specific docker network "ha_network_first"...
INFO: Create ha specific docker network "ha_network_second"...
INFO: Setup cluster...
INFO: Deploying "hanode1"...
INFO: Building crmsh codes on "hanode1"...
INFO: Deploying "hanode2"...
INFO: Building crmsh codes on "hanode2"...
INFO: Starting pacemaker...done
INFO: Cluster services started on hanode1
INFO: Starting pacemaker...done
INFO: Cluster services started on hanode2
[root@docker1 crmsh]# docker exec hanode1 crm_mon -1
Cluster Summary:
  * Stack: corosync
  * Current DC: hanode1 (version 2.1.2+20220126.16c1bab10-1.2-2.1.2+20220126.16c1bab10) - partition with quorum
  * Last updated: Sun Mar 27 14:53:39 2022
  * Last change:  Sun Mar 27 14:53:28 2022 by hacluster via crmd on hanode1
  * 2 nodes configured
  * 0 resource instances configured

Node List:
  * Online: [ hanode1 hanode2 ]

Active Resources:
  * No active resources
```

#### Cleanup what already deployed
```
# ./test/run-functional-tests -d
INFO: Cleanup container "hanode2"...
INFO: Cleanup container "hanode1"...
INFO: Cleanup ha specific docker network "ha_network_first"...
INFO: Cleanup ha specific docker network "ha_network_second"...
```

#### Setup N nodes without cluster configured and started
```
# ./test/run-functional-tests -x -n 2
INFO: Loading docker image liangxin1300/hatbw...
INFO: Create ha specific docker network "ha_network_first"...
INFO: Create ha specific docker network "ha_network_second"...
INFO: Setup cluster...
INFO: Deploying "hanode1"...
INFO: Building crmsh codes on "hanode1"...
INFO: Deploying "hanode2"...
INFO: Building crmsh codes on "hanode2"...
```

#### List existing test cases
```
# ./test/run-functional-tests -l
Index|File Name|Description
  1    bootstrap_bugs.feature                    Regression test for bootstrap bugs                         
  2    bootstrap_init_join_remove.feature        crmsh bootstrap process - init, join and remove            
  3    bootstrap_options.feature                 crmsh bootstrap process - options                          
  4    configure_bugs.feature                    Functional test for configure sub level                    
  5    constraints_bugs.feature                  Verify constraints(order/colocation/location) bug          
  6    crm_report_bugs.feature                   crm report functional test for verifying bugs              
  7    geo_setup.feature                         geo cluster                                                
  8    qdevice_options.feature                   corosync qdevice/qnetd options                             
  9    qdevice_setup_remove.feature              corosync qdevice/qnetd setup/remove process                
 10    qdevice_usercase.feature                  Verify usercase master survive when split-brain            
 11    qdevice_validate.feature                  corosync qdevice/qnetd options validate                    
 12    resource_failcount.feature                Use "crm resource failcount" to manage failcounts          
 13    resource_set.feature                      Use "crm configure set" to update attributes and operations
 14    regression test                           Original regression test
```

#### Run test case 1 and 2
```
[root@docker1 crmsh]# ./test/run-functional-tests 1 2
INFO: Loading docker image liangxin1300/hatbw...
INFO: Create ha specific docker network "ha_network_first"...
INFO: Create ha specific docker network "ha_network_second"...
INFO: Setup cluster...
INFO: Deploying "hanode1"...
INFO: Building crmsh codes on "hanode1"...
INFO: Deploying "hanode2"...
INFO: Building crmsh codes on "hanode2"...
INFO: Deploying "hanode3"...
INFO: Building crmsh codes on "hanode3"...
@bootstrap
Feature: Regression test for bootstrap bugs<font color="#555753"> # app/test/features/bootstrap_bugs.feature:2</font>
  Tag @clean means need to stop cluster service if the service is available
  Need nodes: hanode1 hanode2 hanode3
  @clean
  Scenario: Set placement-strategy value as &quot;default&quot;(bsc#1129462) 
  ...
  ...
  ...
```


liangxin1300:
Hi @gao-yan @aleksei-burlakov 
Could please take a look at this PR, and give me some feedback/suggestions?

Thanks!

data-manifest
```diff
@@ -59,11 +59,9 @@ test/cibtests/004.exp.xml
 test/cibtests/004.input
 test/cib-tests.sh
 test/cibtests/shadow.base
-test/containerized-regression-tests.sh
```
aleksei-burlakov, line 62: It's more a niggle, the test/containerized-regression-tests.sh, test/docker_scripts.sh was removed in another commit.

Dockerfile
```diff
@@ -1,33 +1,22 @@
-FROM opensuse/leap:15.2
+FROM opensuse/tumbleweed
```
aleksei-burlakov, line 1: IMHO, I would stick to a certain release

test/run-functional-tests
```diff
@@ -0,0 +1,400 @@
+#!/bin/bash
+
+DOCKER_IMAGE="liangxin1300/hatbw"
```
aleksei-burlakov, line 3: May we make it overridable `DOCKER_IMAGE=${DOCKER_IMAGE:-"liangxin1300/hatbw"}`

aleksei-burlakov:
I really like the PR! I have a small question. I have built my own image 
$ doker build . -t crmsh-test
And if I use this image, the script asks me for the password
```
$ DOCKER_IMAGE=crmsh-test ./run-functional-tests -n 2
WARNING: docker cgroup driver suggest to be "systemd"
INFO: Loading docker image crmsh-test...
INFO: Create ha specific docker network "ha_network_first"...
INFO: Create ha specific docker network "ha_network_second"...
INFO: Setup cluster...
INFO: Deploying "hanode1"...
INFO: Building crmsh codes on "hanode1"...
INFO: Deploying "hanode2"...
INFO: Building crmsh codes on "hanode2"...
(root@hanode2) Password:
```
The default image however works well. There is possibly something missing in the Dockerfile.

test/run-functional-tests
liangxin1300: Nice!

aleksei-burlakov:
> > I really like the PR! I have a small question. I have built my own image $ doker build . -t crmsh-test And if I use this image, the script asks me for the password
> > ```
> > $ DOCKER_IMAGE=crmsh-test ./run-functional-tests -n 2
> > ```
> 
> That is a good idea!:)
> 
> > WARNING: docker cgroup driver suggest to be "systemd"
> > INFO: Loading docker image crmsh-test...
> > INFO: Create ha specific docker network "ha_network_first"...
> > INFO: Create ha specific docker network "ha_network_second"...
> > INFO: Setup cluster...
> > INFO: Deploying "hanode1"...
> > INFO: Building crmsh codes on "hanode1"...
> > INFO: Deploying "hanode2"...
> > INFO: Building crmsh codes on "hanode2"...
> > (root@hanode2) Password:
> > ```
> > 
> > The default image however works well. There is possibly something missing in the Dockerfile.
> > ```
> 
> I put the ssh key from building host into image, the details in Dockerfile:
> 
> ```
> ARG ssh_prv_key
> ARG ssh_pub_key
> # docker build -t hatbw --build-arg ssh_prv_key="$(cat /root/.ssh/id_rsa)" --build-arg ssh_pub_key="$(cat /root/.ssh/id_rsa.pub)" .
> ```
> 
> And
> 
> ```
> RUN mkdir -p /root/.ssh && chmod 0700 /root/.ssh
> RUN echo "$ssh_prv_key" > /root/.ssh/id_rsa && chmod 600 /root/.ssh/id_rsa
> RUN echo "$ssh_pub_key" > /root/.ssh/id_rsa.pub && chmod 600 /root/.ssh/id_rsa.pub
> RUN echo "$ssh_pub_key" > /root/.ssh/authorized_keys && chmod 600 /root/.ssh/authorized_keys
> ```
> 
> So all nodes from this image will share the same ssh key

Aha, I see 👍

Dockerfile
liangxin1300: I think when we created branch crmsh-4.4, then we change the image based on leap:15.4, that seems make more sense

test/run-functional-tests
```diff
@@ -0,0 +1,419 @@
...
+	cmd=$2
+	docker exec -t $name /bin/sh -c "$cmd"
+}
+
+
+deploy_ha_node() {
+	node_name=$1
+	project_path=$(dirname $(dirname `realpath $0`))
+	docker_options="-d --name $node_name -h $node_name --privileged --shm-size 1g -v $project_path:/app"
+        build_cmd="cd /app;./autogen.sh && ./configure --prefix /usr && make install && make install-crmconfDATA prefix="
```
zzhou1, line 185: I have impression to suggest to move "./autogen.sh && ./configure" outside of the container execution environment. Only need in the host environment, let run-functional-tests prepares it once. It's a waste for 10+ of containers to repeat it.

test/run-functional-tests
```diff
@@ -0,0 +1,419 @@
...
+
+	info "Deploying \"$node_name\"..."
+	docker run $docker_options $DOCKER_IMAGE &> /dev/null
```
zzhou1, line 188: In a fresh rebooted machine, docker containers already deployed by "run-functional-test" might fail to start. "docker start" is expected instead of "docker run". So, additional detection and handling are expected.

test/run-functional-tests
```diff
@@ -0,0 +1,419 @@
...
+	# check if docker.service started
+	systemctl is-active $DOCKER_SERVICE &> /dev/null
+	if [ "$?" -ne 0 ];then
+		fatal "$DOCKER_SERVICE is not active"
+	fi
+	# check if docker cgroup driver is systemd
+	docker info 2> /dev/null|grep -q "Cgroup Driver: systemd"
+	if [ "$?" -ne 0 ];then
+		warning "docker cgroup driver suggest to be \"systemd\""
+	fi
```
zzhou1, line 106: And the further check is expected to detect if the host is the cgroup2 environment. That says, this only works for cgroup2 so far. And, currently, it works for Tumbelweed and 15.4+ host. This should return the fatal error if try on 15.3, for example.

zzhou1 (changes requested):

test/run-functional-tests
liangxin1300: I've already tried this on 15SP3, basically, it works well, although not sure which test case might failed on it, that's the reason we make this warning:
```
WARNING: docker cgroup driver suggest to be "systemd"
```

test/run-functional-tests
liangxin1300: That seems require host must install `autoconf`

test/run-functional-tests
zzhou1: I read the misleading here. My intention was not for the code to detect "Cgroup Driver". That's fine.

I should put the comment in the beginning of check_docker_env(). The point is to clarify with the user currently this tool does not support cgroup v1 host, since it need --hostns stuff as for a different `docker_options` later on.

test/run-functional-tests
liangxin1300: Changed

run `./autogen.sh && ./configure --prefix /usr` on host before create container

test/run-functional-tests
liangxin1300: Changed

run ./autogen.sh && ./configure --prefix /usr on host before create container

test/run-functional-tests
liangxin1300: Changed as `docker run --restart always`, and run `crm cluster enable` before running `crm cluster start`
So after rebooted, the containers will started, too

test/run-functional-tests
liangxin1300: Anyway, most of functionalities seems work now even on Cgroup Driver
It make sense here just warning not error and exit

test/run-functional-tests
```diff
@@ -0,0 +1,430 @@
...
+	if [ "$node_name" == "node-without-ssh" ];then
+		docker_exec $node_name "systemctl stop sshd.service"
+	fi
+	if [ "$node_name" != "qnetd-node" ];then
+		docker_exec $node_name "rpm -e corosync-qnetd"
+	fi
+	docker_exec $node_name "rm -rf /run/nologin"
+	docker_exec $node_name "echo 'StrictHostKeyChecking no' >> /etc/ssh/ssh_config"
+	info "Building crmsh codes on \"$node_name\"..."
+	docker_exec $node_name "$build_cmd" 1> /dev/null || \
```
aleksei-burlakov, line 213: I can't help seeing some magic, when run `$ ./run-functional-tests -n 2` it shoots the "Building failed!", even that there was no error in the `docker_exec $node_name "$build_cmd"` command. And when creating only 1 node cluster `$ ./run-functional-tests -n 1` it works like charm. @liangxin1300 have you encountered such behavior?

test/run-functional-tests
zzhou1: Yeah,not much we can change GitHub host

liangxin1300: No, we can install packages on it, but:
- still be mess after install autoconf and automake, need more investigations
- will make GH CI slower for each test case

test/run-functional-tests
zzhou1: I see the misreading still.

Here is the sample code in my head
```
[ $(findmnt /sys/fs/cgroup -no FSTYPE) == "cgroup2" ] ||  { echo Exit. cgroup2 is not detected; exit; }
```

zzhou1: It looks I have some missing docker knowledge between cgroup v1 and v2 in the different distro. I agree to leave it as it is for the time being, and revisit when it gets crystal clear.

zzhou1 (approved):

PR #966, opened 2022-05-05
replace csync2

liangxin1300:


crmsh/bootstrap.py
```diff
@@ -828,32 +829,46 @@ def append_to_remote_file(fromfile, remote_node, tofile):
...
-    """.format(utils.this_node(), csync2_file_list), CSYNC2_CFG)
+        """.format(me, csync2_file_list), CSYNC2_CFG)
 
+    logger.info("Enable and start csync2.socket")
     utils.start_service("csync2.socket", enable=True)
-    with logger_utils.status_long("csync2 checking files"):
-        invoke("csync2", "-cr", "/")
+
+    if _context.type == "join":
+        logger.warning("To initially synchronize all files configured by csync2, please run \"crm cluster init csync2\" on suitable time(this may take a while)")
```
zzhou1, line 871:
```suggestion
        logger.warning("csync2 database is not initiated yet. Before using csync2 for the first time, please run \"crm cluster init csync2\" on any one node. Note, this may take a while.")
```

liangxin1300:
Maybe consider the other solution:
- default behavior is doing csync2 sync
- add an option like "-S" on join side to skip this action

Issue #970, opened 2022-05-09
WARNING: crmadmin -S <HOST> unexpected output


freyes:
Pacemaker changed the output string of "crmadmin -S <HOST>" in 2.1.0 with the commit https://github.com/ClusterLabs/pacemaker/commit/c26c9951d863e83126f811ee5b91a174fe0cc991 , this is different from the ouput that wait4dc() expects https://github.com/ClusterLabs/crmsh/blob/master/crmsh/utils.py#L898 .

Example output of `crm -w -F node maintenance <HOST>` of a cluster running on Ubuntu 22.04

```
root@juju-0c8f53-zaza-723eab24403d-4:~# crm -w -F node maintenance juju-0c8f53-zaza-723eab24403d-4
WARNING: crmadmin -S juju-0c8f53-zaza-723eab24403d-5 unexpected output: Controller on juju-0c8f53-zaza-723eab24403d-5 in state S_IDLE: ok (exit code: 0)
root@juju-0c8f53-zaza-723eab24403d-4:~# echo $?
1
root@juju-0c8f53-zaza-723eab24403d-4:~# crmadmin -S juju-0c8f53-zaza-723eab24403d-5
Controller on juju-0c8f53-zaza-723eab24403d-5 in state S_IDLE: ok
root@juju-0c8f53-zaza-723eab24403d-4:~# dpkg -l pacemaker crmsh | cat
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version        Architecture Description
+++-==============-==============-============-===========================================
ii  crmsh          4.3.1-1ubuntu2 all          CRM shell for the pacemaker cluster manager
ii  pacemaker      2.1.2-1ubuntu3 amd64        cluster resource manager
```

liangxin1300:
Thanks for pointing this!
I will take a look

liangxin1300:
@freyes The fix for branch crmsh-4.3 is in #972
BTW, I highly suggest to use crmsh 4.4.0 or master branch with pacemaker 2.1+

freyes:
> @freyes The fix for branch crmsh-4.3 is in #972 
Thanks, much appreciated.

>BTW, I highly suggest to use crmsh 4.4.0 or master branch with pacemaker 2.1+
ack

[...]

PR #972, opened 2022-05-11
[crmsh-4.3] Fix: utils: wait4dc: Make change since output of 'crmadmin -S' change

liangxin1300:
Backport from #971, to fix issue #970

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -888,16 +888,10 @@ def wait4dc(what="", show_progress=True):
             return False
         cmd = "crmadmin -S %s" % dc
         rc, s = get_stdout(add_sudo(cmd))
-        if not s.startswith("Status"):
-            common_warn("%s unexpected output: %s (exit code: %d)" %
-                        (cmd, s, rc))
+        if rc != 0:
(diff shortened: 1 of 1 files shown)


zzhou1 (approved):

freyes:
hi all, I was able to test this change successfully on a ubuntu 22.04 system using the testing package available at https://launchpad.net/~freyes/+archive/ubuntu/lp1972730 .

Thanks for the patch.

tverdoff:
Hi @liangxin1300 

I have trapped into the same issue, updated to crmsh 4.3.0 but it appears that fix went to crmsh-4.3 branch at the moment when 4.3.1 had already been tagged. At the same time 4.3.2 tag is not set, please clarify when the 4.3.2 version will be issued. The issue is quite critical because resources do not start after restart command:

```
crm resource restart <resource-id>
INFO: ordering vip-10.0.2.20 to stop
WARNING: crmadmin -S configurator unexpected output: Controller on configurator in state S_TRANSITION_ENGINE: ok (exit code: 0)

```

the fix helps, but not included in RPM.

liangxin1300:
@tverdoff Done for tag 4.3.2

PR #973, opened 2022-05-18
Return earlier in `crm cluster join` if any cluster service is active

### LINKED ISSUE #808: `crm cluster join` should not start if any cluster service is running
`crm cluster join` should have prerequisites, and not start in the very beginning if any cluster service is running, eg. sbd, corosync,
pacemaker, corosync-qdevice

xxyzz:
Resolves #808

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -428,6 +428,10 @@ If stage is not specified, each stage will be invoked in sequence.
         options, args = parse_options(parser, args)
         if options is None or args is None:
             return
+        for service in ["sbd", "corosync", "pacemaker", "corosync-qdevice"]:
+            if utils.service_is_active(service):
+                logger.warning("'{}' service must be stopped to proceed.".format(service))
+                return
 
         stage = ""
         if len(args) == 1:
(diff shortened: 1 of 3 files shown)


crmsh/ui_cluster.py
```diff
@@ -428,6 +428,8 @@ def do_join(self, context, *args):
         options, args = parse_options(parser, args)
         if options is None or args is None:
             return
+        if any([utils.service_is_active(service) for service in ["sbd", "corosync", "pacemaker", "corosync-qdevice"]]):
+            return
```
zzhou1, line 432: For the better user experience, before returning, to print a message up front is useful for the system admin to take the proper action(s) to react on it. Something like,

```suggestion
            logger.warning(" '{}' service must stop to proceed.".format(service)
            return False
```

zzhou1 (changes requested):
Nice engagement, thanks!

crmsh/ui_cluster.py
xxyzz: Thanks!

zzhou1 (approved):

Issue #819, opened 2021-05-28
wishlist: filter `crm configure show` by type of the primitive


klaernie:
it would be really neat if `crm configure show` could be filtered by the type of the primitive, so that one could for example filter down to only the IPaddr2 resources.

My specific use case is scanning multiple machines to keep an inventory database up to date, where it would be nice to filter out which IPs are used by the clusters.

Right now I use the following one liner to grab the information needed, but this isn't the most practical to maintain:
```sh
env PAGER=cat crm configure show | perl -pe 's/\\\n//' | perl -ne 'if(m/prm_([^\s]+)_IP\sIPaddr2.*ip=([\d.]+).*cidr_netmask=(\d+)/){ print "$1|$2/$3\n" }'
```

I realize that this is a pure wishlist item, maybe it's a good first contribution for someone to get their feet wet.

liangxin1300:
Hi @klaernie ,
Currently crmsh support `crm configure show type:primitive`, is that what you want?

klaernie:
Hi @liangxin1300 

That helps partially, but of cause still lists more resources than in my case the IP addresses. Most of the clusters I run scans for have 5-10 LVM volume groups and logical volumes, 2-3 application resources and only one IPaddr2 resource.

liangxin1300:
> Hi @liangxin1300
> 
> That helps partially, but of cause still lists more resources than in my case the IP addresses. Most of the clusters I run scans for have 5-10 LVM volume groups and logical volumes, 2-3 application resources and only one IPaddr2 resource.

Seems like this issue https://github.com/ClusterLabs/crmsh/issues/698
I will put this into my backlog 
Thank you @klaernie !

Issue #698, opened 2020-12-17
enhancement: configuration search


sjvudp:
In crm shell `configure` context you can

- `edit` items specifying their name or a pattern that occurs in the resource name
- `show` items specifying their name or a pattern that occurs in the resource name

Given a highly consistent naming pattern for resources that would be fine to easily select *related* resources that way.
However the real world is not perfect and sometimes one wants to select resources containing some (more or less) magic string.

Therefore I suggest an enhancement that will allow matching not only based on the resource name, but also on the resource content. While I don't want to propose a specific syntax really, I'm sketching how it *could* be done:

"`crm(...) configure# grep `*pattern*" could list the names of all resources where either the name matches pattern, or where the text representation matches pattern (a possible extension might be "`grep xml` *pattern*" where the XML would be considered instead).

To combine such search result with commands easily, a set of additional commands can be imagined:

- "`show` [`all`] `matching` *pattern*": Display all resources matching (as described above) pattern. The `xml` extension would be possible, too (match XML and display XML)
- "`edit` [`all`] `matching` *pattern*": Edit all resources (in a single editor) matching pattern (as described above). The `xml` extension would be possible, too (match XML and display XML)
- "`delete` [`all`] `matching` *pattern*": Deleting all resources matching pattern (as described above) is probably a bad idea (thinking of `order` and `colocation` (among others))
- `filter` [`all`] `matching` *pattern*": ...restrict resources fed into `filter`...
- ... (possibly more)

(`all` is just some syntactic fill-word ("sugar") with no change in semantics, meaning it *does not* "match everything")

sjvudp:
Another thinkable option would be combining the restriction `changed` with the new feature, so that the searching would be restricted to any items that were changed (directly or indirectly through dependencies). Unsure about the syntax. A new keyword `matching-changed` (`matchingchanged`)?

liangxin1300:
Thanks for the suggestions!

zzhou1:
Yeah, I like `crm configure show related:*foo* related:*bar*`. But, it stays within "the resource name" makes the usage limited. Example, I may want to show all resources based on ocf:hearbeat:Filesystem.

I hesitate to add yet another subcommand `crm configure grep`, since the subcommand list under `crm configure` is pretty long already. With that, I incline to improve the subcommands like  "show | edit | delete ..."

PR #978, opened 2022-06-02
Dev: cinconfig: enable "related:" prefix to show the objects by given ra type

liangxin1300:
```
crm(live/tbw-1)configure# show related:Dummy
primitive d1 Dummy
primitive d2 Dummy
primitive d3 ocf:pacemaker:Dummy
crm(live/tbw-1)configure# show related:IPaddr2
primitive vip1 IPaddr2 \
        params ip=10.10.10.123 \
        op monitor interval=10s
primitive vip2 IPaddr2 \
        params ip=10.10.10.124 \
        op monitor interval=10s
crm(live/tbw-1)configure# show related:ocf:heartbeat:Dummy
primitive d1 Dummy
primitive d2 Dummy
```

Should be part of requirements from #698 and #819 

Others:
- [x] Complete the help info

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -3135,6 +3135,31 @@ class CibFactory(object):
         return [x for x in self.cib_objects
                 if x.updated or x.origin == "user"]
 
+    def get_elems_on_ra_type(self, spec):
+        """
+        Get elements by given ra class:provider:type or class:type or only type
+        return [] if no results
+        """
+        content_list = spec.split(':')[1:]
+        if len(content_list) > 3:
+            return []
+        if len(content_list) == 3:
+            cls, provider, tp = content_list
+            return [x for x in self.cib_objects 
+                    if x.node.get("type") == tp
+                    and x.node.get("provider") == provider
+                    and x.node.get("class") == cls]
+        if len(content_list) == 2:
+            cls, tp = content_list
+            return [x for x in self.cib_objects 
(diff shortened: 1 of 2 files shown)


klaernie:
That looks cool! Can't wait to see this in production! Thanks, @liangxin1300 !

zzhou1:
> Hi @klaernie, @zzhou1, @sjvudp
> 
> Could you please take a look at this?
> 

Yes, I like this functionality in general! 

After revisiting this, I have some impression for the better user experience, if possible:

a) first of all, try to avoid adding yet another "ra_type" to the user interface. crmsh user interface is big enough already, and maybe even too big to be consumed by the end user efficiently.

b) with that, I propose to extend this functionality to the existing 'related' prefix, 
    ie, `crm configure show related:Dummy`

liangxin1300:
> > Hi @klaernie, @zzhou1, @sjvudp
> > Could you please take a look at this?
> 
> Yes, I like this functionality in general!
> 
> After revisiting this, I have some impression for the better user experience, if possible:
> 
> a) first of all, try to avoid adding yet another "ra_type" to the user interface. crmsh user interface is big enough already, and maybe even too big to be consumed by the end user efficiently.
> 
> b) with that, I propose to extend this functionality to the existing 'related' prefix, ie, `crm configure show related:Dummy`

Nice suggestion
Changed

Thanks!

zzhou1 (approved):

Issue #983, opened 2022-06-15
Argument list too long: '/bin/sh' 


jgrund:
Relevant code:

https://github.com/ClusterLabs/crmsh/blob/d0150a8874d800caf7abb57c8f17e1a73f9dd0a6/crmsh/ui_node.py#L341

When trying to run `crm node standby` I see the following error:

```
[root@node7 ~]#  crm --debug node standby
2022/06/15 17:35:23
DEBUG: pacemaker version: [err: ][out: CRM Version: 2.1.0-8.el8 (7c3f660707)]
DEBUG: found pacemaker version: 2.1.0-8.el8
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/crmsh/ui_context.py", line 92, in run
    rv = self.execute_command() is not False
  File "/usr/lib/python3.6/site-packages/crmsh/ui_context.py", line 284, in execute_command
    rv = self.command_info.function(*arglist)
  File "/usr/lib/python3.6/site-packages/crmsh/ui_node.py", line 341, in do_standby
    rc = utils.diff_and_patch(xmlutil.xml_tostring(orig_cib), xmlutil.xml_tostring(cib))
  File "/usr/lib/python3.6/site-packages/crmsh/utils.py", line 3079, in diff_and_patch
    rc, cib_diff, err = get_stdout_stderr(cmd)
  File "/usr/lib/python3.6/site-packages/crmsh/utils.py", line 812, in get_stdout_stderr
    stderr=subprocess.PIPE)
  File "/usr/lib64/python3.6/subprocess.py", line 729, in __init__
    restore_signals, start_new_session)
  File "/usr/lib64/python3.6/subprocess.py", line 1364, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 7] Argument list too long: '/bin/sh'
ERROR: node.standby: [Errno 7] Argument list too long: '/bin/sh'
```

What appears to be happening is the cib being printed is too long and overflows `MAX_ARG_STRLEN`

jgrund:
Looks like the argument issue is fixed. However, `crm node standby` returned a 0 exit code on a node but the node never moved into standby. This happened on 1/4 nodes tested. Not sure if it's related to this change or not.

liangxin1300:
[...]

PR #984, opened 2022-06-16
Fix: utils: use -o and -n to compare files instead of strings for crm_diff

liangxin1300:


ismaell:
Hi @liangxin1300, would you mind adding a reference tag to #983 (for better tracking) on the commit itself like:

    Fixes: <https://github.com/ClusterLabs/crmsh/issues/983>

or `Ref: ...`.

And also mention the issue it fixes in the message as well. It really helps the maintenance workflow.

ismaell (approved):
LGTM. Also tested.

zzhou1 (approved):

Issue #977, opened 2022-05-30
crmsh not working when using ACL


heinervdm:
I'm having multiple users with different access rights in my cluster. Therefore all my users are in the haclient group (RHEL 8) and I've definded several ACL entries in my cluster configuration.

When using crmsh I'm getting a PermissionError. Version 4.2.0 was working fine.

`crm configure show
Traceback (most recent call last):
  File "/usr/sbin/crm", line 31, in <module>
    log.setup_logging()
  File "/usr/lib/python3.6/site-packages/crmsh/log.py", line 445, in setup_logging
    shutil.chown(CRMSH_LOG_FILE, constants.HA_USER, constants.HA_GROUP)
  File "/usr/lib64/python3.6/shutil.py", line 1052, in chown
    os.chown(path, _user, _group)
PermissionError: [Errno 1] Operation not permitted: '/var/log/crmsh/crmsh.log'`

From the Backtrace, I assume that crmsh, wants to change the file owner after writing to the file. But without root permissions you are not allowed to change the owner to someone else.
What should be done there is: chgrp instead of chown.

liangxin1300:
Hi @heinervdm 
Could you please show me the reproduce steps?

Thanks!

heinervdm:
Hi @liangxin1300,
I've created a new Testinstallation with RockyLinux 8 (yum install --enablerepo=ha corosync pacemaker)
and installed crmsh.

I have the following versions installed:
`$corosync -v
Corosync Cluster Engine, version '3.1.5'
Copyright (c) 2006-2021 Red Hat, Inc.

Built-in features: dbus systemd xmlconf vqsim nozzle snmp pie relro bindnow
Available crypto models: nss openssl
Available compression models: zlib lz4 lz4hc lzo2 lzma bzip2

$pacemaker --version
Pacemaker 2.1.2-4.el8
Written by Andrew Beekhof

$crm --version
crm 4.4.0`

I've created a testuser and added it to the haclient group:
`useradd -m -G haclient hatest`

I'm having the following dummy crm configuration:
`$crm configure show
node 1: node1
node 2: node2
primitive ip IPaddr2 \
	params ip=10.2.0.10 cidr_netmask=24
property cib-bootstrap-options: \
	cluster-name=test \
	stonith-enabled=false \
	enable-acl=true \
	no-quorum-policy=ignore
role admin \
	write cib
acl_target hatest \
	admin`

When I then run crm configure show as hatest user I'm getting the following error:
[...]

PR #991, opened 2022-07-07
Fix: #977 crmsh not working when using ACL

heinervdm:
Set file user of logfile to current user und HA_GROUP, as changing the user to someone else is only allowed for root.
Set umask to 002, to make log file group writable.
Also add a GroupWriteRotatingFileHandler, taken from https://stackoverflow.com/a/6779307 to keep permissions when log file is rotated.
I couldn't test the rotation, as my logfile was always empty...

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -102,6 +102,18 @@ class DebugCustomFilter(logging.Filter):
             return True
 
 
+class GroupWriteRotatingFileHandler(logging.handlers.RotatingFileHandler):
+    """
+    A custom rotating file handler which keeps log files group wirtable after rotating
+    Source: https://stackoverflow.com/a/6779307
+    """
+    def _open(self):
+        prevumask=os.umask(0o002)
+        rtv=logging.handlers.RotatingFileHandler._open(self)
+        os.umask(prevumask)
+        return rtv
+
+
 LOGGING_CFG = {
     "version": 1,
     "disable_existing_loggers": "False",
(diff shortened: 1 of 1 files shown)


liangxin1300:
Hi @heinervdm 
Sorry for late!

I'm trying to use your code by these steps
- Under root, run `crm cluster init -y` to setup a cluster
- Under root, run `crm configure show`
- Under root, run `useradd -m -G haclient hatest`
- Run `su - hatest`
- Under hatest, run `/usr/sbin/crm cluster show`, I got
```
  File "/usr/sbin/crm", line 31, in <module>
    log.setup_logging()
  File "/usr/lib/python3.8/site-packages/crmsh/log.py", line 458, in setup_logging
    shutil.chown(CRMSH_LOG_FILE, os.getuid(), constants.HA_GROUP)
  File "/usr/lib64/python3.8/shutil.py", line 1310, in chown
    os.chown(path, _user, _group)
PermissionError: [Errno 1] Operation not permitted: '/var/log/crmsh/crmsh.log'
```

crmsh/log.py
```diff
@@ -105,6 +105,18 @@ def filter(self, record):
             return True
 
 
+class GroupWriteRotatingFileHandler(logging.handlers.RotatingFileHandler):
+    """
+    A custom rotating file handler which keeps log files group wirtable after rotating
```
SergeCroise, line 110: writable

SergeCroise, line 110:
```suggestion
    A custom rotating file handler which keeps log files group writable after rotating
```

nicholasyang2022:
Thank you for your contribution. Your commit has been cherry-picked to #1040 and merged to master.

PR #998, opened 2022-08-05
Dev: bootstrap: Don't sync csync2 when peer node's csync2 service not ready

liangxin1300:
This will cause `ERROR: Connection to remote host <nodename> failed.` to log.

And this is introduced since 96b5f8419, to avoid sync failed if configured files changed;
96b5f8419 should not be a problem anymore since e098387e2


nicholasyang2022:
Looks good. `csync2_update` will be called in `join_cluster` and files get synced.

But when will `init_csync2_remote` be called? I cannot find any usage of it.

liangxin1300:
> Looks good. `csync2_update` will be called in `join_cluster` and files get synced.
> 
> But when will `init_csync2_remote` be called? I cannot find any usage of it.

In https://github.com/ClusterLabs/crmsh/blob/master/crmsh/bootstrap.py#L1467
Join node will call `csync2_remote` stage on init node, then on init node, it will call `init_csync2_remote`

nicholasyang2022 (approved):

Issue #720, opened 2021-01-11
Multiple " target-role=Stopped"


liangxin1300:
When resource has multiple meta entries, multiple `target-role=Stopped` will be added while stopping the resource:
```
primitive d Dummy \
        meta priority=123 target-role=Stopped \
        meta resource-stickiness=0 allow-migrate=true target-role=Stopped \
        op monitor interval=10s
```
@zzhou1 @gao-yan Do you think this is a bug?

zzhou1:
From the conceptual level, I do read it as a cusmetic issue. I found another cusmetic one as below very recently. No sure if the root cause is same, at least it is similiar.

```
sle15sp2-2:~ # crm config show p_fs
primitive p_fs Filesystem \
        op monitor interval=30s \
        op_params OCF_CHECK_LEVEL=20 \
        params device="/dev/vg1/lv1" directory="/srv/nfs" fstype=xfs
sle15sp2-2:~ # crm resource trace p_fs
INFO: Trace for p_fs is written to /var/lib/heartbeat/trace_ra/
INFO: Trace set, restart p_fs to trace non-monitor operations
sle15sp2-2:~ # crm config show p_fs
primitive p_fs Filesystem \
        op start interval=0 \
        op_params trace_ra=1 \
        op stop interval=0 \
        op_params trace_ra=1 \
        op monitor interval=30s \
        op_params OCF_CHECK_LEVEL=20 trace_ra=1 \
        params device="/dev/vg1/lv1" directory="/srv/nfs" fstype=xfs
sle15sp2-2:~ # crm resource untrace p_fs
sle15sp2-2:~ # crm config show p_fs
primitive p_fs Filesystem \
        op monitor interval=30s \
        op_params OCF_CHECK_LEVEL=20 \
        params device="/dev/vg1/lv1" directory="/srv/nfs" fstype=xfs
```

liangxin1300:
> From the conceptual level, I do read it as a cusmetic issue. I found another cusmetic one as below very recently. No sure if the root cause is same, at least it is similiar.
> 
> ```
> sle15sp2-2:~ # crm config show p_fs
> primitive p_fs Filesystem \
>         op monitor interval=30s \
>         op_params OCF_CHECK_LEVEL=20 \
>         params device="/dev/vg1/lv1" directory="/srv/nfs" fstype=xfs
> sle15sp2-2:~ # crm resource trace p_fs
[...]

PR #1008, opened 2022-08-15
Fix: ui_resource: Avoid adding multi meta attributes

liangxin1300:
Like avoid adding multi target-role enties for each meta instance
To fix upstream issue https://github.com/ClusterLabs/crmsh/issues/720

backport #979 

Before change
```
# After run `resource stop d2` and `resource manage d2`
node 1: 15sp4-1
node 2: 15sp4-2
primitive d2 Dummy \
        meta description=d2 target-role=Stopped is-managed=true \
        meta priority=3 target-role=Stopped is-managed=true \
```
After changed
```
# run `resource start d2` and `resource unmanage d2`, still can change existing multi attribute value
# For the newly added resource d3, only set one entry for meta attributes, after run `resource stop d3` and `resource manage d3`
node 1: 15sp4-1
node 2: 15sp4-2
primitive d2 Dummy \
        meta description=d2 target-role=Started is-managed=false \
        meta priority=3 target-role=Started is-managed=false \
        op monitor timeout=20s interval=10s \
        op stop timeout=20s interval=0s
primitive d3 Dummy \
        meta description=d2 target-role=Stopped is-managed=true \
        meta priority=3
```

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -98,8 +98,8 @@ def set_deep_meta_attr_node(target_node, attr, value):
         for nvpair in nvpairs:
             nvpair.set("value", value)
     else:
-        for n in xmlutil.get_set_nodes(target_node, "meta_attributes", create=True):
-            xmlutil.set_attr(n, attr, value)
+        nodes = xmlutil.get_set_nodes(target_node, "meta_attributes", create=True)
+        xmlutil.set_attr(nodes[0], attr, value)
     return True
 
 


liangxin1300:
Hi @gao-yan
Please take a look at this pr

Thanks!

crmsh/ui_resource.py
```diff
@@ -98,8 +98,8 @@ def set_deep_meta_attr_node(target_node, attr, value):
         for nvpair in nvpairs:
             nvpair.set("value", value)
     else:
-        for n in xmlutil.get_set_nodes(target_node, "meta_attributes", create=True):
-            xmlutil.set_attr(n, attr, value)
+        nodes = xmlutil.get_set_nodes(target_node, "meta_attributes", create=True)
+        xmlutil.set_attr(nodes[0], attr, value)
```
gao-yan, line 102: Of course, purely from crmsh, we should avoid creation of multiple attributes sets first of all.

But given that already existing multiple attributes sets, the old way is probably the best we could do... Since the multiple attributes sets might have been intentionally configured by users together with different time-based rules, so the first set is not necessarily the one that applies to the expected time/need unless we could understand the contexts and the rules...

gao-yan: Indeed. Okay.

PR #1012, opened 2022-08-18
Dev: log: Record msg during status_long and release it afterward

liangxin1300:
The console format will mass if there are messages during status_long:
```
INFO: Waiting for cluster...INFO: This the demo info
...........done
```
After applied this PR:
```
INFO: Waiting for cluster..............done
INFO: This the demo info
```

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -13,6 +13,8 @@ from . import constants
 
 
 CRMSH_LOG_FILE = "/var/log/crmsh/crmsh.log"
+STATUS_LONG_PREFIX = "%s..."
+BUFFER_MSG_LIST = []
 
 
 class ConsoleCustomHandler(logging.StreamHandler):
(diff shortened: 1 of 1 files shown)


nicholasyang2022:
This change seems problematic.

1. It changes the order in which logs get printed.
2. It buffers logs in memory. If program get stuck during `status_long`, buffered messages will not be printed.
3. It pass data between logger and handler via global variable, breaking the layered design of `logging`.

It makes the log inaccuracy and might cause difficulties debugging the program.

liangxin1300:
Thanks for review!
> This change seems problematic.
> 
>     1. It changes the order in which logs get printed.
Yes, the info and warn logs will be printed after status_long, not during status_long, to avoid the mess format
But the error logs will be printed in order
Consider info and warn logs are just under and close to the status_long message, and logs in file are always in order
I think this is acceptable and easy to understand for customer

>     2. It buffers logs in memory. If program get stuck during `status_long`, buffered messages will not be printed.
The buffered messages should also be printed if any exception raised, yes, I miss this

>     3. It pass data between logger and handler via global variable, breaking the layered design of `logging`.
I see what you mean. I will try to find a more graceful way to handle this:)

Anyway, this PR's priority is low
> 
> It makes the log inaccuracy and might cause difficulties debugging the program.

liangxin1300:
Or, @zzhou1, maybe we can directly drop the `done` message, like:
```
INFO: Waiting for cluster..............
WARNING: other warnings
INFO: other messages
```
I think if the new logs printed after `...` without any errors, that already means done successfully
Then this will be the most simple solution

What do you think?

PR #1013, opened 2022-08-19
fix: configure: refresh cib before showing or modifying if no pending changes has been made (bsc#1202465)

nicholasyang2022:
cib_factory is cached in interactive `crm configure`. Refreshing before showing and modifying ensures concurrent commits to cib visible.

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -3254,6 +3254,10 @@ class CibFactory(object):
             return False
         return self.modified_elems() or self.remove_queue
 
+    def ensure_cib_updated(self):
+        if options.interactive and not self.has_cib_changed():
+            self.refresh()
+
     def _verify_constraints(self, node):
         '''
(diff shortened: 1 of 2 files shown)


crmsh/cibconfig.py
```diff
@@ -3254,6 +3254,10 @@ def has_cib_changed(self):
             return False
         return self.modified_elems() or self.remove_queue
 
+    def ensure_cib_updated(self):
+        if not self.has_cib_changed():
```
liangxin1300, line 3258: I suggest to add another check, like
`if options.interactive and not self.has_cib_changed():`
Since it has no such issue under non-interactive mode:)

liangxin1300:
And please backport it to branch crmsh-4.4 for 15sp4, should be easy to backport

liangxin1300 (approved):
Good work!
Thanks!

PR #1017, opened 2022-08-22
Dev: bootstrap: Adjust cluster properties including priority-fencing-delay

liangxin1300:
When join/remove node, add/remove qdevice, add sbd via stage,
Set property priority=1 for 2 nodes cluster without qdevice
Set priority-fencing-delay=2*pcmk_delay_max


crmsh/bootstrap.py
```diff
@@ -1844,6 +1844,39 @@ def update_nodeid(nodeid, node=None):
...
+    if not out:
+        return
+    pcmk_delay_max_v_list = re.findall("pcmk_delay_max=(\d+)s*", out)
+    if not pcmk_delay_max_v_list:
+        utils.set_property("priority-fencing-delay", 0)
+    elif is_2node_wo_qdevice:
+        utils.set_property("priority-fencing-delay", 2*int(max(pcmk_delay_max_v_list)), conditional=True)
+    else:
+        utils.set_property("priority-fencing-delay", 0)
+
```
zzhou1, line 1878: I would suggest

```
    if pcmk_delay_max_v_list and is_2node_wo_qdevice:
        utils.set_property("priority-fencing-delay", 2*int(max(pcmk_delay_max_v_list)), conditional=True)
    else:
        utils.set_property("priority-fencing-delay", 0)
```

crmsh/bootstrap.py
```diff
@@ -1844,6 +1844,37 @@ def update_nodeid(nodeid, node=None):
...
+    """
+    Adjust priority in rsc_defaults
+
+    Set priority=1 when current cluster is 2 nodes without qdevice;
+    else set priority=0
+    """
+    if is_2node_wo_qdevice:
+        utils.set_property("priority", 1, property_type="rsc_defaults", property_id="rsc-options", conditional=True)
+    else:
+        utils.set_property("priority", 0, property_type="rsc_defaults", property_id="rsc-options")
```
gao-yan, line 1857: Does it mean we always want it to be put into "rsc-options" meta set, no matter if there's any other existing set? Wondering what is the consideration here.

gao-yan (approved):
Very sensible improvement. Thanks, @liangxin1300 !

Only a couple of nitpickings.

crmsh/utils.py
```diff
@@ -3028,15 +3022,26 @@ def check_no_quorum_policy_with_dlm():
...
-    _value = get_property(property_name)
-    value_from_cib = int(_value.strip('s')) if _value else 0
-    if value_from_cib < value_from_calculation:
-        cmd = "crm configure property {}={}".format(property_name, value_from_calculation)
-        get_stdout_or_raise_error(cmd)
+    origin_value = get_property(property_name, property_type)
+    if origin_value and str(origin_value) == str(property_value):
+        return
+    if conditional:
+        value_from_cib = int(origin_value.strip('s')) if origin_value else 0
```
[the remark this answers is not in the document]
liangxin1300: changed, compare two values after both parsed by `crm_msec`

crmsh/bootstrap.py
liangxin1300: Thanks for remind me this.

I changed inside `set_property`, check if property_id exists, if not, set the property_id as None, to use the default one

crmsh/bootstrap.py
```diff
@@ -1844,6 +1844,37 @@ def update_nodeid(nodeid, node=None):
...
+    Adjust priority-fencing-delay
+
+    When pcmk_delay_max is set in fence agent,
+    and the current cluster is 2 nodes without qdevice,
+    set priority-fencing-delay=2*pcmk_delay_max
+    """
+    out = utils.get_stdout_or_raise_error("crm configure show related:stonith")
+    if not out:
+        return
+    pcmk_delay_max_v_list = re.findall("pcmk_delay_max=(\d+)s*", out)
```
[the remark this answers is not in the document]
liangxin1300: Changed, use `crm_msec`

crmsh/utils.py
```diff
@@ -3028,15 +3022,30 @@ def check_no_quorum_policy_with_dlm():
...
+    if origin_value and str(origin_value) == str(property_value):
+        return
+    if conditional:
+        if crm_msec(origin_value) >= crm_msec(property_value):
+            return
+    if property_id:
+        out = get_stdout_or_raise_error("crm configure show")
+        if not re.search(" {}:".format(property_id), out):
+            logger.warning("No such id \"{}\" exist".format(property_id))
+            property_id = None
```
gao-yan, line 3043: AFAIU, this way, it'll basically let crm_attribute create the set with its convention. But OTOH, any invocations of `crm configure rsc_defaults` won't use this set anyways, right?  Not sure if we could make everything sensible with this PR,  otherwise we could still stick with your previous way. Then an improvement in the future could be to make both "crm configure" and "crm bootstrap" consistently follow crm_attribute's convention.

liangxin1300: I think here calling `crm configure rsc_defaults` is better, since in cibconfig.py, already did a lots of work, if the `rsc-option` not exist, `rsc-option` will be created and make sure all rsc_defaults' properties are under this id

liangxin1300: Already changed to use `crm configure property|rsc_defaults`

PR #1020, opened 2022-08-25
Dev: log: print begin and end marker in different lines in status_long

nicholasyang2022:
The console format will mass if there are messages during status_long:

```
INFO: BEGIN Starting pacemaker(delaying start of sbd for 71s) ...INFO: This is demo log
...........done
```

After applied this PR when succeeded:

```
INFO: BEGIN Starting pacemaker(delaying start of sbd for 71s) ...
INFO: This is demo log.
INFO: END Starting pacemaker(delaying start of sbd for 71s).
```

When failed:

```
INFO: BEGIN Starting pacemaker(delaying start of sbd for 71s) ...
INFO: This is demo log.
ERROR: FAIL Starting pacemaker(delaying start of sbd for 71s).
ERROR: cluster.start: Failed on ha-cluster-1: Exited with error code 1, Error output: Failed to start pacemaker.service: Unit pacemaker.service is masked.
```

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -46,14 +46,11 @@ class ConsoleCustomFormatter(logging.Formatter):
         "ERROR": constants.RED
     }
     FORMAT = "%(levelname)s: %(message)s"
-    FORMAT_RAW = "%(message)s"
 
-    def __init__(self, lineno=-1, raw_msg=False, fmt=None):
+    def __init__(self, lineno=-1, fmt=None):
         self.lineno = lineno
         if fmt:
             super().__init__(fmt=fmt)
-        elif raw_msg:
-            super().__init__(fmt=self.FORMAT_RAW)
         else:
             super().__init__(fmt=self.FORMAT)
 
@@ -246,18 +243,6 @@ class LoggerUtils(object):
             console_handler.terminator = "\n"
 
     @contextmanager
-    def raw_msg(self):
-        """
-        Only print raw message, without levelname
-        """
-        console_handler = self.get_handler("console")
-        try:
-            console_handler.setFormatter(ConsoleCustomFormatter(raw_msg=True))
-            yield
-        finally:
-            console_handler.setFormatter(ConsoleCustomFormatter())
-
-    @contextmanager
     def only_file(self):
         """
         Only log to file in bootstrap logger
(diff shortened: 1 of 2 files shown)


crmsh/log.py
```diff
@@ -311,27 +311,19 @@ def line_number(self):
         finally:
             self.reset_lineno(self.__save_lineno)
 
-    def status_done(self):
-        """
-        Log info message "done" in console
-        """
-        with self.raw_msg():
```
liangxin1300, line 318: The `raw_msg` function, and the `raw_msg` parameter in `ConsoleCustomFormatter` class, should also be removed

liangxin1300:
And please backport to crmsh-4.4 branch
Thanks!

crmsh/log.py
liangxin1300: And please remove `status_progress` function in bootstrap.py, it also make mess like:
```
INFO: BEGIN Waiting for cluster ...
...........INFO: END Waiting for cluster.
```

nicholasyang2022: > And please remove `status_progress` function in bootstrap.py, it also make mess like:

Instead of removing `status_progress`, we can make a progress bar which occupies a whole line of a terminal, so that logs will not interleave with it.

liangxin1300: OK

crmsh/log.py
```diff
@@ -423,6 +400,22 @@ def update_err(self, obj_id, cibadm_opt, xml, rc):
...
+        self._i = 0
+
+    def progress(self):
+        try:
+            width, _ = os.get_terminal_size()
+        except OSError:
+            # not a terminal
+            return
+        self._i = (self._i + 1) % width
+        line = '\r{}>>{}'.format('.' * self._i, '.' * (width - 2 - self._i))
```
liangxin1300, line 414: I think take the whole width of terminal looks too long, maybe this one looks better?
```
INFO: BEGIN Waiting for cluster ...
...........>>
INFO: END Waiting for cluster.
```
This is the code I change:
```
diff --git a/crmsh/bootstrap.py b/crmsh/bootstrap.py
index afe887b0..19488e4a 100644
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -373,6 +373,7 @@ def wait_for_cluster():
     with logger_utils.status_long("Waiting for cluster") as progress_bar:
         while True:
             if is_online():
+                print()
                 break
             status_progress(progress_bar)
             sleep(2)
diff --git a/crmsh/log.py b/crmsh/log.py
index bee8f6f6..6903616b 100644
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -411,7 +411,7 @@ class ProgressBar:
             # not a terminal
             return
         self._i = (self._i + 1) % width
-        line = '\r{}>>{}'.format('.' * self._i, '.' * (width - 2 - self._i))
+        line = '\r{}>>'.format('.' * self._i)
         sys.stdout.write(line)
         sys.stdout.flush()
```
And for sure the caller should remember to print new line:)

zzhou1: Make sens to me, I prefer to delete ">>" even.

```
-        line = '\r{}>>{}'.format('.' * self._i, '.' * (width - 2 - self._i))
+        line = '\r{}'.format('.' * self._i)
```

nicholasyang2022: > I think take the whole width of terminal looks too long, maybe this one looks better?

It works by consuming the whole width of terminal so that another log written to terminal start from a new line. If it looks too long, `-` can be replaced by space.

zzhou1: > replaced by space.
Nice tricky point, to shift the coming text to the new line ;)

zzhou1 (approved):

PR #1024, opened 2022-09-01
Fix: crash_test: do not use firewalld to isolate a cluster node

nicholasyang2022:
Removing ports from firewalld zone will not break established connections.

### DIFF
--- a/crmsh/crash_test/main.py
+++ b/crmsh/crash_test/main.py
@@ -155,7 +155,7 @@ For each --kill-* testcase, report directory: {}'''.format(context.logfile,
     group_mutual.add_argument('--fence-node', dest='fence_node', metavar='NODE',
                               help='Fence specific node')
     group_mutual.add_argument('--split-brain-iptables', dest='sp_iptables', action='store_true',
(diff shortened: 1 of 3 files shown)


zzhou1 (changes requested):
Great, I like the idea to delete do_block_firewalld() and it does simplify code. The goal of the use case to simulate an isolated cluster node and it is good enough to let do_block_iptables() do the job on-the-fly.

Question, did you verify 5-node for diskless sbd use case, btw?

Minor comments, 
- It is useful tracking reference to add bsc#1192467 in the commit log. 
- I prefer the title to be more use case oriented, something like
  "Fix: crash_test: do not use firewalld to isolate the cluster node"

nicholasyang2022:
> Question, did you verify 5-node for diskless sbd use case, btw?

I verified it today. The isolated node self fences after SBD_TIMEOUT and remaining nodes form new membership after stonish timeout.

zzhou1 (approved):

liangxin1300 (approved):
LGTM
Thanks!

PR #1027, opened 2022-09-06
fix initializing a cluster step by step

nicholasyang2022:
1. refactor `SBDTimeout` so that it can be calculated without calling `init_sbd` first.
2. add `remote_auth` to the list of init stages so that authkey for pacemaker can be created.


liangxin1300:
Looks good to me
Thanks!

liangxin1300 (approved):

PR #1028, opened 2022-09-15
Dev: bootstrap: Adjust sbd related timeout when add/remove qdevice

liangxin1300:
* After adding qdevice via stage on a running cluster with sbd running, the parameter `pcmk_delay_max` is useless and should be removed
* After removing qdevice from a two-node cluster with sbd running, the parameter `pcmk_delay_max` should be added

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1375,6 +1375,10 @@ def init_qdevice():
 
     qdevice_inst.config_and_start_qdevice()
 
+    if _context.stage == "qdevice" and utils.service_is_active("sbd.service"):
+        from .sbd import SBDTimeout
+        SBDTimeout.adjust_sbd_timeout_related_cluster_configuration()
+
 
 def init():
     """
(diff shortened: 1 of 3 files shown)


zzhou1 (commented):
The biggest gap here is those use cases using non-sbd fence agents. We should decouple the functionality outside of sbd alone.

liangxin1300:
> The biggest gap here is those use cases using non-sbd fence agents. We should decouple the functionality outside of sbd alone.

There are other places to call `SBDTimeout.adjust_sbd_timeout_related_cluster_configuration`, I think we should create another PR to do it

zzhou1 (approved):
> > The biggest gap here is those use cases using non-sbd fence agents. We should decouple the functionality outside of sbd alone.
> 
> There are other places(like join and remove) to call `SBDTimeout.adjust_sbd_timeout_related_cluster_configuration`, I think we should create another PR to do it

[TEAM-6763](https://jira.suse.com/browse/TEAM-6763) is created to track those non-sbd use cases.

PR #1029, opened 2022-09-16
Dev: cibconfig: "crm config show related:xxx" provides partial search among class, provider, type fields

liangxin1300:
```
crm(live/15sp4-1)configure# show related:ocf
primitive d Dummy
primitive vip IPaddr2 \
	params ip=10.10.10.125

crm(live/15sp4-1)configure# show related:stonith
primitive stonith-sbd stonith:external/sbd \
	params
primitive ttt stonith:rps10 \
	params pcmk_delay_max=30 serial_to_targets=sf
primitive xxx stonith:fence_eps \
	params pcmk_delay_base=10

crm(live/15sp4-1)configure# show related:ip
primitive vip IPaddr2 \
        params ip=10.10.10.125
```

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -3135,30 +3135,37 @@ class CibFactory(object):
         return [x for x in self.cib_objects
                 if x.updated or x.origin == "user"]
 
-    def get_elems_on_ra_type(self, spec):
+    def get_elems_of_ra_partial_search(self, spec):
         """
-        Get elements by given ra class:provider:type or class:type or only type
+        Get elements by given ra class:provider:type or class:type or only class/provider/type
         return [] if no results
         """
+        def match_type(obj, tp):
+            type_res = obj.node.get("type")
+            return type_res and tp.lower() in type_res.lower()
+
         content_list = spec.split(':')[1:]
         if len(content_list) > 3:
             return []
         if len(content_list) == 3:
             cls, provider, tp = content_list
             return [x for x in self.cib_objects 
-                    if x.node.get("type") == tp
+                    if match_type(x, tp)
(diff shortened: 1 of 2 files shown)


crmsh/cibconfig.py
```diff
@@ -3158,7 +3158,8 @@ def get_elems_on_ra_type(self, spec):
             tp = content_list[0]
             if not tp:
                 return []
-            return [x for x in self.cib_objects if x.node.get("type") == tp]
+            return [x for x in self.cib_objects if x.node.get("type") == tp or
+                    x.node.get("class") == tp]
```
zzhou1, line 3162: Absolutely better UX! 

Further improvement, I prefer to do partial match to give more flexibility for human-being, something like,

```
tp in x.node.get("class") or tp in x.node.get("provider") or tp in x.node.get("type") 
```

The same for the other two situations: class:provider:type and class:type

zzhou1: Also, I would suggest a slightly different title for the git commit log, eg.

```
Dev: cibconfig: "crm config show related:xxx" provides partial search among class, provider, type fields
```

liangxin1300: @zzhou1 
Thanks! Already changed!

crmsh/cibconfig.py
```diff
@@ -3137,7 +3137,7 @@ def modified_elems(self):
 
     def get_elems_on_ra_type(self, spec):
```
zzhou1, line 3138: Maybe rename the function name as well, something like,
```suggestion
    def get_elems_of_ra_partial_search(self, spec):
```

zzhou1 (approved):

PR #1030, opened 2022-09-17
Dev: bootstrap: Adjust pcmk_delay_max and stonith-timeout for all configured fence agents

liangxin1300:


crmsh/cibconfig.py
```diff
@@ -2890,6 +2890,32 @@ def prim_id_list(self):
         "List of primitives ids (for group completion)."
         return [x.obj_id for x in self.cib_objects if x.obj_type == "primitive"]
 
+    def fence_id_list(self):
+        """
+        List all configured fence agent's id
+        """
+        return [x.obj_id for x in self.cib_objects if x.node.get('class') == "stonith"]
+
+    def fence_id_list_with_pcmk_delay(self):
```
zzhou1, line 2899: 1. I wonder to be more strict criteria of this function is to check `pcmk_delay_max != 0`. In case some left over CIB with explicit `pcmk_delay_max=0`, in theory.

2. I have no obvious use case to include  `pcmk_delay_base` here. Along with this, I would prefer the precise function naming: fence_id_list_with_pcmk_delay_max()

crmsh/bootstrap.py
```diff
@@ -2417,4 +2406,52 @@ def get_stonith_timeout_generally_expected():
...
+def adjust_pcmk_delay():
+    """
+    For each fence agent,
+    add parameter pcmk_delay_max when cluster is two-node cluster without qdevice
+    else remove pcmk_delay_max
+    """
+    cib_factory.refresh()
+
+    if utils.is_2node_cluster_without_qdevice():
+        for res in cib_factory.fence_id_list_without_pcmk_delay():
```
zzhou1, line 2420: To make adjust_pcmk_delay() more general even beyond bootstrap, I suggest to adjust only if pcmk_delay_max < PCMK_DELAY_MAX .

crmsh/bootstrap.py
```diff
@@ -2417,4 +2406,52 @@ def get_stonith_timeout_generally_expected():
...
+            cmd = "crm resource param {} set pcmk_delay_max {}s".format(res, PCMK_DELAY_MAX)
+            utils.get_stdout_or_raise_error(cmd)
+            logger.debug("Add parameter 'pcmk_delay_max={}s' for resource '{}'".format(PCMK_DELAY_MAX, res))
+    else:
```
zzhou1, line 2424: Minor argument, the "else" branch contributes nothing to the cluster behavior, deleting it could save 4 lines of code ;)

zzhou1 (commented):
Very nice improvement!

crmsh/bootstrap.py
liangxin1300: No harm to keep it, else I need to adjust the test cases:)

crmsh/cibconfig.py
liangxin1300: > 1. I wonder to be more strict criteria of this function is to check `pcmk_delay_max != 0`. In case some left over CIB with explicit `pcmk_delay_max=0`, in theory.

changed when the `value > 0`

> 2. I have no obvious use case to include  `pcmk_delay_base` here. Along with this, I would prefer the precise function naming: fence_id_list_with_pcmk_delay_max()

I prefer to keep this since I see it's the same purpose with pcmk_delay_max. OK, no harm to keep it

crmsh/bootstrap.py
liangxin1300: Seems still no harm to keep it unchanged?:)

liangxin1300: `fence_id_list_without_pcmk_delay` will show the fence agent id list which don't contain pcmk_delay_max/base, or the value <= 0

crmsh/cibconfig.py
zzhou1: I insist to remove the part in related to pcmk_delay_base here. Reasons are:
a) no use case deals with it so far. 
b) using it here just creates unnecessary brain power to understand the code, and to think about why, in the end it is not related to pcmk_delay_max use case at all.

crmsh/bootstrap.py
zzhou1: Agree, just stay within the bootstap scope, it is enough. 

I keep thinking this is a low hanging fruit to go extra miles for `crm cluster health` concept in the near future.

crmsh/cibconfig.py
liangxin1300: Already changed. removed the base one

zzhou1 (approved):

PR #1009, opened 2022-08-16
Feature: replace root by a custom user with root privileges

aleksei-burlakov:
It enables creating the cluster under a custom user
Assume the users are alice@node1 and non@node2.

1) Bring the alice into sudoers file on the node1
and bob on the node2
for example
`$ sudo echo "alice ALL=(ALL:ALL) NOPASSWD: ALL" >> /etc/sudoers`
(use `visudo` rather than echo)

2) add /usr/sbin to the PATH for alice and bob (writ it in the ~/.bashrc)
`$ export PATH=$PATH:/usr/sbin`

And now you can do the usual routines under alice

```
alice@node1$ crm cluster init -y
bob@node2$ crm cluster join -c alice@node1 -y
or
alice@node1$ crm cluster init -N bob@node2 -y
```

The crm will start as a usual process with the same context as the
current user (alice/bob). However when it would lack permissions
for example to read/write file it would try again as the super-user.

PR #1031, opened 2022-09-20
Fix: ha-cluster-join fails to add nodes to a existing cluster after the version update (bsc#1201785)

nicholasyang2022:
# Problem

1. When directory `~hacluster/.ssh` does not exist, crm cluster join will failed
2. When file `~hacluster/.ssh/id_*.pub` does not exist, crm cluster join will succeed with a warning, but passwordless ssh is not set up.

# Fix

1. If the directory does not exists, failed with a suggestion to run crm cluster init ssh.
2. if public key does not exists, fail the join procedure.

A demo util for remote ssh calling is used here. Maybe it needs to be replaced with non-privileged remote utils from #1009 in the future.

# Work in Progress:

- [ ] checks should be added to `crm report -u user`
- [ ] unit tests

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -26,6 +26,7 @@ from pathlib import Path
 from contextlib import contextmanager
 from . import config
 from . import utils
+from . import remoteutil
 from . import xmlutil
 from .cibconfig import mkset_obj, cib_factory
 from . import corosync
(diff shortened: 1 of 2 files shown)


aleksei-burlakov:
Thank you for your contribution. There are exist already many functions for remote execution mostly in crmsh/utils.py. Some of them utilize the ssh, the others the parallax. I don't think we should add more of those, or maybe I missed the point.

liangxin1300:
Hi @nicholasyang2022 ,

About the fix solution:
```
If the directory does not exists, failed with a suggestion to run crm cluster init ssh.
if public key does not exists, fail the join procedure.
```
Can't we automate these steps instead of giving suggestion or fail the procedure?

nicholasyang2022:
closed in favor of #1033.

PR #1032, opened 2022-09-22
Fix: ui_cluster: 'crm cluster stop' failed to stop services

liangxin1300:
## Problem
1. Start corosync.service alone, like, manually run 'systemctl start corosync', then run 'crm cluster stop' will failed
2. Run 'crm cluster stop --all' quickly after 'crm cluster start --all', then got erro 'ERROR: No DC found currently, please wait if the cluster is still starting'

## Solution:
For the case 1, simply stop corosync.service;
For the case 2, should retry to get the DC in a loop, with a default timeout(dc-deadtime+5s), then
    raise the error while reach the timeout


zzhou1 (approved):
Very nice improvement, great!

PR #1033, opened 2022-09-26
Fix: ha-cluster-join fails to add nodes to a existing cluster after the version update (bsc#1201785)

nicholasyang2022:
# Mechanism

* add a mechanism for updating cluster config after version update:
  - A sequence is written to `~hacluster/crmsh/upgrade_seq` and crmsh checks it to determine whether upgrade is needed whenever it runs.
  - Upgrade is performed only when all nodes in cluster are running compatible version of crmsh. It is determined by comparing `CURRENT_UPGRADE_SEQ` in upgradeutil.py.
  - Upgrade needs user confirm and is only performed in interactive mode.
* add health check and fix for passwordless ssh authentication for hacluster
* run health fix for passwordless ssh authentication for hacluster when with that upgrade mechanism
* run health fix for passwordless ssh authentication for hacluster when a node fails to join the cluster because of it

# Use Cases

Upgrade is performed in the following scenarios:

1. Running any crmsh command in interactive mode when crmsh on all nodes of a cluster have been upgraded to a same version.
2. Joining a cluster: crmsh is running on the joining node, which is not a member of the cluster, so it cannot perform upgrade on existing nodes.
2. When a new node is joining the cluster, and fails because of the lack of hacluster's keys,.

A warning is printed in the following scenarios:

1. Nodes of a cluster are running different version of crmsh.
2. Upgraded is needed and crmsh is not running in interactive mode.

### DIFF
--- a/crmsh/upgradeutil.py
+++ b/crmsh/upgradeutil.py
@@ -1,0 +1,180 @@
+import logging
+import os.path
+import typing
+
+import parallax
+import sys
+
+import crmsh.parallax
+import crmsh.utils
+
+
+# pump this seq when upgrade check need to be run
+CURRENT_UPGRADE_SEQ = 1
+DATA_DIR = os.path.expanduser('~hacluster/crmsh')
+SEQ_FILE_PATH = DATA_DIR + '/upgrade_seq'
+# touch this file to force a upgrade process
+FORCE_UPGRADE_FILE_PATH = DATA_DIR + '/upgrade_forced'
+
+
+logger = logging.getLogger(__name__)
+
+
+class _SkipUpgrade(Exception):
+    pass
+
+
+def _get_file_content(path, default=None):
+    try:
+        with open(path, 'rb') as f:
+            return f.read()
+    except FileNotFoundError:
+        return default
+
+
+def _parallax_run(nodes: str, cmd: str) -> typing.Dict[str, typing.Tuple[int, bytes, bytes]]:
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    ret = dict()
+    for node, result in parallax.run(nodes, cmd, parallax_options).items():
+        if isinstance(result, parallax.Error):
+            logger.warning("SSH connection to remote node %s failed.", node, exc_info=result)
+            raise result
+        ret[node] = result
+    return ret
+
+
+def _is_upgrade_needed(nodes):
+    """decide whether upgrading is needed by checking local sequence file"""
+    needed = False
+    try:
+        os.stat(FORCE_UPGRADE_FILE_PATH)
+        needed = True
+    except FileNotFoundError:
+        pass
+    if not needed:
+        try:
+            local_seq = int(_get_file_content(SEQ_FILE_PATH, b'').strip())
+        except ValueError:
+            local_seq = 0
+        needed = CURRENT_UPGRADE_SEQ > local_seq
+    return needed
+
+
+def _is_cluster_target_seq_consistent(nodes):
+    cmd = '/usr/bin/env python3 -m crmsh.upgradeutil get-seq'
+    try:
+        results = list(_parallax_run(nodes, cmd).values())
+    except parallax.Error as e:
+        raise _SkipUpgrade() from None
+    try:
+        return all(CURRENT_UPGRADE_SEQ == int(stdout.strip()) if rc == 0 else False for rc, stdout, stderr in results)
+    except ValueError as e:
+        logger.warning("Remote command '%s' returns unexpected output: %s", cmd, results, exc_info=e)
+        return False
+
+
+def _get_minimal_seq_in_cluster(nodes):
+    try:
+        return min(
+            int(stdout.strip()) if rc == 0 else 0
+            for rc, stdout, stderr in _parallax_run(nodes, 'cat {}'.format(SEQ_FILE_PATH)).values()
+        )
+    except ValueError:
+        return 0
+
+
+def _upgrade(nodes, seq):
+    logger.info("Upgrading cluster...")
+    if seq <= 0:
+        seq_0_setup_hacluster_passwordless(nodes)
+
+
+def upgrade_if_needed():
+    nodes = crmsh.utils.list_cluster_nodes()
+    if nodes and _is_upgrade_needed(nodes):
+        logger.info("crmsh version is newer than its configuration. Configuration upgrade is needed.")
+        try:
+            if not _is_cluster_target_seq_consistent(nodes):
+                logger.warning("crmsh version is inconsistent in cluster.")
+                raise _SkipUpgrade()
+            seq = _get_minimal_seq_in_cluster(nodes)
+            logger.debug("Upgrading crmsh configuration from seq %s to %s.", seq, CURRENT_UPGRADE_SEQ)
+            _upgrade(nodes, seq)
+        except _SkipUpgrade:
+            logger.warning("Configuration upgrade skipped.")
+            return
+        crmsh.parallax.parallax_call(
+            nodes,
+            "mkdir -p '{}' && echo '{}' > '{}'".format(DATA_DIR, CURRENT_UPGRADE_SEQ, SEQ_FILE_PATH),
+        )
+        crmsh.parallax.parallax_call(nodes, 'rm -f {}'.format(FORCE_UPGRADE_FILE_PATH))
+        logger.debug("Configuration upgrade finished.", seq, CURRENT_UPGRADE_SEQ)
+
+
+def force_set_local_upgrade_seq():
+    """Create the upgrade sequence file and set it to CURRENT_UPGRADE_SEQ.
+
+    It should only be used when initializing new cluster nodes."""
+    try:
+        os.mkdir(DATA_DIR)
+    except FileExistsError:
+        pass
+    with open(SEQ_FILE_PATH, 'w', encoding='ascii') as f:
+        print(CURRENT_UPGRADE_SEQ, file=f)
+
+
+def seq_0_setup_hacluster_passwordless(nodes):
+    """setup passwordless ssh authentication by running crm cluster ssh init/join on appreciated nodes."""
+    # https://bugzilla.suse.com/show_bug.cgi?id=1201785
+    logger.debug("upgradeutil: setup passwordless ssh authentication for user hacluster")
+    try:
+        nodes_without_keys = [
+            node for node, result in
+            _parallax_run(
+                nodes,
+                '[ -f ~hacluster/.ssh/id_rsa ] || [ -f ~hacluster/.ssh/id_ecdsa ] || [ -f ~hacluster/.ssh/id_ed25519 ]'
+            ).items()
+            if result[0] != 0
+        ]
+    except parallax.Error:
+        raise _SkipUpgrade() from None
+    if nodes_without_keys:
+        if not crmsh.utils.ask("Configuration upgrade: setup passwordless ssh authentication for user hacluster?"):
+            raise _SkipUpgrade()
+        if len(nodes_without_keys) == len(nodes):
+            # pick one node to run init ssh on it
+            init_node = nodes_without_keys[0]
+            # and run join ssh on other nodes
+            join_nodes = list()
+            join_nodes.extend(nodes)
+            join_nodes.remove(init_node)
+            join_target_node = init_node
+        else:
+            nodes_with_keys = set(nodes) - set(nodes_without_keys)
+            # no need to init ssh
+            init_node = None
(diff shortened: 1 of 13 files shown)


crmsh/upgradeutil.py
```diff
@@ -0,0 +1,93 @@
...
+
+def _upgrade(nodes, seq):
+    logger.info("Upgrading cluster...")
+    if seq == 0:
+        seq_0_setup_hacluster_passwordless(nodes)
+
+
+def upgrade_if_needed():
+    nodes = crmutils.list_cluster_nodes()
+    if nodes is not None:
```
liangxin1300, line 65: use `if nodes` might be more safe, since the return value might be None or []

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,93 @@
...
+        print(CURRENT_UPGRADE_SEQ, file=f)
+
+
+def seq_0_setup_hacluster_passwordless(nodes):
+    results = _parallax_call(nodes, '[ -f ~hacluster/.ssh/id_rsa ] || [ -f ~hacluster/.ssh/id_ecdsa ] || [ -f ~hacluster/.ssh/id_ed25519 ]')
+    for node, result in results.items():
+        if isinstance(result, Exception):
+            logger.error('Failed to connect host %s.', node, exc_info=result)
+            raise result
+    # TODO: ask for user confirm
```
liangxin1300, line 80: Yes, if current is in the interactive mode, should ask for user confirm

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,93 @@
...
+    logger.info("Upgrading passwordless ssh authentication for user hacluster...")
+    nodes_without_keys = [node for node, result in results.items() if result[0] != 0]
+    if nodes_without_keys:
+        init_node = nodes_without_keys[0]
+        join_nodes = list()
+        join_nodes.extend(nodes)
+        join_nodes.remove(init_node)
+        remoteutil.check_call(init_node, 'crm cluster init ssh -y')
+        results = _parallax_call(nodes, 'crm cluster join ssh -c {} -y'.format(init_node))
```
liangxin1300, line 89: I suggest use a loop here, to avoid possible race condition

nicholasyang2022:
wait krig/parallax#15

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,103 @@
...
+def _upgrade(nodes, seq):
+    logger.info("Upgrading cluster...")
+    if seq == 0:
+        seq_0_setup_hacluster_passwordless(nodes)
+
+
+def upgrade_if_needed():
+    nodes = crmutils.list_cluster_nodes()
+    if nodes:
+        _upgrade_if_needed(nodes, lambda seq: _upgrade(nodes, seq))
```
zzhou1, line 75: Given a 5 nodes cluster, what's the upgrade sequence would be, node by node from scratch? It reasonable to deal with those upgraded nodes, but probably too aggressive to deal with those old nodes?

nicholasyang2022: The sequence is run `crm cluster init ssh -y` on one of the nodes and `crm cluster join ssh  -c {} -y` on others. Do you mean we need to deal with a partially upgraded cluster, with nodes running different versions of ha stack?

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,103 @@
...
+    #return ret
+
+
+def _upgrade_if_needed(nodes, fn: typing.Callable[[int], None]):
+    seq = 0
+    # FIXME: too slow, check localhost only
+    results = list(_parallax_call(nodes, 'test -f {}'.format(FORCE_UPGRADE_FILE_PATH)).values())
+    if any(isinstance(item, Exception) for item in results):
+        logger.info("parallax failed: nodes: %s, results: %s", nodes, results)
+        return
```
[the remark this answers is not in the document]
nicholasyang2022: It is a good point. A cluster should eventually be upgraded to a consistent version. So I would rather not to upgrade when some nodes are unreachable, unless the cluster is quite large and the possibility of temporary network failure were significantly high so that it were impossible to reach all nodes at the same time.

It also remains me that I should test more failing cases.

crmsh/upgradeutil.py
zzhou1: > The sequence is run `crm cluster init ssh -y` on one of the nodes and `crm cluster join ssh -c {} -y` on others. Do you mean we need to deal with a partially upgraded cluster, with nodes running different versions of ha stack?

What you described here is to construct a brand new cluster, and even you may assume all nodes are upgraded already.  However, the use case scenario of this PR is more complex and it is to upgrade an existing cluster, and some nodes are upgraded and some might not yet.

nicholasyang2022: OK. I will delay the ssh init/join sequence until all nodes have been upgraded to new version.

zzhou1: In regarding to the version upgrade, I'm not sure if crmsh could leverage the bigger picture from pacemaker, for example,
https://clusterlabs.org/pacemaker/doc/2.1/Pacemaker_Administration/html/upgrading.html#rolling-node-by-node

liangxin1300:
Hi @nicholasyang2022, as parallax's request get accepted, I think you can remove remoteutil.py code and using parallax now?

liangxin1300:
Loop #742 here

crmsh/sbd.py
```diff
@@ -1,6 +1,8 @@
 import os
 import re
 import shutil
+import subprocess
+
```
liangxin1300, line 5: line 4-5 seems useless here?

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+def _parallax_run(nodes, cmd):
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    return parallax.run(nodes, cmd, parallax_options)
+
+
+def _upgrade_if_needed(nodes, fn: typing.Callable[[int], None]):
+    seq = 0
+    forced = 0 == subprocess.call(['test', '-f', FORCE_UPGRADE_FILE_PATH])
+    needed = forced
```
liangxin1300, line 42: Since `forced` not used again in this function, how about
```
needed = 0 == subprocess.call(['test', '-f', FORCE_UPGRADE_FILE_PATH])
```
directly?

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+        init_node = nodes_without_keys[0]
+        join_nodes = list()
+        join_nodes.extend(nodes)
+        join_nodes.remove(init_node)
+        crmparallax.parallax_call([init_node], 'crm cluster init ssh -y')
+        for node in join_nodes:
+            result = next(iter(_parallax_run([node], 'crm cluster join ssh -c {} -y'.format(init_node)).values()))
+            if isinstance(result, Exception):
+                logger.error('Failed to join ssh on host %s.', node, exc_info=result)
+                raise result
```
liangxin1300, line 112: I think It'd be better to add one or two log to describe what is this process doing:)

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+def force_set_local_upgrade_seq():
+    try:
+        os.mkdir(DATA_DIR)
+    except FileExistsError:
+        pass
+    with open(SEQ_FILE_PATH, 'w', encoding='ascii') as f:
+        print(CURRENT_UPGRADE_SEQ, file=f)
+
+
+def seq_0_setup_hacluster_passwordless(nodes):
```
liangxin1300, line 93: Better to add docstrings here

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def _parallax_run(nodes, cmd):
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    return parallax.run(nodes, cmd, parallax_options)
+
+
+def _upgrade_if_needed(nodes, fn: typing.Callable[[int], None]):
+    seq = 0
```
liangxin1300, line 40: Better to add docstrings here, and add more comments for the major logic for this function

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+        return default
+
+
+def _parallax_run(nodes, cmd):
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    return parallax.run(nodes, cmd, parallax_options)
+
+
+def _upgrade_if_needed(nodes, fn: typing.Callable[[int], None]):
```
liangxin1300, line 39: Better to add docstrings here, and add more comments for the major logic for this function

nicholasyang2022: OK. And I will refactor this function to avoid passing lambda functions here.

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,112 @@
...
+        with open(path, 'rb') as f:
+            return f.read()
+    except FileNotFoundError:
+        return default
+
+
+def _parallax_run(nodes, cmd):
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    return parallax.run(nodes, cmd, parallax_options)
```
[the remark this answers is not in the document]
nicholasyang2022: `_parallax_run()` use a different API convention from those in `crmsh.parallax`. It do not raise ValueError in a failure scenario. So I think it is not suitable to move it to `crmsh.parallax`.

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,178 @@
...
+        return default
+
+
+def _parallax_run(nodes: str, cmd: str) -> typing.Dict[str, typing.Tuple[int, bytes, bytes]]:
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    ret = dict()
+    for node, result in parallax.run(nodes, cmd, parallax_options).items():
+        if isinstance(result, parallax.Error):
+            logger.warning("SSH connection to remote node %s failed.", exc_info=result)
```
liangxin1300, line 40: There might miss node to match the`%s`  in the warning log

crmsh/upgradeutil.py
```diff
@@ -0,0 +1,179 @@
...
+
+def _parallax_run(nodes: str, cmd: str) -> typing.Dict[str, typing.Tuple[int, bytes, bytes]]:
+    parallax_options = parallax.Options()
+    parallax_options.ssh_options = ['StrictHostKeyChecking=no', 'ConnectTimeout=10']
+    ret = dict()
+    for node, result in parallax.run(nodes, cmd, parallax_options).items():
+        if isinstance(result, parallax.Error):
+            logger.warning("SSH connection to remote node %s failed.", node, exc_info=result)
+            raise result
+        dict[node] = result
```
liangxin1300, line 43: This is wrong

liangxin1300 (approved):
Great work, thanks!

zzhou1 (approved):

gao-yan (commented):
Good idea on the rewording. Probably we could be even more clear here, something like "Upgrade of crmsh configuration" rather than just "Configuration upgrade"?

nicholasyang2022:
Updates:

hacluster's ssh key pair will be generated automatically now when a new node joins a cluster.

liangxin1300:
Good, now the join node can join the running cluster successfully, without failures
```
15sp4-3:~ # crm cluster join -c 15sp4-1 -y
WARNING: chronyd.service is not configured to start at system boot.
INFO: Configuring SSH passwordless with root@15sp4-1
Password: 
INFO: Configuring SSH passwordless with hacluster@15sp4-1
INFO: Configuring SSH passwordless with root@15sp4-2
Password: 
INFO: Configuring SSH passwordless with hacluster@15sp4-2
```
My steps:

1. node1 and node2 setup a cluster
2. remove hacluster's .ssh directory on both node1 and node2
3. node1, node2 and node3 update the new codes
4. node3 joining, then I found .ssh for hacluster already back on node1 and node2

liangxin1300:
And if just update codes for existing nodes, steps:

1. node1 and node2 setup cluster
2. remove hacluster's .ssh directory on both node1 and node2
3. node1 and node2 update the codes
4.  run any crm command on any nodes(node1 or node2), the process like:

15sp4-1:~ # crm
Upgrade of crmsh configuration: Setup passwordless ssh authentication for user hacluster (y/n)? y
INFO: Upgrade of crmsh configuration succeeded.
5. then .ssh on node1 and node2 already back

PR #1034, opened 2022-09-28
Dev: report: make sure 'crm -d report' could increase the verbosity

liangxin1300:
## Problem
`crm -d report` and `crm -d history peinputs` cannot print report debug messages

## Changes:
- Dev: report: make sure 'crm -d report' could increase the verbosity
  (equal to `crm report -v`)
- Dev: history: print necessary log messages when collecting report


nicholasyang2022 (approved):
Great

PR #1035, opened 2022-09-29
Dev: bootstrap: Add option -x to skip csync2 initialization stage during the whole cluster bootstrap

liangxin1300:
## Motivation

On some environments, the csync2 initialization process consumes too much time comparing to the whole cluster bootstrap process, and could be blamed as too slow. csync2 is not the critical component for the cluster stack. Adding an option to skip that could make it flexible to give users a choice, and users can initialize csyn2 for the cluster later on if need

Another bad situations did happen in the past, csync2 was broken because of its dependency, eg. gnutls bsc#1172695. This functionality gives the flexibility to provide the workaround to setup the cluster without csync2.

## Solution
- On init side, add `-x` option to skip csync2 configure and sync; Or, can set environment variable 'SKIP_CSYNC2_SYNC=true' on init node:
```
# csync2 will not be configured and csync2.socket will not started
crm cluster init -x -y
```
- Based on above, use scp to retrieve all related config files when node joining; use parallax.copy to sync files when files changed
- Based on above, give a hint for user, run `crm cluster init csync2 -y' on any one node, before using csync2 for the first time, then csync2 will be configured:
```
# a hint when joining and detect not using csync2
csync2 is not initiated yet. Before using csync2 for the first time, please run "crm cluster init csync2 -y" on any one node. Note, this may take a while.
```
```
# on any node of a running cluster
crm cluster init csync2 -y
INFO: Configuring csync2
INFO: Starting csync2.socket service on 15sp4-1
INFO: Starting csync2.socket service on 15sp4-2
INFO: Starting csync2.socket service on 15sp4-3
INFO: BEGIN csync2 syncing files
INFO: END csync2 syncing files
INFO: Done (log saved to /var/log/crmsh/crmsh.log)
```
- Consider the scenarios for adding sbd via stage, adding qdevice via stage, and remove node, which all need to sync the configure files
- Some log info changed for csync2
-  Keep original behavior unchanged when '-x' or 'SKIP_CSYNC2_SYNC' not set
## Test
```
  @clean
  Scenario: Skip using csync2 by -x option
    Given   Cluster service is "stopped" on "hanode1"
    Given   Cluster service is "stopped" on "hanode2"
    When    Run "crm cluster init -y -x" on "hanode1"
    Then    Cluster service is "started" on "hanode1"
    And     Service "csync2.socket" is "stopped" on "hanode1"
    And     Service "csync2.socket" is "stopped" on "hanode2"
    When    Run "crm cluster init csync2 -y" on "hanode1"
    Then    Service "csync2.socket" is "started" on "hanode1"
    And     Service "csync2.socket" is "started" on "hanode2"
```

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -34,7 +34,8 @@ from . import corosync
 from . import tmpfiles
 from . import lock
 from . import userdir
-from .constants import SSH_OPTION, QDEVICE_HELP_INFO, STONITH_TIMEOUT_DEFAULT, REJOIN_COUNT, REJOIN_INTERVAL, PCMK_DELAY_MAX, WAIT_TIMEOUT_MS_DEFAULT
+from .constants import SSH_OPTION, QDEVICE_HELP_INFO, STONITH_TIMEOUT_DEFAULT,\
+        REJOIN_COUNT, REJOIN_INTERVAL, PCMK_DELAY_MAX, CSYNC2_SERVICE, WAIT_TIMEOUT_MS_DEFAULT
 from . import ocfs2
 from . import qdevice
 from . import parallax
@@ -57,7 +58,7 @@ SYSCONFIG_FW = "/etc/sysconfig/SuSEfirewall2"
 SYSCONFIG_FW_CLUSTER = "/etc/sysconfig/SuSEfirewall2.d/services/cluster"
 PCMK_REMOTE_AUTH = "/etc/pacemaker/authkey"
 COROSYNC_CONF_ORIG = tmpfiles.create()[1]
-SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service", "csync2.socket"]
+SERVICES_STOP_LIST = ["corosync-qdevice.service", "corosync.service", "hawk.service", CSYNC2_SERVICE]
 USER_LIST = ["root", "hacluster"]
 WATCHDOG_CFG = "/etc/modules-load.d/watchdog.conf"
 BOOTH_DIR = "/etc/booth"
@@ -91,6 +92,7 @@ class Context(object):
         self.no_overwrite_sshkey = None
         self.nic_list = []
         self.node_list = []
+        self.node_list_in_cluster = []
         self.unicast = None
         self.multicast = None
         self.admin_ip = None
@@ -126,6 +128,7 @@ class Context(object):
         self.cloud_type = None
         self.is_s390 = False
         self.profiles_data = None
+        self.skip_csync2 = None
         self.profiles_dict = {}
         self.default_nic_list = []
         self.default_ip_list = []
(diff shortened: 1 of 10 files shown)


crmsh/bootstrap.py
```diff
@@ -2173,9 +2177,16 @@ def bootstrap_join(context):
         lock_inst = lock.RemoteLock(cluster_node)
         try:
             with lock_inst.lock():
+                _context.node_list_in_cluster = utils.fetch_cluster_node_list_from_node(cluster_node)
```
liangxin1300, line 2180: Key point is at the beginning of cluster join, get the list of node in cluster from init node, assign to `_context.node_list_in_cluster`, which will be used in `sync_file`

crmsh/bootstrap.py
```diff
@@ -1508,44 +1532,46 @@ def join_csync2(seed_host):
...
-        # Sync new config out.  This goes to all hosts; csync2.cfg definitely
-        # needs to go to all hosts (else hosts other than the seed and the
-        # joining host won't have the joining host in their config yet).
-        # Strictly, the rest of the files need only go to the new host which
-        # could theoretically be effected using `csync2 -xv -P $(hostname)`,
-        # but this still leaves all the other files in dirty state (becuase
-        # they haven't gone to all nodes in the cluster, which means a
-        # subseqent join of another node can fail its sync of corosync.conf
-        # when it updates expected_votes.  Grrr...
+    logger.info("Configuring csync2")
```
liangxin1300, line 1536: Major changes in `join_csync2` are the log messages

crmsh/bootstrap.py
```diff
@@ -932,31 +939,48 @@ def append_to_remote_file(fromfile, remote_node, tofile):
...
-    with logger_utils.status_long("csync2 checking files"):
-        invoke("csync2", "-cr", "/")
+    for host in host_list:
+        logger.info("Starting {} service on {}".format(CSYNC2_SERVICE, host))
+        utils.start_service(CSYNC2_SERVICE, enable=True, remote_addr=host)
+
+    _msg = "syncing" if _context.skip_csync2 else "checking"
+    with logger_utils.status_long("csync2 {} files".format(_msg)):
+        if _context.skip_csync2:
+            csync2_update("/")
```
nicholasyang2022, line 981: Why is `csync2_update` called when it is "skipped"?

liangxin1300: When running `crm cluster init -x -y` then `crm cluster join`, above piece of codes will not be executed, the csync2.socket will not started and enabled
Then to configure and enable csync2, need to run `crm cluster init csync2 -y`, above piece of codes will be executed.
Since on that time, csync2 db need to be initialized, so need to run `csync2_update`

crmsh/bootstrap.py
```diff
@@ -2106,11 +2101,20 @@ def bootstrap_init(context):
             utils.fatal("Expected NODE argument to csync2_remote")
         _context.cluster_node = args[1]
 
+    if stage and _context.cluster_is_running and not utils.service_is_active(CSYNC2_SERVICE):
+        _context.skip_csync2 = True
```
liangxin1300, line 2105: When running `crm cluster init csync2 -y`, this flag `skip_csync2` will be set as `True`

nicholasyang2022 (approved):

PR #1040, opened 2022-10-25
fix: fail to open log file even if user is in haclient group (bsc#1204670)

nicholasyang2022:
The file had been created with umask 0022 in usual so that it was not group-writable.

Call chown and chmod explicitly to fix it.

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -416,14 +416,6 @@ class ProgressBar:
         sys.stdout.flush()
 
 
-def setup_directory_for_logfile():
-    """
-    Create log file's parent directory
-    """
-    _dir = os.path.dirname(CRMSH_LOG_FILE)
-    os.makedirs(_dir, exist_ok=True)
-
-
 def setup_logging(only_help=False):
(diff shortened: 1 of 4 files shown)


nicholasyang2022:
See also: https://build.opensuse.org/request/show/1031002 for packaging changes to create log directory.

crmsh/log.py
```diff
@@ -432,10 +424,17 @@ def setup_logging(only_help=False):
     if only_help:
         LOGGING_CFG["handlers"]["file"] = {'class': 'logging.NullHandler'}
     else:
-        setup_directory_for_logfile()
+        # dirname(CRMSH_LOG_FILE) should be created by package manager during installation
```
fmherschel, line 427: Looks good for me.

nicholasyang2022:
Cherry pick a commit from #991, which fixes ownership and mode for new files created in log rotation.

liangxin1300 (approved):
Great work!
Thank you @nicholasyang2022

PR #1041, opened 2022-10-27
[crmsh-4.4] fix: fail to open log file even if user is in haclient group (bsc#1204670)

nicholasyang2022:
The file had been created with umask 0022 in usual so that it was not
group-writable.

Call chown and chmod explicitly to fix it.


liangxin1300 (approved):

liangxin1300:
backport from #1040

PR #1044, opened 2022-11-04
Dev: behave: Add functional test for parallax OSError: [Errno 24] Too many open files

liangxin1300:
https://github.com/krig/parallax/issues/17


nicholasyang2022:
We should setrlimit explicitly to ensure it reproduces stably in different environments.

nicholasyang2022 (approved):

PR #1045, opened 2022-11-08
Dev: report: collect ra trace files from specified directories

liangxin1300:
## Problem
After #939 , if user specify other directory to trace ra log, crm report can't collect the log

## Solution
Since the "trace_dir" attribute will be removed from cib after untrace, crm report need to parse crmsh.log to extract custom trace ra log directory on each node

## Functional test
```
  @clean
  Scenario: crm report collect trace ra log
    When    Run "crm configure primitive d Dummy" on "hanode1"
    And     Run "crm configure primitive d2 Dummy" on "hanode1"
    Then    Resource "d" is started on "hanode1"
    And     Resource "d2" is started on "hanode2"
    When    Run "crm resource trace d monitor" on "hanode1"
    Then    Expected "Trace for d:monitor is written to /var/lib/heartbeat/trace_ra/Dummy" in stdout
    When    Wait "10" seconds
    And     Run "crm resource untrace d" on "hanode1"
    And     Run "crm resource trace d2 monitor /trace_d" on "hanode1"
    Then    Expected "Trace for d2:monitor is written to /trace_d/Dummy" in stdout
    When    Wait "10" seconds
    And     Run "crm resource untrace d2" on "hanode1"
    And     Run "crm report report" on "hanode1"
    Then    Directory "trace_ra" in "report.tar.bz2"
    And     Directory "trace_d" in "report.tar.bz2"
    When    Run "rm -rf report.tar.bz2 report" on "hanode1"
```

### DIFF
--- a/crmsh/report/collect.py
+++ b/crmsh/report/collect.py
@@ -50,12 +50,21 @@ def collect_ocfs2_info():
 
 
 def collect_ratraces():
-    trace_dir = os.path.join(constants.HA_VARLIB, "trace_ra")
-    if not os.path.isdir(trace_dir):
+    """
+    Collect ra trace file from default /var/lib/heartbeat/trace_ra and custom one
+    """
+    # since the "trace_dir" attribute been removed from cib after untrace
+    # need to parse crmsh log file to extract custom trace ra log directory on each node
+    log_contents = ""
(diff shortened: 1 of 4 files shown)


liangxin1300:
@aleksei-burlakov @gao-yan Could you please take a look at this?
Thanks!

aleksei-burlakov (approved):
The traceback of the `collect_ratraces` is obfuscated imho, but the change itself looks good to me  👍

Issue #1050, opened 2022-11-11
join process should return immediately when joining self


liangxin1300:
Instead of waiting
```
15sp4-2:~ # crm cluster join -c 10.10.10.42 -y
WARNING: chronyd.service is not configured to start at system boot.
WARNING: Cluster is inactive on 10.10.10.42. Retry in 10 seconds
WARNING: Cluster is inactive on 10.10.10.42. Retry in 10 seconds
WARNING: Cluster is inactive on 10.10.10.42. Retry in 10 seconds

15sp4-2:~ # crm cluster join -c 15sp4-2 -y
WARNING: chronyd.service is not configured to start at system boot.
WARNING: Cluster is inactive on 15sp4-2. Retry in 10 seconds
WARNING: Cluster is inactive on 15sp4-2. Retry in 10 seconds
WARNING: Cluster is inactive on 15sp4-2. Retry in 10 seconds
```

zzhou1:
Even further user experience improvement to reorder the detailed join steps. Example, SBD_DELAY_START is really meaningless in the beginning of the join process. It just creates unnecessary waiting time. It makes sense to delay until the joining node is up and running. Not sure, if you want to create another issue ;)

PR #1055, opened 2022-11-16
Dev: bootstrap: join process return immediately when joining itself

liangxin1300:
fix issue #1050 
- Dev: bootstrap: join process return immediately when joining itself
- Dev: unittest: Add unit test for previous commit

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -175,6 +175,8 @@ class Context(object):
         """
         Validate -N/--nodes option
         """
+        if not self.node_list:
+            return
         self.node_list = utils.parse_append_action_argument(self.node_list)
         me = utils.this_node()
         if me in self.node_list:
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -20,6 +20,7 @@ import time
 import readline
 import shutil
 import yaml
+import socket
 from string import Template
 from lxml import etree
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


crmsh/bootstrap.py
```diff
@@ -184,6 +187,18 @@ def _validate_nodes_option(self):
...
             utils.ping_node(node)
 
+    def _validate_cluster_node(self):
+        """
+        Validate cluster_node on join side
+        """
+        if self.cluster_node and self.type == 'join':
+            try:
+                # self.cluster_node might be hostname or IP address
+                if socket.gethostbyaddr(self.cluster_node)[0] == utils.this_node():
```
nicholasyang2022, line 197: Should we compare ip addresses here? Reverse hostname lookup may not be reliable enough. For example, when multiple names resolve to a same ip in hosts file, reverse lookup returns only the 1st name.

crmsh/bootstrap.py
```diff
@@ -185,6 +188,19 @@ def _validate_nodes_option(self):
...
 
+    def _validate_cluster_node(self):
+        """
+        Validate cluster_node on join side
+        """
+        if self.cluster_node and self.type == 'join':
+            try:
+                # self.cluster_node might be hostname or IP address
+                ip_addr = socket.gethostbyname(self.cluster_node)
+                if utils.InterfacesInfo.ip_in_local(ip_addr):
```
liangxin1300, line 199: Changed as `gethostbyname`, and to check this IP is in the local

nicholasyang2022 (approved):

PR #1058, opened 2022-11-21
Fix: corosync: show corosync ring status if has fault

liangxin1300:


nicholasyang2022:
It will not work, since an exception is raised when rc != 0.

liangxin1300:
> It will not work, since an exception is raised when rc != 0.

It works, if rc!=0, the `FAULTY` output still in stdout

# corosync-cfgtool -s 2> /dev/null
Printing ring status.
Local node ID 2
RING ID 0
        id      = 10.10.10.42
        status  = Marking ringid 0 interface 10.10.10.42 FAULTY
RING ID 1
        id      = 20.20.20.42
        status  = ring 1 active with no faults

nicholasyang2022 (approved):

PR #1065, opened 2022-11-23
Fix: cibconfig: Complete promotable=true and interleave=true for Promoted/Unpromoted resource

liangxin1300:
When configure Promoted/Unpromoted resource, "promotable=true" is required and "interleave=true" is often
preferred.
So the solution is to complete this meta attribute.
Possible cases:
- input like `clone <clone-id> <res-id>`
  If detected this primitive resource contain "promote" and "demote" actions,
  then "meta promotable=true interleave=true" will be completed
- if the input contain `promotable` or `interleave`
  then crmsh doesn't complete it

```
crm(live/15sp4-1)configure# primitive stateful-1 ocf:pacemaker:Stateful
crm(live/15sp4-1)configure# clone c1 stateful-1
node 1: 15sp4-1
node 2: 15sp4-2
	op monitor timeout=20s interval=10s role=Promoted \
	op monitor timeout=20s interval=11s role=Unpromoted \
	op stop timeout=20s interval=0s \
	op promote timeout=10s interval=0s \
	op demote timeout=10s interval=0s
clone c1 stateful-1 \
        meta promotable=true interleave=true
```

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -25,7 +25,7 @@ from . import cibstatus
 from . import crm_gv
 from . import ui_utils
 from . import userdir
-from .ra import get_ra, get_properties_list, get_pe_meta, get_properties_meta
+from .ra import get_ra, get_properties_list, get_pe_meta, get_properties_meta, RAInfo
 from .utils import ext_cmd, safe_open_w, pipe_string, safe_close_w, crm_msec
 from .utils import ask, lines2cli, olist
 from .utils import page_string, cibadmin_can_patch, str2tmp, ensure_sudo_readable
@@ -42,7 +42,7 @@ from .xmlutil import merge_attributes, is_cib_element, sanity_check_meta
 from .xmlutil import is_simpleconstraint, is_template, rmnode, is_defaults, is_live_cib
 from .xmlutil import get_rsc_operations, delete_rscref, xml_equals, lookup_node, RscState
 from .xmlutil import text2elem, is_related, check_id_ref, xml_tostring
-from .xmlutil import sanitize_cib_for_patching
+from .xmlutil import sanitize_cib_for_patching, is_attr_set, get_set_nodes, set_attr
 from .cliformat import get_score, nvpairs2list, abs_pos_score, cli_acl_roleref, nvpair_format
 from .cliformat import cli_nvpair, cli_acl_rule, rsc_set_constraint, get_kind, head_id_format
 from .cliformat import simple_rsc_constraint, cli_rule, cli_format
@@ -793,7 +793,7 @@ def id_for_node(node, id_hint=None):
     return obj_id
 
 
-def postprocess_cli(node, oldnode=None, id_hint=None):
+def postprocess_cli(node, oldnode=None, id_hint=None, complete_advised=False):
     """
     input: unprocessed but parsed XML
     output: XML, obj_type, obj_id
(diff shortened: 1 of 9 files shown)


crmsh/cibconfig.py
```diff
@@ -1621,6 +1621,21 @@ def _repr_cli_head(self, format_mode):
...
+        primitive_list = self.node.xpath('primitive')
+        if not primitive_list:
+            return
+        if is_promotable(self.node):
+            return
+        p = primitive_list[0]
+        ra_inst = RAInfo(p.get('class'), p.get('type'), p.get('provider'))
+        ra_actions_dict = ra_inst.actions()
+        if "promote" in ra_actions_dict and "demote" in ra_actions_dict:
+            logger.warning("\"meta promotable=true\" is needed if want to configure Promoted/Unpromoted ra")
```
gao-yan, line 1637: Let's rephrase it a little like:
"Configure \"meta promotable=true\" if resource instances need to be Promoted/Unpromoted with this resource agent".

gao-yan (commented):
Good idea and LGTM. Only a nitpicking about the phrasing of the warning.

crmsh/cibconfig.py
```diff
@@ -816,26 +816,49 @@ def postprocess_cli(node, oldnode=None, id_hint=None):
...
+def complete_advised_meta(node):
+    """
+    Complete advised meta attributes
+    """
+    if node.tag != "clone":
+        return
+    primitive_list = node.xpath('primitive')
+    if not primitive_list:
+        return
+    if is_promotable(node):
```
gao-yan, line 833: IMO as long as "meta promotable=" is explicitly set by the user, no matter what value it is. we should probably leave it and return, since that's what is apparently needed by the user.

liangxin1300: Already changed
Thanks!

crmsh/cibconfig.py
```diff
@@ -816,26 +816,55 @@ def postprocess_cli(node, oldnode=None, id_hint=None):
...
+    """
+    if node.tag != "clone":
+        return
+    primitive_list = node.xpath('primitive')
+    if not primitive_list:
+        return
+    set_list = []
+    for meta_item in ["promotable", "interleave"]:
+        if not is_attr_set(node, meta_item):
+            set_list.append(meta_item)
```
gao-yan, line 836: IIUC we could just generally add `interleave=true` as long as it's not set here, no matter if it's promotable clone or not?

gao-yan (approved):
Generally looks good to me. Thanks for the nice work, @liangxin1300 !

Only a question about possibly generic setting of `interleave=true` for any types of clones in below.

crmsh/cibconfig.py
```diff
@@ -816,26 +816,59 @@ def postprocess_cli(node, oldnode=None, id_hint=None):
...
+    ra_inst = RAInfo(p.get('class'), p.get('type'), p.get('provider'))
+    ra_actions_dict = ra_inst.actions()
+    if ra_actions_dict and "promote" in ra_actions_dict and "demote" in ra_actions_dict:
+        meta_node = get_set_nodes(node, "meta_attributes", create=True)[0]
+        for item in set_list:
+            set_attr(meta_node, item, "true")
+    # Add interleave=true as long as it's not set, no matter if it's promotable clone or not
+    elif "interleave" in set_list:
+        meta_node = get_set_nodes(node, "meta_attributes", create=True)[0]
+        set_attr(meta_node, "interleave", "true")
```
gao-yan, line 850: Somehow if with a consolidated logic, we shouldn't have to call set_attr() for "interleave" separately for different cases, but I understand this way also works :-)

gao-yan (approved):
LGTM, thanks @liangxin1300

PR #1067, opened 2022-11-24
Dev: bootstrap: add timeout to wait_for_cluster and wait_for_resource

### LINKED ISSUE #885: Should set a timeout value for wait_for_cluster and wait_for_resource functions
When calling `wait_for_cluster` to wait node online, or calling `wait_for_resource` to wait ra started,
should set a timeout value, not allow to wait forever, should raise an exception when reach the timeout

nicholasyang2022:
Fix #885.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -33,7 +33,7 @@ from . import corosync
 from . import tmpfiles
 from . import lock
 from . import userdir
-from .constants import SSH_OPTION, QDEVICE_HELP_INFO, STONITH_TIMEOUT_DEFAULT, REJOIN_COUNT, REJOIN_INTERVAL, PCMK_DELAY_MAX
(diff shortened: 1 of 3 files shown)


liangxin1300:
Better to create a separate commit for the unit test code

crmsh/bootstrap.py
```diff
@@ -679,7 +685,7 @@ def init_cluster_local():
         logger.warning("You should change the hacluster password to something more secure!")
 
     start_pacemaker(enable_flag=True)
-    wait_for_cluster()
+    wait_for_cluster(timeout_ms=120000)
```
liangxin1300, line 688: Since this number is used at many places, better to use a constant in constants.py

liangxin1300: Or, how about wait this time as a default value?

liangxin1300 (approved):

PR #1075, opened 2022-11-26
Change: store the upgrade_seq in /etc/crm/

aleksei-burlakov:
Previously it was `~/hacluster/crmsh/`
We don't want hardcoded users, let's store it in /etc/crm/

### DIFF
--- a/crmsh/upgradeutil.py
+++ b/crmsh/upgradeutil.py
@@ -12,7 +12,7 @@ import crmsh.utils
 
 # pump this seq when upgrade check need to be run
 CURRENT_UPGRADE_SEQ = (1, 0)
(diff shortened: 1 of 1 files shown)


aleksei-burlakov:
@nicholasyang2022 , @liangxin1300 what do you think if we move it to /etc/crm ?

nicholasyang2022:
I prefer `/var/lib/crmsh`. It is some kind of data, which is not expected to be modified manually.

aleksei-burlakov:
> I prefer `/var/lib/crmsh`. It is some kind of data, which is not expected to be modified manually.

I agree, the main idea not to keep it in ~hacluster

liangxin1300 (approved):

PR #1076, opened 2022-11-28
Dev: ui_context: redirect `foo -h`/`foo --help` to `help foo` (bsc#1205735)

nicholasyang2022:
So that a consistent help text can be obtained from both help subcommand
and --help argument.

One exception is `report -h`, since its output has more detail than
 `help report`.


liangxin1300:
Please see https://github.com/ClusterLabs/crmsh/blob/master/crmsh/ui_context.py#L50
and https://github.com/ClusterLabs/crmsh/blob/master/crmsh/ui_context.py#L81

Can you make the solution as a uniformed one?

I mean when the sub-command not use argparse like `cluster init/join/remove` and `crm report`, other sub-commands, when using like `crm sub-command -h/--help`, auto translate this as `crm help sub-command`?

liangxin1300 (approved):
LGTM, nice work!

Thank you!

PR #1089, opened 2022-12-12
Dev: utils: add auto_convert_role flag for handle_role_for_ocf_1_1 function

liangxin1300:
## Problem
When cib schema version support OCF 1.1 and set `OCF_1_1_SUPPORT = yes` in /etc/crm/crm.conf, if origin cib configuration contain original Master/Slave for role, command `crm configure show` will be messy:
```
INFO: Convert deprecated "Master" to "Promoted"
node 1: 15sp4-1
node 2: 15sp4-2
xml <primitive id="st" class="ocf" provider="pacemaker" type="Stateful"> \
  <operations> \
    <op name="monitor" role="Master" interval="10s" timeout="20s" id="st-monitor-10s"/> \
    <op name="start" timeout="20s" interval="0s" id="st-start-0s"/> \
    <op name="stop" timeout="20s" interval="0s" id="st-stop-0s"/> \
    <op name="promote" timeout="10s" interval="0s" id="st-promote-0s"/> \
    <op name="demote" timeout="10s" interval="0s" id="st-demote-0s"/> \
  </operations> \
</primitive>
```

## Solution
Add `utils.auto_convert_role` flag to control role convert, when doing configuring from cli, this flag value is True, otherwise, it's False.
Then, `crm configure show` will keep origin cib configuration unchanged

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -858,9 +858,11 @@ def parse_cli_to_xml(cli, oldnode=None):
     complete = False
     comments = []
     if isinstance(cli, str):
+        utils.auto_convert_role = False
         for s in lines2cli(cli):
             node = parse.parse(s, comments=comments)
     else:  # should be a pre-tokenized list
+        utils.auto_convert_role = True
         complete = True
         node = parse.parse(cli, comments=comments, ignore_empty=False, complete_advised=complete)
     if node is False:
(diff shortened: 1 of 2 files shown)


crmsh/cibconfig.py
```diff
@@ -858,9 +858,11 @@ def parse_cli_to_xml(cli, oldnode=None):
     complete = False
     comments = []
     if isinstance(cli, str):
+        utils.auto_convert_role = False
```
gao-yan, line 861: I'm kind of confused about the context here and why automatic conversion of the role names would break parsing and impact `crm configure show`. But as long as this resolves that :)

gao-yan (approved):
LGTM

Issue #413, opened 2019-01-23
Better to articulate `crm cluster stop` INFO output


zzhou1:
~ # crm cluster stop
INFO: Cluster services stopped

Basically, the term of "services" can be systemd services, ie, pacemaker.service, corosync.service, sbd.service. Also, it is widely used to refer to cluster resource agents to the end user.

So, better to improve INFO to clarify about this command. Eg, it stops the whole stack, both resource agents and systemd pacemaker/corosync.service/services.

I can understand the code. It stops corosync.service. The consequence result is a graceful shutdown of pacemaker which stops all resource agents on top.

krig:
Good idea, yes :)

PR #1106, opened 2022-12-28
Dev: ui_cluster: Improve cluster start/stop INFO

liangxin1300:
s/Cluster services/The cluster stack/
    
 To clarify that it starts/stops the whole stack, both resource agents and systemd pacemaker/corosync.service/services.

Related issue #413


zzhou1 (changes requested):
There are several other places to use "Cluster services" as well.

liangxin1300:
> There are several other places to use "Cluster services" as well.

No need to change info for `start` command, right?

zzhou1 (approved):

PR #1110, opened 2023-01-06
Dev: behave: Run all cases twice under normal user

liangxin1300:
Changes:
- Add `-u` option for ./test/run-functional-tests， to run behave under normal user `alice`, which will be created when the container is set up
- Run each behave functional test case twice, one under root, one under normal user `alice`
- Remove `test/features/bootstrap_init_join_remove_alice.feature`
- Add `sudo` before most of commands


liangxin1300:
loop #1009

aleksei-burlakov:
I saw you were working on Saturday=) I really like the change except for one thing. In the function `run_command` you unconditionally add sudo. I think it's fair for now if we want to see the change in sle 15.5. However, I also think it's a fair point to use sudo only when it's necessary in the future.

PR #1116, opened 2023-01-28
Dev: utils: Avoid using magic number

liangxin1300:


nicholasyang2022:
I think we can not assume that the gid is 90. We should use group name in text form.

liangxin1300:
Changed

nicholasyang2022 (approved):

PR #1122, opened 2023-02-09
Dev: utils: Add sudo for the commands in user hints

liangxin1300:


zzhou1 (approved):
Nice guide to the end user

PR #1123, opened 2023-02-12
Dev: bootstrap: allow the cluster to operate with ssh session under non-root sudoer (jsc#PED-290)

nicholasyang2022:
Many servers do not permit root to login from ssh for enhanced security.
This change enable crmsh to operate the cluster with this setup.

The implementation is to do most of work as root, and su to a non-root user,
which is only used for ssh login, only when a command is needed to be executed
on a remote node. The remote command is executed in a `sudo /bin/sh` shell, so
it also has privilege.

This relies on crmsh being called with sudo, so that crmsh will be able to
determine which non-root user to su to when it needs to create a ssh
session. When crmsh is not called with sudo, it should work the same way
as before.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -937,7 +937,7 @@ def _load_core_hosts() -> typing.Optional[typing.Tuple[typing.List[str], typing.
     return users, hosts
 
 
-def _fetch_core_host(local_user, remote_user, remote_host) -> typing.Tuple[typing.List[str], typing.List[str]]:
+def _fetch_core_hosts(local_user, remote_user, remote_host) -> typing.Tuple[typing.List[str], typing.List[str]]:
     # FIXME: when the remote_host is initialized with -d, this cmd will print extra debug log to stdout
     cmd = 'crm options show core.hosts'
     result = utils.su_subprocess_run(
@@ -1036,8 +1036,8 @@ def generate_ssh_key_pair_on_remote(
     # pass cmd through stdin rather than as arguments. It seems sudo has its own argument parsing mechanics,
     # which breaks shell expansion used in cmd
     cmd = '''
-    [ -f ~/.ssh/id_rsa ] || ssh-keygen -q -t rsa -f ~/.ssh/id_rsa -C "Cluster internal on $(hostname)" -N ''
-    '''
+[ -f ~/.ssh/id_rsa ] || ssh-keygen -q -t rsa -f ~/.ssh/id_rsa -C "Cluster internal on $(hostname)" -N ''
+'''
     result = utils.su_subprocess_run(
         local_sudoer,
         'ssh {} {}@{} sudo -H -u {} /bin/sh'.format(constants.SSH_OPTION, remote_sudoer, remote_host, remote_user),
@@ -1945,7 +1945,7 @@ def setup_passwordless_with_other_nodes(init_node):
                      tokens[1], tokens[2]))
         else:
             cluster_nodes_list.append(tokens[1])
-    user_list, host_list = _fetch_core_host(local_user, remote_user, init_node)
+    user_list, host_list = _fetch_core_hosts(local_user, remote_user, init_node)
     user_list.append(local_user)
     host_list.append(utils.this_node())
     _save_core_hosts(user_list, host_list, sync_to_remote=False)
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -870,7 +870,7 @@ def init_ssh():
     public_key_list = list()
     for i, node in enumerate(node_list):
         remote_user = _context.user_list[i]
-        ssh_copy_id(local_user, remote_user, node)
+        utils.ssh_copy_id(local_user, remote_user, node)
         # After this, login to remote_node is passwordless
         public_key = swap_public_ssh_key(node, local_user, remote_user, local_user, remote_user, add=True)
         public_key_list.append(public_key)
@@ -923,18 +923,17 @@ def _save_core_hosts(user_list: typing.List[str], host_list: typing.List[str], s
 
 
 def _load_core_hosts() -> typing.Optional[typing.Tuple[typing.List[str], typing.List[str]]]:
-    try:
-        li = config.get_option('core', 'hosts')
-    except ValueError:
-        return None
     users = list()
     hosts = list()
+    li = config.get_option('core', 'hosts')
+    if li == ['']:
+        return users, hosts
     for s in li:
-        s.split('@', 2)
-        if len(s) != 2:
+        parts = s.split('@', 2)
+        if len(parts) != 2:
             utils.fatal('Malformed config core.hosts: {}'.format(s))
-        users.append(s[0])
-        users.append(s[1])
+        users.append(parts[0])
+        hosts.append(parts[1])
     return users, hosts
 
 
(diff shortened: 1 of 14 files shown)


crmsh/utils.py
```diff
@@ -112,6 +112,8 @@ def user_of(host):
             node = item
         if host == node:
             return user
+    raise ValueError('Failed to get user of host {}'.format(host))
```
liangxin1300, line 115: Many behave test cases blocked here now.
When calling this for the first time, 'core.hosts' may not be set yet.

liangxin1300:
> qnetd & qdevice: need to use util functions for remote call instead of calling ssh directly

Yes, should consider qnetd node also set `PasswordAuthentication no` and `PermitRootLogin no`

crmsh/bootstrap.py
```diff
@@ -1784,7 +1903,7 @@ def setup_passwordless_with_other_nodes(init_node):
     remote_user = _context.user_list[0]
     local_user = _context.current_user
     cmd = "ssh {} {}@{} crm_node -l".format(SSH_OPTION, remote_user, init_node)
```
liangxin1300, line 1905: should add `sudo` before `crm_node`, if remote_user is not root

liangxin1300:
@nicholasyang2022 Please add more details both in the commit log and in the pr description, like the purpose of `config.core.hosts`, like what problem does this pr to address, and how?

crmsh/bootstrap.py
```diff
@@ -840,29 +835,110 @@ def init_ssh():
...
+    value = [f'{user}@{host}' for user, host in zip(user_list, host_list)]
+    config.set_option('core', 'hosts', value)
+    # TODO: it is saved in ~root/.config/crm/crm.conf, is it as suitable path?
+    config.save()
+    if sync_to_remote:
+        assert "'" not in value
+        crmsh.parallax.parallax_call(host_list, "crm options set core.hosts '{}'".format(', '.join(value)))
+
+
+def _load_core_hosts() -> typing.Optional[typing.Tuple[typing.List[str], typing.List[str]]]:
```
liangxin1300, line 899: Is this function called by anywhere?

crmsh/parallax.py
```diff
@@ -3,8 +3,11 @@
 
 
 import os
+import struct
```
liangxin1300, line 6: This import isn't used

liangxin1300:
The latest commit's behavior changed when join with sudo:
```
xin@15sp4-2:~> sudo crm cluster join -c 15sp4-1 -y
WARNING: chronyd.service is not configured to start at system boot.
INFO: Configuring SSH passwordless with root@15sp4-1
```
Above the sudoer is xin, so crmsh should configure passwordless using xin, not root

nicholasyang2022:
> The latest commit's behavior changed when join with sudo:
> 
> ```
> xin@15sp4-2:~> sudo crm cluster join -c 15sp4-1 -y
> WARNING: chronyd.service is not configured to start at system boot.
> INFO: SSH key for root does not exist, hence generate it now
> INFO: Configuring SSH passwordless with root@15sp4-1
> ```
> 
> Above the sudoer is xin, so crmsh should configure passwordless using xin, not root

@liangxin1300  We should also consider the traditional use case that root-based ssh is used. The behave is changed to:

* When users are explicitly specified in -N arguments, operate in non-root mode.
* When users are not specified in -N arguments, operate in root mode.
* When there are no -N arguments, ask for which mode to operate in

liangxin1300:
> > The latest commit's behavior changed when join with sudo:
> > ```
> > xin@15sp4-2:~> sudo crm cluster join -c 15sp4-1 -y
> > WARNING: chronyd.service is not configured to start at system boot.
> > INFO: SSH key for root does not exist, hence generate it now
> > INFO: Configuring SSH passwordless with root@15sp4-1
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Above the sudoer is xin, so crmsh should configure passwordless using xin, not root
> 
> @liangxin1300 We should also consider the traditional use case that root-based ssh is used. The behave is changed to:
> 
>     * When users are explicitly specified in -N arguments, operate in non-root mode.
> 
>     * When users are not specified in -N arguments, operate in root mode.
> 
>     * When there are no -N arguments, ask for which mode to operate in

I saw this when init:
```
alice@hanode1:~> sudo crm cluster init -y
INFO: crmsh is executed with sudo. Initialize the cluster as: ([root]/alice)?  root
```

I prefer not to ask user to confirm this, reasons:

- User use `alice`, configure sudoer, and add `sudo` before crm commands, he mostly don't want to use root back, for security concern.
- Under Azure instance, with `PermitRootLogin no`, it will failed on join node when running `sudo crm cluster join -c hanode1 -y`
```
alice@hanode2:~> sudo  crm cluster join -c hanode1 -y
WARNING: No NTP service found.
INFO: Configuring SSH passwordless with root@hanode1
/usr/bin/ssh-copy-id: INFO: Source of key(s) to be installed: "/root/.ssh/id_rsa.pub"
/usr/bin/ssh-copy-id: INFO: attempting to log in with the new key(s), to filter out any that are already installed
/usr/bin/ssh-copy-id: INFO: 1 key(s) remain to be installed -- if you are prompted now it is to install the new keys
root@hanode1: Permission denied (publickey).
ERROR: cluster.join: Failed to login to remote host root@hanode1
```
See https://github.com/ClusterLabs/crmsh/actions/runs/4184668113, all behave cases failed under Azure like ssh environment
@zzhou1 What do you think?

zzhou1:
> > > The latest commit's behavior changed when join with sudo:
> > > ```
> > > xin@15sp4-2:~> sudo crm cluster join -c 15sp4-1 -y
> > > WARNING: chronyd.service is not configured to start at system boot.
> > > INFO: SSH key for root does not exist, hence generate it now
> > > INFO: Configuring SSH passwordless with root@15sp4-1
> > > ```
> > > 
> > >   
> > > Above the sudoer is xin, so crmsh should configure passwordless using xin, not root
> > 
> > 
> > @liangxin1300 We should also consider the traditional use case that root-based ssh is used. The behave is changed to:
> > ```
> > * When users are explicitly specified in -N arguments, operate in non-root mode.
> > 
> > * When users are not specified in -N arguments, operate in root mode.
> > 
> > * When there are no -N arguments, ask for which mode to operate in
> > ```
> 
> I saw this when init:
> 
> ```
> alice@hanode1:~> sudo crm cluster init -y
> INFO: crmsh is executed with sudo. Initialize the cluster as: ([root]/alice)?  root
> ```
> 
> I prefer not to ask user to confirm this, reasons:
> 
> * User use `alice`, he configure sudoer for `alice`, and add `sudo` before crm commands, he mostly don't want to use root back, for security concern. (He might even feel confused, worry about the security, "Using root under alice?")
> * Under Azure instance, with `PermitRootLogin no`, it will failed on join node when running `sudo crm cluster join -c hanode1 -y`
> 
> ```
> alice@hanode2:~> sudo  crm cluster join -c hanode1 -y
> WARNING: No NTP service found.
> INFO: Configuring SSH passwordless with root@hanode1
> /usr/bin/ssh-copy-id: INFO: Source of key(s) to be installed: "/root/.ssh/id_rsa.pub"
> /usr/bin/ssh-copy-id: INFO: attempting to log in with the new key(s), to filter out any that are already installed
> /usr/bin/ssh-copy-id: INFO: 1 key(s) remain to be installed -- if you are prompted now it is to install the new keys
> root@hanode1: Permission denied (publickey).
> ERROR: cluster.join: Failed to login to remote host root@hanode1
> ```
> 
> See https://github.com/ClusterLabs/crmsh/actions/runs/4184668113, all behave cases failed under Azure like ssh environment
> 
> * As far as I know, few our customers know `-N` option. That means customer no chance to use non-root user if he not use `-N` option to specify it (with -y option)
> 
> @zzhou1 What do you think?

I agree with Xin.

nicholasyang2022:
@liangxin1300  The behavior is changed to:

* When users are explicitly specified in -N arguments, operate in non-root mode.
* When users are explicitly not specified in -N arguments, operate in root mode.
* When there are no -N arguments, determine whether to operate in non-root mode by checking if crmsh is called with sudo.

liangxin1300:
> * When there are no -N arguments, determine whether to operate in non-root mode by checking if crmsh is called with sudo.

Then under non-root user, with `PermitRootLogin no`, `sudo crm cluster join -c <init_node> -y` will failed

liangxin1300:
`crm cluster init` and
`crm cluster join -c <init_node>`
I think these two commands are the most used by users.
We should make sure they work even when `PermitRootLogin no`

liangxin1300 (approved):
Thank you @nicholasyang2022 
Great work!

PR #1124, opened 2023-02-13
DRAFT: Feature: add ansible checks

aleksei-burlakov:
Assume there is a 2-nodes cluster 192.168.145.11 and 192.168.145.12 and the superviser is the `node1`. There is passwordless access to **both** of them (from `node1` to `node1` and to `node2`) under root. Then as the root do (on the superwiser)
$ cp -r checks /usr/share/crmsh/
$ vim /usr/share/crmsh/checks/inventory # add 192.168.145.11 to the hosts: section
$ crm checks <dev|azure> list
$ crm checks <dev|azure> execute all

TODO in this PR: 
1) Enchancing syntax: specifying the hosts, 
they are listed in the inventory file now, but should be in the command, e.g.
$ crm checks azure execute all on `_host1 host2_`
3) Make it no-root.
4) Add tests and documentations

Future work:
Enhancing syntax: custom checks, not simply all, but a la carte
$ crm checks azure execute `corosync, pacemaker, sbd.*`
Besides, I don't like the keyword `all`, possibly
it should be replaced by * (asterisk)

I'm looking forward to you syntax suggestions


aleksei-burlakov:
I think I should be at least ansible 2.12, the tumbleweed has 2.14 and works well.

liangxin1300:
@aleksei-burlakov Thanks for this!
I will take a look after this week:)

aleksei-burlakov:
The world is not ready for this change

PR #1125, opened 2023-02-17
Dev: utils: Change the way to get pacemaker's version

liangxin1300:
- Use '/usr/sbin/pacemakerd --version'
- Raise fatal error if failed to get pacemaker's version
- Remove unused codes


gao-yan (approved):
LGTM. Thanks!

PR #1127, opened 2023-02-20
Dev: log: Redirect debug messages into stderr

liangxin1300:
- Dev: log: Redirect debug messages into stderr
If keep debug and info messages all in stdout, then some scripts that need to handle or analyze the output of the crm command might be messy.

- Dev: bootstrap: Don't save core.debug when saving core.hosts

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -19,12 +19,12 @@ class ConsoleCustomHandler(logging.StreamHandler):
     """
     A custom handler for console
 
-    Redirect ERROR message to sys.stderr
-    Redirect INFO/WARNING/DEBUG message to sys.stdout
+    Redirect ERROR/WARNING/DEBUG message to sys.stderr
+    Redirect INFO message to sys.stdout
     """
 
     def emit(self, record):
-        if record.levelno < logging.ERROR:
+        if record.levelno == logging.INFO:
             stream = sys.stdout
         else:
             stream = sys.stderr
--- a/crmsh/main.py
+++ b/crmsh/main.py
@@ -360,7 +360,7 @@ def run():
             options.batch = True
         user_args = parse_options()
         if config.core.debug:
-            print(utils.debug_timestamp())
+            logger.debug(utils.debug_timestamp())
         term.init()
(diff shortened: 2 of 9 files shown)


crmsh/log.py
```diff
@@ -24,7 +24,7 @@ class ConsoleCustomHandler(logging.StreamHandler):
     """
 
     def emit(self, record):
-        if record.levelno < logging.ERROR:
+        if record.levelno < logging.DEBUG:
```
nicholasyang2022, line 27: The int value of logging.DEBUG is the smallest. Are you going to send all logs to stderr?

crmsh/main.py
```diff
@@ -309,7 +309,11 @@ def compgen():
 def parse_options():
     opts, args = option_parser.parse_known_args()
     utils.check_space_option_value(opts)
+    debug_value_from_opt = opts.debug and not config.core.debug
     config.core.debug = "yes" if opts.debug else config.core.debug
+    if debug_value_from_opt:
+        config.set_option("core", "debug", "yes", internal=True)
+        atexit.register(config.back_to_default)
```
nicholasyang2022, line 316: An exit hook looks complicated. Can we just filter out those internal options before saving?

liangxin1300: > An exit hook looks complicated. Can we just filter out those internal options before saving?

I've tried this way, but since crmsh called `config.save()` just after init_ssh, if I filtered out like `core.debug`, then further debugging messages will missing while running `crm -d cluster init -y`

crmsh/log.py
liangxin1300: Thanks for reminding me! I forget that:)

liangxin1300: Already changed, only let INFO messages in stdout, others in stderr

crmsh/main.py
liangxin1300: Changed again.

When saving core.hosts, filter out core.debug, and set it back after save

nicholasyang2022 (approved):

PR #1135, opened 2023-02-27
Fix: qdevice: Unable to setup qdevice under non-root user (bsc#1208770)

nicholasyang2022:
* When --qnetd-hostname is provided as "user@host", use the specified user
* When --qnetd-hostname is provided as "host", use the same username as the local user

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -117,7 +117,11 @@ class UserOfHost:
         if cached is None:
             ret = self._user_of_impl(host)
             if ret is None:
-                return userdir.getuser()
+                user = userdir.get_sudoer()
+                if user:
+                    return user
+                else:
+                    return userdir.getuser()
             else:
                 self._cache[host] = ret
                 return ret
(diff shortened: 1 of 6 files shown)


liangxin1300:
```
INFO: Configure Qdevice/Qnetd:
INFO: Copy ssh key to qnetd node(root@qnetd-node)
INFO: Configuring SSH passwordless with root@qnetd-node
```
Above two lines are duplicated, keep one
```
root@qnetd-node's password: 
Check whether corosync-qnetd is installed on qnetd-node
```
Above line could be suppressed

crmsh/utils.py
```diff
@@ -3457,4 +3462,43 @@ def check_user_access(level_name):
     else:
         logger.error("Please run this command starting with \"sudo\"")
     raise TerminateSubCommand
+
+
+class HostUserConfig:
```
liangxin1300, line 3467: Please add comments for this class

nicholasyang2022: A docstring is added.

liangxin1300 (approved):

PR #1137, opened 2023-03-02
Dev: bootstrap: Add sudo before crm_node under non-root user on remote node

liangxin1300:
Or will got error:
```
xin@hanode2:~> sudo crm cluster join -c hanode1 -y
INFO: SSH key for xin does not exist, hence generate it now
INFO: Configuring SSH passwordless with xin@hanode1
Password: 
ERROR: cluster.join: Can't fetch cluster nodes list from hanode1: error: Could not connect to controller: Permission denied
```

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1927,7 +1927,7 @@ def setup_passwordless_with_other_nodes(init_node):
     # Fetch cluster nodes list
     remote_user = _context.user_list[0]
     local_user = _context.current_user
-    cmd = 'ssh {} {}@{} PATH=\\"\\$PATH\\":/usr/sbin:/sbin crm_node -l'.format(SSH_OPTION, remote_user, init_node)
+    cmd = f'ssh {SSH_OPTION} {remote_user}@{init_node} sudo crm_node -l'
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -1927,7 +1927,8 @@ def setup_passwordless_with_other_nodes(init_node):
     # Fetch cluster nodes list
     remote_user = _context.user_list[0]
     local_user = _context.current_user
-    cmd = 'ssh {} {}@{} PATH=\\"\\$PATH\\":/usr/sbin:/sbin crm_node -l'.format(SSH_OPTION, remote_user, init_node)
+    remote_sudo = "" if remote_user == 'root' else "sudo "
+    cmd = f'ssh {SSH_OPTION} {remote_user}@{init_node} PATH=\\"\\$PATH\\":/usr/sbin:/sbin {remote_sudo}crm_node -l'
```
nicholasyang2022, line 1931: We can just use `sudo crm_node -l` here. It is not needed any longer to append sbin directories to $PATH. And I think it is OK to prepend sudo even if remote_user is root.

nicholasyang2022 (approved):

PR #1139, opened 2023-03-02
replace 2&>1 by 2>&1 and swap with >/dev/null

peutch:
Hi,
It should be better like this no? Or at least apply for the replace.

Thanks!


liangxin1300 (approved):
Thank you!

PR #1143, opened 2023-03-13
Fix: parallax: Use 'sudo bash -c' when executing commands via sudoer (bsc#1209192)

liangxin1300:
Or, some commands like `rm -rf /etc/corosync/qdevice/net/*` might fail.

Also add more debug messages when configuring qdevice.

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -273,8 +273,12 @@ class LoggerUtils(object):
             self.logger.addHandler(console_handler)
 
     def log_only_to_file(self, msg, level=logging.INFO):
-        with self.only_file():
-            self.logger.log(level, msg)
+        from .config import core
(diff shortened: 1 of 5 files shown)


nicholasyang2022:
Should we add a case in behave test for the use case described in bsc#1209192?

liangxin1300:
> Should we add a case in behave test for the use case described in bsc#1209192?

Yes, I did, in https://github.com/ClusterLabs/crmsh/pull/1121
I found this issue when CI tried to re-setup qdevice.
When I applied this PR to #1121 , the problem was gone.
We can merge this PR first, then we can merge 1121 when we handle all non-root issues.

nicholasyang2022 (approved):

PR #1145, opened 2023-03-15
Dev: bootstrap: Change user shell for hacluster on remote node, in init_ssh_impl function

liangxin1300:
When crmsh detecting and configuring passwordless for hacluster for the old cluster, the user shell on remote node not changed.

This PR fix this problem.
And add a behave functional case to test this case 
`Scenario: Check hacluster's user shell`


nicholasyang2022:
1. Is there a behave test to check user shell after init and join, without upgrading?
2. With this change applied, can we remove the call to `change_user_shell` from `configure_ssh_key`? It is not related to ssh key.

liangxin1300:
>

liangxin1300:
> 1. Is there a behave test to check user shell after init and join, without upgrading?

No, now I add this check

>     2. With this change applied, can we remove the call to `change_user_shell` from `configure_ssh_key`? It is not related to ssh key.

It's related with `configure_ssh_key('hacluster')`, if remove this call, `configure_ssh_key('hacluster')` will failed at once
```
ERROR: cluster.init: Failed to run su hacluster -c 'ssh-keygen -q -f /var/lib/heartbeat/cores/hacluster/.ssh/id_rsa -C 'Cluster Internal on hanode1' -N ''': b''
```

PR #1146, opened 2023-03-16
Dev: utils: Check passwordless between cluster nodes

liangxin1300:
## Problem
When cluster nodes not configured passwordless for both root and hacluster, crmsh will give a warning each time it runs any crm commands such as:
```
WARNING: SSH connection to remote node 15sp4-2 failed.
parallax.Error: Exited with error code 255, Error output: parallax error: SSH requested a password. Please create SSH keys or
use the -A option to provide a password.
parallax error: SSH requested a password. Please create SSH keys or
use the -A option to provide a password.
root@15sp4-2: Permission denied (publickey,password,keyboard-interactive).
```
## Solution

Add function crmsh.utils.check_passwordless_between_nodes.
Return a list of hosts that require passwords between.
Call it when detected the hacluster passwordless was not configured:
```
if nodes and _is_upgrade_needed(nodes) and not crmsh.utils.check_passwordless_between_nodes(nodes):
```
If there were nodes not configured passwordless for root, record it in debug messages `instead of warning message` like:
```
DEBUG: There is no passwordless configured from 15sp4-2 to 15sp4-1 under 'root'
DEBUG: There is no passwordless configured from 15sp4-2 to 15sp4-3 under 'root'
DEBUG: There is no passwordless configured from 15sp4-1 to 15sp4-2 under 'root'
DEBUG: There is no passwordless configured from 15sp4-1 to 15sp4-3 under 'root'
```
## Other changes
- Reuse function `crmsh.utils.check_passwordless_between_nodes` in behave test case
- To reduce confusions, change 'upgrade' terminology to 'configuration fix'

### DIFF
--- a/crmsh/healthcheck.py
+++ b/crmsh/healthcheck.py
@@ -115,6 +115,9 @@ class PasswordlessHaclusterAuthenticationFeature(Feature):
     SSH_DIR = os.path.expanduser('~hacluster/.ssh')
     KEY_TYPES = ['ed25519', 'ecdsa', 'rsa']
 
+    def __str__(self):
+        return "Configure Passwordless for hacluster"
+
     def check_quick(self) -> bool:
         for key_type in self.KEY_TYPES:
             try:
--- a/crmsh/upgradeutil.py
+++ b/crmsh/upgradeutil.py
@@ -124,7 +124,7 @@ def upgrade_if_needed():
     if not crmsh.utils.can_ask(background_wait=False):
         return
     nodes = crmsh.utils.list_cluster_nodes(no_reg=True)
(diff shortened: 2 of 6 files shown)


zzhou1 (approved):

liangxin1300:
@nicholasyang2022 Do you agree with this logic?
```
if nodes and _is_upgrade_needed(nodes) and not crmsh.utils.check_passwordless_between_nodes(nodes):
```

nicholasyang2022:
Is what scenario will passwordless ssh be unavailable? Is it a valid use case? If not, there should be a warning.

PR #1149, opened 2023-03-19
Fix: validate ssh session when the users is determined by guessing (bsc#1209193)

nicholasyang2022:
In the following scenario, `user_of()` is not able to get a user from
`core.hosts`:

1. In the early stage of cluster node initialization
2. When a cluster is upgraded from a previous version
3. Called by external code, for example, the behave test

`user_of()` uses to return `$SUDO_USER` if it exists, or return the
current user if it doesn't. However, this breaks the following use case:

* A cluster is upgraded from a previous version and users operate it
  with sudo

In this case, `user_of()` returns `$SUDO_USER`, but `root` is expected.

This commit consists of the following changes:

1. `user_of()` returns `None` when it is not able to get a user from
   `core.hosts`. It is expected to be used only in early stage of
   initialization.

2. After the early stage of intialization, when an SSH session is
   needed, `user_pair_for_ssh()` is called to get both the local user
   and remote user that can be used for creating it. This function is
   implemented as follow:

    1. If `user_of()` returns valid users for both local and remote host,
       return them.
    2. Try create an ssh session with `root` as both local and remote
       user. If succeeds, return them.
    3. Try create an ssh session with `$SUDO_USER` as both local and remote
       user. If succeeds, return them.

### DIFF
--- a/crmsh/corosync.py
+++ b/crmsh/corosync.py
@@ -107,11 +107,7 @@ def query_qnetd_status():
         raise ValueError("host for qnetd not configured!")
 
     # Configure ssh passwordless to qnetd if detect password is needed
-    local_user = utils.user_of(utils.this_node())
-    try:
-        remote_user = utils.user_of(qnetd_addr)
-    except ValueError:
-        remote_user = 'root'
+    local_user, remote_user = utils.user_pair_for_ssh(qnetd_addr)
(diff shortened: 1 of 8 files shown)


liangxin1300 (approved):
Looks good to me!
Thank!

PR #1153, opened 2023-03-23
 Dev: bootstrap: Generate the public key on the remote if it does not exist

liangxin1300:


nicholasyang2022:
Does it do any harm to walk through all the init process? Splitting the implementation makes the code more complex.

liangxin1300:
So, I think there are two solutions:
1. skip non hacluster part
2. generate public key based on private key, on peer node

liangxin1300:
> So, I think there are two solutions:
> 
>     1. skip non hacluster part
> 
>     2. generate public key based on private key, on peer node
> 
> 
> Which one do you prefer?

@nicholasyang2022 ,I already changed as the second solution

crmsh/bootstrap.py
```diff
@@ -1049,6 +1049,20 @@ def generate_ssh_key_pair_on_remote(
     )
     if result.returncode != 0:
         raise ValueError(codecs.decode(result.stdout, 'utf-8', 'replace'))
+
+    cmd = '''
+[ -f ~/.ssh/id_rsa.pub ] || ssh-keygen -y -f ~/.ssh/id_rsa > ~/.ssh/id_rsa.pub
+'''
```
nicholasyang2022, line 1055: This can be append to the previous `cmd` and called with one ssh session, making it faster.

liangxin1300: Changed

nicholasyang2022 (approved):

PR #1154, opened 2023-03-23
Fix: crm report: sustain if there are offline nodes (bsc#1209480)

aleksei-burlakov:
Without this patch the hb_report will crash trying to reach out to the offline nodes.


liangxin1300 (approved):
LGTM

Thank you! @aleksei-burlakov

PR #1166, opened 2023-04-04
Fix: help: Long time to load and parse crm.8.adoc

liangxin1300:
This issue was introduced from commit 7629864d;
For those commands that were using argparse, they should get the output when needed, instead of loading each time.


liangxin1300:
@nicholasyang2022 
Please take a look.

When it's ready, it need to backport to crmsh-4.3 and crmsh-4.4

nicholasyang2022:
It would be more readable if `set_lvl_name` is named at a higher abstraction level, for example, `set_long_lazy_load_source`.

liangxin1300:
> It would be more readable if `set_lvl_name` is named at a higher abstraction level, for example, `set_long_lazy_load_source`.

Changed

nicholasyang2022 (approved):

PR #1171, opened 2023-04-12
Fix: lock: Join node failed to wait init node finished (bsc#1210332)

liangxin1300:
This issue was introduced since
69b18cdb7e37581c0e3173b86bb6dde5f16b6ab0.
The join node cannot be excluded because it uses a different lock directory.


nicholasyang2022 (approved):
LGTM

PR #1176, opened 2023-04-25
Dev: behave: Add functional test for bsc#1210693

liangxin1300:
- Expected cluster setup via root ssh key by default
- Add \<user\>@ when joining for non-root case

### DIFF
--- a/test/features/steps/utils.py
+++ b/test/features/steps/utils.py
@@ -4,7 +4,7 @@ import tarfile
 import glob
 import re
 import socket
-from crmsh import utils, bootstrap, parallax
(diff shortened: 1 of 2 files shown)


test/features/steps/utils.py
```diff
@@ -46,7 +46,24 @@ def add_sudo(cmd):
     return cmd
 
 
+def _wrap_cmd_non_root(cmd):
```
nicholasyang2022, line 49: Comment is needed here to document what this function serves.

liangxin1300: Added

nicholasyang2022 (approved):

PR #1177, opened 2023-04-26
 Fix: bootstrap: crm cluster join default behavior change in ssh key handling (bsc#1210693)

nicholasyang2022:
Problem
=======

In previous versions, `sudo crm cluster join -c <host>` use user root
for ssh sessions. However, in crmsh 4.5.0, the sudoer user is used. This
change in default behavior may break existing scripts.

Solution
========

Keep the default behavior unchanged and require using `sudo crm cluster
join -c <user>@<host>` to authenticate as a non-root user.


liangxin1300 (approved):

liangxin1300:
LGTM, THANKS!

PR #1181, opened 2023-05-10
Fix: bootstrap: `init --qnetd-hostname` fails when username is not specified (bsc#1211200)

nicholasyang2022:
The previous implementation fails to fallback to the current user, when user is neither specified in command line arguments nor saved in `core.hosts`.


liangxin1300 (approved):
LGTM, thanks!

PR #1185, opened 2023-05-16
Dev: workflows: Temporarily disabled unit_test and submit job

liangxin1300:
- Since the code will change dramatically, unit_test is temporarily disabled.
- Since submit to network:ha-clustering:Factory will
  break down current functionality for corosync2, submit job is temporarily
  disabled; When the corosync3 package is available, enable this job again.


zzhou1 (approved):
Once Factory switched to corosync3, we would re_enable this unit_test with Tumbleweed docker image

PR #1194, opened 2023-05-25
 Dev: behave: use a standalone agent instead of crmsh.parallax to run commands

### LINKED ISSUE #1182: Refactor: remove the dependency to `crmsh.parallax` from functional test code
With #1177, crmsh simply uses the user specified in command line of config files for ssh session and fallback to *the curent user* when it is not specified. The complex logic of *guessing ssh user* in `UserOfHost._guess_user_for_ssh` is not needed any longer.

However, the functional test depends on `crmsh.parallax` to run commands on nodes. These commands may run before the cluster is setup, with users not saved in config files, so it still needs the *guessing* logic to pick a user for ssh session.

If the dependency to `crmsh.parallax` from functional test code is removed, the implementation of `crmsh.parallax` can be simplified, making it more reliable.

nicholasyang2022:
# Background

It is expected that crmsh uses the current user only unless another user
is specified explicitly. However, the code in functional tests needs operate
with different users to cover different use cases, requiring a complex
procedure to choose a user.

# Changes

1. A standalone agent is used to running command in functional test. It is
used only in function test, so the test code and production code is decoupled.

    Close #1182.

2. With the above changes, the user-selecting procedure is simplified in
production code. And several subcommands are found depending on the old
procedure and falls back to sudoer user implicitly when the current user fails
to create ssh session. This implicit fallback behavior is removed, the affected
subcommands are:

    * `crm cluster init qdevice --qnetd-hostname <host>`
    * `crm cluster geo_join --cluster-node <host>`
    * `crm cluster geo_init_arbitrator --cluster-node <host>`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -2738,7 +2738,7 @@ def geo_fetch_config(node):
         try:
             local_user = utils.user_of(utils.this_node())
         except utils.UserOfHost.UserNotFoundError:
-            local_user = userdir.getuser()
+            local_user = user
         remote_user = user
     else:
         try:
@@ -2752,6 +2752,10 @@ def geo_fetch_config(node):
     configure_ssh_key(local_user)
     logger.info("Retrieving configuration - This may prompt for %s@%s:", remote_user, node)
     utils.ssh_copy_id(local_user, remote_user, node)
+    user_by_host = utils.HostUserConfig()
+    user_by_host.add(local_user, utils.this_node())
+    user_by_host.add(remote_user, node)
+    user_by_host.save_local()
     cmd = "tar -c -C '{}' .".format(BOOTH_DIR)
     with tempfile.TemporaryDirectory() as tmpdir:
         pipe_outlet, pipe_inlet = os.pipe()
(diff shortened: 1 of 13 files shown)


nicholasyang2022:
This pull request requires a change in the container image used by github actions. The commit `switch to the new container image` is used to switch to a temporary image to run the actions in this pull request. It should not be merged.

crmsh/bootstrap.py
```diff
@@ -2731,12 +2733,15 @@ def bootstrap_init_geo(context):
...
+    print(node)
     user, node = utils.parse_user_at_host(node)
+    print(user)
     if user is not None:
         try:
             local_user = utils.user_of(utils.this_node())
         except utils.UserOfHost.UserNotFoundError:
-            local_user = userdir.getuser()
+            local_user = user
+        print(local_user)
```
liangxin1300, line 2744: These 3 `print` are here on purpose, or forget to remove?

test/features/steps/behave_agent.py
liangxin1300, line 1: Is this file the same with `/features/steps/behave_agent.py`?
Try to keep one

nicholasyang2022: Currently, they are not identical. In the container side, only the server part is used.

I tried to keep the server side as simple as possible and do not foresee any need to modify it. I prefer to use this simpler way unless there is a real need to change it frequently.

In fact, the client side has been modified in the development of this PR, making them not identical.

crmsh/bootstrap.py
nicholasyang2022: They are forgotten.

liangxin1300:
LGTM

Thanks!

liangxin1300 (approved):

PR #1197, opened 2023-06-07
Dev: crmcheck: (new feature) call trento to do checks

aleksei-burlakov:
This commit adds two new commands

$ crm check list
$ crm check execute <check-id1> [check-id2|...]

How it works?
It gets the trento-wand key from the crm.conf and sends the rest-api request together with the key to the running wanda (wanda address is in crm.conf). Then it receives the response and shows the result.

If there is no wanda key, it gets the credentials from the crm.conf, send them to the trento-web (web address is in crm.conf), and it gets the wanda key in return. The rest is the same.

The easiest way to try it:
$ zypper in docker-compose
$ git clone https://github.com/ClusterLabs/crmsh.git
$ cd crmsh
$ git fetch origin pull/<ID>/head:rest-api-trento
$ git checkout rest-api-trento
$ cd test
$ ./run-functional-tests 4
$ docker exec -ti hanode1 bash
$ crm check list
$ crm check execute 00081D 156F64

Should be easy, but if it fails, please ask me.

### DIFF
--- a/crmsh/config.py
+++ b/crmsh/config.py
@@ -300,7 +300,16 @@ DEFAULTS = {
         'single_node': opt_boolean('no'),
         'sanitize_rule': opt_string('passw.*'),
         'verbosity': opt_string('0')
-    }
+    },
+    'trento': {
+        'web_url': opt_string('http://127.0.0.1:4000'),
+        'wanda_url': opt_string('http://127.0.0.1:4001'),
+        'wanda_key': opt_string(''),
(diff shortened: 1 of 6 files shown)


nicholasyang2022:
1. `requests` should be added as dependency in `setup.py`.
2. The access token should be cached and and reused.
3. Instead of saving cleartext password to disk, a refresh token should be obtained and saved.

PR #1198, opened 2023-06-12
Dev: ui_cluster: Use 'CustomAppendAction' instead of 'append' argparse action

liangxin1300:
`CustomAppendAction` is a custom class for argparse append action:
    - Flatten the value like '-s "/dev/sda1;/dev/sda2"'
    - Detect duplicated input

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -4,7 +4,7 @@
 
 import sys
 import re
-from argparse import ArgumentParser, RawDescriptionHelpFormatter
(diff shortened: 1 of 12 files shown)


crmsh/ui_cluster.py
```diff
@@ -68,6 +68,15 @@ def get_cluster_name():
     return cluster_name
 
 
+class CustomAppendAction(Action):
```
nicholasyang2022, line 71: The name should describe what it does more precisely.

nicholasyang2022 (approved):

PR #1201, opened 2023-06-19
medium: ui_node: fix cib rollback race on node standby

lge:
node standby / online had a race:
The current cib is queried ("base of change"), then manipulated to create,delete or change the values of, the "standby" nvpair(s). Then the current cib is queried *again* ("base of diff"), this time to be used as the "original", to produce a cib diff, which is then send to pacemaker to be applied.

Any changes to the cib between "base of change" and "base of diff" will be "rolled back", which can include master-scores, or operation results.

This has lead to attempted demotion of already stopped resources, resulting in spurious operation failures.

Fixed by using the same cib as basis for both the changes, and the diff. Currently this fix uses copy.deepcopy() on an already parsed cib object.

If someone feels like "optimizing" things, it would be more efficient to use the original xml result in tempfile form, and pass that to crm_diff directly, instead of re-serializing to xml twice.

### DIFF
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -3,6 +3,7 @@
 # See COPYING for license information.
 
 import re
+import copy
 import subprocess
 from argparse import ArgumentParser, RawDescriptionHelpFormatter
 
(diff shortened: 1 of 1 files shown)


liangxin1300 (approved):
Good catch!
Thank you @lge !

PR #1205, opened 2023-06-28
Fix: bootstrap: fix the validation of option -N and -c (bsc#1212436) 

nicholasyang2022:
These options should be able to input interactively on prompt. So
they are not required to present in command line args and the
parsing should be deferred.


liangxin1300:
Looks good.
Please backport to crmsh-4.5 branch

Thanks!

liangxin1300 (approved):

PR #1206, opened 2023-06-30
Dev: behave: add coverage and codecov.io for behave tests

nicholasyang2022:
Collect coverage data by patching the entrance file `/usr/sbin/crm` and use codecov.io to collect and combine the result from concurrent runs of different tests.


liangxin1300 (approved):
LGTM

Thanks!

PR #1212, opened 2023-07-06
Fix: upgradeutil: support the change of path of upgrade_seq in crmsh-4.5 (bsc#1213050)

nicholasyang2022:
Copy the seq file from the old path `~hacluster/crmsh/` to the new one `/var/lib/crmsh`, so that it can be found when upgraded to 4.5.


liangxin1300 (approved):
Nice fix!

Thanks!

PR #1219, opened 2023-07-25
Dev: requirements: remove parallax

nicholasyang2022:
parallax is not a dependency anymore and has already been removed in setup.py. It should also be removed from requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,3 @@
 lxml
 PyYAML
 python-dateutil
-parallax


liangxin1300 (approved):

liangxin1300:
Why codecov failed?

nicholasyang2022:
> Why codecov failed?

We upload 22 reports in each run of the pipeline. Sometimes it exceeds the rate limit of some web API.

PR #1222, opened 2023-07-31
 Fix: upgradeutil: reduce the timeout for getting sequence from remote node (bsc#1213797)

nicholasyang2022:
During a rolling upgrade, some of the cluster nodes may be unreachable.
In certain cases, it takes a long time to wait for a EHOSTUNREACH error.
Reducing the timeout helps crmsh to detect unreachable nodes faster and
to give up quickly.


crmsh/upgradeutil.py
```diff
@@ -137,8 +147,7 @@ def upgrade_if_needed():
         return
     nodes = crmsh.utils.list_cluster_nodes(no_reg=True)
     if nodes is not None and len(nodes) > 1 \
-            and _is_upgrade_needed(nodes) \
-            and not crmsh.utils.check_passwordless_between_nodes(nodes):
```
liangxin1300, line 141: Then `crmsh.utils.check_passwordless_between_nodes` is useless, should be removed

liangxin1300 (approved):

PR #1224, opened 2023-07-31
[crmsh-4.5] Fix: upgradeutil: reduce the timeout for getting sequence from remote node (bsc#1213797)

nicholasyang2022:
backport

* #1222


crmsh/upgradeutil.py
```diff
@@ -137,8 +147,7 @@ def upgrade_if_needed():
         return
     nodes = crmsh.utils.list_cluster_nodes(no_reg=True)
     if nodes is not None and len(nodes) > 1 \
-            and _is_upgrade_needed(nodes) \
-            and not crmsh.utils.check_passwordless_between_nodes(nodes):
```
liangxin1300, line 141: Here is the same. `check_passwordless_between_nodes` is useless now, should be removed

liangxin1300 (approved):

PR #1226, opened 2023-08-02
Dev: corosync: Enable logging.to_logfile

liangxin1300:


liangxin1300:
> Why do we change this config?

I think it's better to have a specific log file for corosync, and this is what pcs did
I also communicate with Emil, and he agreed with this change

nicholasyang2022 (approved):

PR #1229, opened 2023-08-04
Fix: codecov: use token in plaintext

nicholasyang2022:
As secrets in Github Actions are not provided to pull requests, the codecov token has to be provided in plaintext. This token is only authorized to upload reports to codecov.

See https://github.com/codecov/codecov-action/issues/557.


liangxin1300 (approved):
Nice improvement!
Thanks!

PR #1232, opened 2023-08-10
 Fix: ui_cluster: Improve the process of 'crm cluster stop' (bsc#1213889)

liangxin1300:
- Fix bsc#1213889: don't try to get DC when local service is unavailable
- Split into small functions to make the whole process clear
- Support checking and setting dlm option on remote node

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -190,40 +190,73 @@ class Cluster(command.UI):
         for node in node_list:
             logger.info("The cluster stack started on {}".format(node))
 
+    @staticmethod
+    def _node_ready_to_stop_cluster_service(node):
+        """
+        Check if the specific node is ready to stop cluster service
+
+        If both corosync.service and pacemaker.service is active, return True
+        If some services started, stop them first and return False
+        """
+        corosync_active = utils.service_is_active("corosync.service", remote_addr=node)
+        sbd_active = utils.service_is_active("sbd.service", remote_addr=node)
+        pacemaker_active = utils.service_is_active("pacemaker.service", remote_addr=node)
+
+        if not corosync_active:
+            if sbd_active:
+                utils.stop_service("corosync", remote_addr=node)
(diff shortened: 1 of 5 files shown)


crmsh/ui_cluster.py
```diff
@@ -190,25 +190,36 @@
...
-                    logger.info("The cluster stack stopped on {}".format(node))
-                else:
-                    logger.info("The cluster stack already stopped on {}".format(node))
-                node_list.remove(node)
-            elif not utils.service_is_active("pacemaker.service", remote_addr=node):
+    @staticmethod
+    def _stop_partial_service(node):
+        """
+        Stops partial cluster service
+        """
```
nicholasyang2022, line 197: Can you rename this function or change the docstring to make the meaning of return value more clear?

crmsh/ui_cluster.py
```diff
@@ -190,25 +190,36 @@ def do_start(self, context, *args):
...
+
+        return True
+
+    @staticmethod
+    def _wait_for_dc():
+        """
+        Wait for the cluster's DC to become available
+        """
+        if not utils.service_is_active("pacemaker.service"):
             return
```
nicholasyang2022, line 223: More comments are needed here to document why DC is waited for only when pacemaker.service is active.

nicholasyang2022 (changes requested):

crmsh/ui_cluster.py
```diff
@@ -190,25 +190,40 @@ def do_start(self, context, *args):
...
+            return False
+
+        return True
+
+    @staticmethod
+    def _wait_for_dc():
+        """
+        Wait for the cluster's DC to become available
+        """
+        # Cannot get DC when local service is not available
```
nicholasyang2022, line 225: Why can we ignore DC when local service is not started? What if DC is running on another node?

liangxin1300: Yeah, better to consider the remote case

PR #1241, opened 2023-09-12
[crmsh-4.6] Dev: refactor shell calling routines

nicholasyang2022:
1. Move shell calling routines to module `crmsh.sh` and group them into 3 classes.
2. Move `ServiceManager` from `utils` to a new module.
3. Move `UserOfHost` from `utils` to a new module.
4. Make the dependencies between `crmsh.sh`, `UserOfHost` and `ServiceManager` injectable.
5. Partially move ssh key related routines to module `crmsh.ssh_key`.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -44,6 +44,7 @@ from . import qdevice
 from . import parallax
 from . import log
 from .service_manager import ServiceManager
+from .sh import ShellUtils
 from .ui_node import NodeMgmt
 from .user_of_host import UserOfHost, UserNotFoundError
 
@@ -400,7 +401,7 @@ def invoke(*args):
     Return (boolean, stdout, stderr)
     """
     logger_utils.log_only_to_file("invoke: " + " ".join(args))
-    rc, stdout, stderr = utils.get_stdout_stderr(" ".join(args))
+    rc, stdout, stderr = ShellUtils().get_stdout_stderr(" ".join(args))
     if stdout:
         logger_utils.log_only_to_file("stdout: {}".format(stdout))
     if stderr:
@@ -527,7 +528,7 @@ def partprobe():
     #  ATT Training Engineer
     #  Data Center Engineer
     #  bheaton@suse.com
-    _rc, out, _err = utils.get_stdout_stderr("sfdisk -l")
+    _rc, out, _err = ShellUtils().get_stdout_stderr("sfdisk -l")
     disks = re.findall(r'^Disk\s*(/.+):', out, re.M)
     invoke("partprobe", *disks)
 
@@ -728,12 +729,12 @@ def init_cluster_local():
 
     # reset password, but only if it's not already set
     # (We still need the hacluster for the hawk).
-    _rc, outp = utils.get_stdout("passwd -S hacluster")
+    _rc, outp = ShellUtils().get_stdout("passwd -S hacluster")
     ps = outp.strip().split()[1]
     pass_msg = ""
     if ps not in ("P", "PS"):
         logger_utils.log_only_to_file(': Resetting password of hacluster user')
-        rc, outp, errp = utils.get_stdout_stderr("passwd hacluster", input_s=b"linux\nlinux\n")
+        rc, outp, errp = ShellUtils().get_stdout_stderr("passwd hacluster", input_s=b"linux\nlinux\n")
         if rc != 0:
             logger.warning("Failed to reset password of hacluster user: %s" % (outp + errp))
         else:
@@ -1519,7 +1520,7 @@ def init_cluster():
     """
     init_cluster_local()
 
-    _rc, nnodes = utils.get_stdout("crm_node -l")
+    _rc, nnodes = ShellUtils().get_stdout("crm_node -l")
     nnodes = len(nnodes.splitlines())
     if nnodes < 1:
         utils.fatal("No nodes found in cluster")
@@ -1877,7 +1878,7 @@ def update_expected_votes():
     nodecount = 0
     expected_votes = 0
     while True:
-        rc, nodelist_text = utils.get_stdout("cibadmin -Ql --xpath '/cib/status/node_state'")
+        rc, nodelist_text = ShellUtils().get_stdout("cibadmin -Ql --xpath '/cib/status/node_state'")
         if rc == 0:
             try:
                 nodelist_xml = etree.fromstring(nodelist_text)
@@ -2150,7 +2151,7 @@ def join_cluster(seed_host, remote_user):
         if ipv6_flag and not is_unicast:
             # for ipv6 mcast
             nodeid_dict = {}
-            _rc, outp, _ = utils.get_stdout_stderr("crm_node -l")
+            _rc, outp, _ = ShellUtils().get_stdout_stderr("crm_node -l")
             if _rc == 0:
                 for line in outp.splitlines():
                     tokens = line.split()
@@ -2172,7 +2173,7 @@ def join_cluster(seed_host, remote_user):
         invoke("corosync-cfgtool -R")
 
         # Ditch no-quorum-policy=ignore
-        _rc, outp = utils.get_stdout("crm configure show")
+        _rc, outp = ShellUtils().get_stdout("crm configure show")
         if re.search('no-quorum-policy=.*ignore', outp):
             invoke("crm_attribute --attr-name no-quorum-policy --delete-attr")
 
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -425,7 +425,7 @@ def crm_configure_load(action, configuration):
 
     configuration_tmpfile = utils.str2tmp(configuration)
     tmpfiles.add(configuration_tmpfile)
-    sh.auto_shell().get_stdout_or_raise_error(f"crm -F configure load {action} {configuration_tmpfile}")
+    sh.cluster_shell().get_stdout_or_raise_error(f"crm -F configure load {action} {configuration_tmpfile}")
 
 
 def wait_for_resource(message, resource, timeout_ms=WAIT_TIMEOUT_MS_DEFAULT):
@@ -459,7 +459,7 @@ def get_node_canonical_hostname(host: str) -> str:
     """
     Get the canonical hostname of the cluster node
     """
-    rc, out, err = sh.auto_shell().get_stdout_stderr_no_input(host, 'crm_node --name')
+    rc, out, err = sh.cluster_shell().get_rc_stdout_stderr_without_input(host, 'crm_node --name')
     if rc != 0:
         utils.fatal(err)
     return out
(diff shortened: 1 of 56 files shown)


crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+3 different implementations are provided:
+
+1. LocalShell allows to run command on local host as various users. It is the most feature-rich one, allowing
+   interactive I/O from/to the terminal.
+2. SshShell allows to run command on both local and remote hosts. When running on a remote host, it creates a direct
+   connection from a specified local_user to the destination host and user. User input from terminal is not allowed, as
+   the command is passed through the stdin.
+3. AutoShell runs command on cluster nodes. It leverages su, sudo and ssh to obtain a appreciated session on a
+   destination node. It is only available after ssh bootstrap as it depends on the knowledge about cluster node and user
+   configurations.
```
liangxin1300, line 15: Since this kind of shell is `only available after ssh bootstrap` and it's for the cluster nodes, how about renaming it as `ClusterShell`?

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
+"""Run shell commands.
+
+This module provides various methods to run shell commands, on both local and remote hosts, as current or another user.
+There many variant of the methods to allow fine-gain control of parameter passing and error handling.
+
+3 different implementations are provided:
+
+1. LocalShell allows to run command on local host as various users. It is the most feature-rich one, allowing
+   interactive I/O from/to the terminal.
+2. SshShell allows to run command on both local and remote hosts. When running on a remote host, it creates a direct
```
liangxin1300, line 10: I think `SshShell` should be renamed as `SSHShell`, since `SSH` is a specific word instead of an English word

crmsh/pyshim.py
```diff
@@ -0,0 +1,21 @@
...
+            nonlocal cached_return_value
+            key = (tuple(args), tuple(sorted(kwargs.items())))
+            try:
+                return cached_return_value[key]
+            except KeyError:
+                ret = f(*args, **kwargs)
+                cached_return_value[key] = ret
+                return ret
+
+        return wrapper
```
liangxin1300, line 21: crmsh already has `crmsh.cache` module, try to reuse it, or replace it

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+    def subprocess_run_no_input(self, host: str, user: typing.Optional[str], cmd: str, **kwargs):
+        assert 'input' not in kwargs and 'stdin' not in kwargs
+        assert host is None or host == self.local_shell.hostname()
+        if user is None:
+            user = 'root'
+        return self.local_shell.su_subprocess_run(user, cmd, **kwargs)
+
+
+def auto_shell():
+    return AutoShell(LocalShell(), user_of_host.instance())
```
liangxin1300, line 321: I see there are still some functions in utils.py that calling `subprocess.Popen` or `subprocess.call` like `ext_cmd` `ext_cmd_nosudo` `get_stdout` and `get_stdout_stderr`

How about merging them into this module?

crmsh/service_manager.py
```diff
@@ -0,0 +1,100 @@
...
+class ServiceManager(object):
+    """
+    Class to manage systemctl services
+    """
+
+    def __init__(self, shell: crmsh.sh.AutoShell = None):
+        if shell is None:
+            self._shell = crmsh.sh.auto_shell()
+        else:
+            self._shell = shell
```
liangxin1300, line 16:
```Python
self._shell = shell or crmsh.sh.auto_shell()
```

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+        return self.geteuid() == 0 or self.get_effective_user_name() == user
+
+    def su_subprocess_run(
+            self,
+            user: str,
+            cmd: str,
+            tty=False,
+            preserve_env: typing.Optional[typing.List[str]] = None,
+            **kwargs
+    ):
```
liangxin1300, line 105: Please add a doc string for this method to introduce what is this method for or when is the time to use it.

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+        else:
+            if rc not in success_exit_status:
+                to_raise = True
+        if not to_raise:
+            return Utils.decode_str(stdout).strip()
+        else:
+            raise CommandFailure(cmd, host, None, Utils.decode_str(stderr).strip())
+
+
+class LocalOnlyAutoShell(AutoShell):
```
liangxin1300, line 308: Here needs doc string

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+        else:
+            return self.local_shell.su_subprocess_run(
+                self.local_user,
+                'ssh {} {}@{} /bin/sh'.format(constants.SSH_OPTION, user, host),
+                input=cmd.encode('utf-8'),
+                **kwargs,
+            )
+
+
+class AutoShell:
```
liangxin1300, line 252: Here needs doc string

crmsh/sh.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def get_raw_stdout_stderr_no_input(self, host, cmd) -> typing.Tuple[int, bytes, bytes]:
+        result = self.subprocess_run_no_input(
+            host, None, cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        return result.returncode, result.stdout, result.stderr
+
+    def get_stdout_stderr_no_input(self, host, cmd) -> typing.Tuple[int, str, str]:
```
liangxin1300, line 284: The method name is not accurate since return value contains rc code, too

crmsh/bootstrap.py
```diff
@@ -828,24 +832,50 @@ def _parse_user_at_host(s: str, default_user: str) -> typing.Tuple[str, str]:
...
-    utils.start_service("sshd.service", enable=True)
-    configure_ssh_key(local_user)
+def init_ssh_impl(local_user: str, ssh_public_key: typing.Optional[ssh_key.Key], user_node_list: typing.List[typing.Tuple[str, str]]):
+    """ Configure passwordless SSH.
+
+    The local_user on local host will be configured.
+    If user_node_list is not empty, those user and host will also be configured.
+    If ssh_public_key is specified, it will be added to authorized_keys; if not, a new key pair will be generated for each node.
+    """
+    ServiceManager(sh.LocalOnlyAutoShell(sh.LocalShell())).start_service("sshd.service", enable=True)
```
liangxin1300, line 854: I noticed that `ServiceManager(sh.LocalOnlyAutoShell(sh.LocalShell()))` is used in several places. Would it be worthwhile to simplify/wrapper it?

crmsh/ssh_key.py
```diff
@@ -0,0 +1,83 @@
...
+            else:
+                with tempfile.NamedTemporaryFile('w', encoding='utf-8') as tmp:
+                    os.chown(tmp.fileno(), user_info.pw_uid, user_info.pw_gid)
+                    print(key.public_key(), file=tmp)
+                    cmd = "ssh-copy-id -f -i '{}' '{}@{}' &> /dev/null".format(tmp.name, user, host)
+                    logger.info("Configuring SSH passwordless with %s@%s", user, host)
+                    result = self._shell.local_shell.su_subprocess_run(self._shell.local_user, cmd, tty=True)
+            if result.returncode != 0:
+                raise Error(f'Failed configuring SSH passwordless with {user}@{host}.')
+            # TODO: error handling
```
liangxin1300, line 83: Please add more doc strings for the methods in this module

crmsh/user_of_host.py
```diff
@@ -0,0 +1,116 @@
...
+            return user, user
+        else:
+            return None
+
+
+_user_of_host_instance = UserOfHost()
+
+
+def instance():
+    return _user_of_host_instance
```
liangxin1300, line 116: Same for this module, needs more doc strings

crmsh/sh.py
liangxin1300:
```
    def get_stdout_stderr_raw
    def get_stdout_stderr
    def get_rc_and_error
    def get_stdout_or_raise_error
    def subprocess_run_no_input
    def get_raw_stdout_stderr_no_input
    def get_stdout_stderr_no_input
```
Do the names look confusing? How about
```
    def get_stdout_stderr_raw
    def get_stdout_stderr
    def get_rc_and_error
    def get_stdout_or_raise_error
    def subprocess_run_without_input
    def get_stdout_stderr_raw_without_input
    def get_stdout_stderr_without_input
```

crmsh/service_manager.py
nicholasyang2022: I think `crmsh.sh.auto_shell() if shell is None or shell` is better, as it is more explicit.

crmsh/user_of_host.py
nicholasyang2022: This class is just moved out of utils.py without significant changes. Adding doc strings is not trivial as it is required to inspect the detail, and I think it does not worth to spend time doing it in recent.

crmsh/pyshim.py
nicholasyang2022: functools.cache and crmsh.cache are quite different in features and designs.

1. functools.cache has simpler interface and does not support expiration and invalidation.
2. functools.cache is scoped by functions. crmsh.cache is global-scoped. A global-scoped cache requires selecting keys carefully, or limit its usage to avoid conflicts.

A feature-rich cache supporting all the use case of both functools.cache and crmsh.cache is usually implemented as a standalone library. It is an overkill to implement one in crmsh.

crmsh/ssh_key.py
nicholasyang2022: This module is under construction and significant changes are expected to implement ssh-agent feature.

crmsh/pyshim.py
liangxin1300: Good.
Better to leave above points as comments in this py file?

crmsh/sh.py
nicholasyang2022: I moved `get_stdout` and `get_stdout_stderr` to `crmsh.sh`.

For direct calls to `subprocess.Popen` or `subprocess.call`, most of them use some advanced features provided by `subprocess` module, for example, piping data between 2 subprocesses. There is no need to create wrapper for these usages, or the wrapper class will be as complex as `subprocess` module.

For `ext_cmd` and `ext_cmd_sudo`, it seem that they are related the config `core.user`. I did not spent much time to inspect it. So I'd like to keep them unchanged.

crmsh/bootstrap.py
nicholasyang2022: For now, the usage of `LocalOnlyAutoShell` is quite limited. It is only used by `ServiceManager` during the early stage of bootstrap.

I prefer not to create a wrapper for it, or the wrapper function may look more confusing.

liangxin1300 (approved):
Nice work!

Thanks!

PR #1248, opened 2023-09-22
Fix: constants: Add several resource meta attributes (bsc#1215319)

liangxin1300:
To align with pacemaker upstream, add:
'critical', 'allow-unhealthy-nodes' and 'container-attribute-target'


gao-yan (approved):
Looks good. Thanks!

PR #1253, opened 2023-09-25
Fix: utils: Add 'sudo' only when there is a sudoer(bsc#1215549)

liangxin1300:
backport #1247

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -185,7 +185,9 @@ class UserOfHost:
     def _guess_user_for_ssh(host: str) -> typing.Tuple[str, str]:
         args = ['ssh']
         args.extend(constants.SSH_OPTION_ARGS)
-        args.extend(['-o', 'BatchMode=yes', host, 'sudo', 'true'])
+        sudo_str = 'sudo ' if userdir.get_sudoer() else ''
+        cmd_str = f"-o BatchMode=yes {host} {sudo_str}true"
+        args.extend(shlex.split(cmd_str))
         rc = subprocess.call(
             args,
             stdin=subprocess.DEVNULL,
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -185,9 +185,10 @@ class UserOfHost:
     def _guess_user_for_ssh(host: str) -> typing.Tuple[str, str]:
         args = ['ssh']
         args.extend(constants.SSH_OPTION_ARGS)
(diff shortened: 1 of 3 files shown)


nicholasyang2022 (approved):

nicholasyang2022 (changes requested):
It is preferred to avoid using split to parse cmdline arguments to avoid shell injection.

liangxin1300:
> It is preferred to avoid using split to parse cmdline arguments to avoid shell injection.

Changed on all 3 PRs

Thanks for pointing this!

nicholasyang2022:
This change does not work. If whitespace characters exist in `host`, `shlex.split` will still split it. To avoid shell injection, arguments should not be composed into a command line without validation. For this case, it is preferred to operate on the argument array instead of command line string.

liangxin1300:
> This change does not work. If whitespace characters exist in `host`, `shlex.split` will still split it. To avoid shell injection, arguments should not be composed into a command line without validation. For this case, it is preferred to operate on the argument array instead of command line string.

Changed
Please take a look.
Thanks!

nicholasyang2022 (approved):
~~{"ssh", "true"} is not the same as {"ssh", "", "true"}. However, it seems that ssh ignore an empty arg and this does not cause any problem.~~

crmsh/utils.py
```diff
@@ -185,7 +185,8 @@ def _get_user_of_host_from_config(host):
     def _guess_user_for_ssh(host: str) -> typing.Tuple[str, str]:
         args = ['ssh']
         args.extend(constants.SSH_OPTION_ARGS)
-        args.extend(['-o', 'BatchMode=yes', host, 'sudo', 'true'])
+        sudo_str = 'sudo ' if userdir.get_sudoer() else ''
+        args.extend(['-o', 'BatchMode=yes', host, sudo_str, 'true'])
```
nicholasyang2022, line 189: 1. When passing arguments as array of strings, space is not needed. Space is the separator of arguments in shell. See `init main(int argc, char **argv)` and `execve(1)`.
2. An empty argument is not the same as non-existing argument. `argv = {"true"}` and  `argv = {"", "true"}` is not the same.

```suggestion
        if userdir.get_sudoer():
            args.extend(['-o', 'BatchMode=yes', host, 'sudo', 'true'])
        else:
            args.extend(['-o', 'BatchMode=yes', host, 'true'])
```

nicholasyang2022 (approved):

PR #1257, opened 2023-10-07
[crmsh-4.6] Fix: report: Pick up tarball suffix dynamically (bsc#1215438)

liangxin1300:
To address a regression introduced after #1256 , due to hard coding the target tarball suffix as 'bz2' in the crm history.


aleksei-burlakov (approved):
Thank you, LGTM!

PR #1259, opened 2023-10-07
[crmsh-4.4] Fix: report: Pick up tarball suffix dynamically (bsc#1215438)

liangxin1300:
backport #1257 
Fix regression after #1251


aleksei-burlakov (approved):
Thanks for backporting, LGTM!

PR #1261, opened 2023-10-11
[crmsh-4.6] Dev: bootstrap: implement ssh-agent support (jsc#PED-5774) 

nicholasyang2022:
Use Cases
=========

In a typical cloud-based deployment, a server may have password-based
authentication disabled for ssh, and an adminstrator's ssh public key
is added to authorized_keys during initialization.

For this case, it is impossible for crmsh to log into cluster node with
interactive authentication and create new key pairs for further
operations. Instead, crmsh can make use of the administrator's key, by
authenticating ssh session with ssh-agent forwarded form the
adminstrator's PC.

Usage Example
=============

```text
alice@alice-pc: ~> ssh -A root@node1
root@node1:~ # crm cluster init --use-ssh-agent -y
root@node1:~ # exit
alice@alice-pc: ~> ssh -A root@node2
root@node2:~ # crm cluster join --use-ssh-agent -c node1 -y
```

### DIFF
--- a/crmsh/sh.py
+++ b/crmsh/sh.py
@@ -21,6 +21,7 @@ be used.
 import logging
 import os
 import pwd
+import re
 import socket
 import subprocess
 import typing
@@ -95,15 +96,18 @@ class LocalShell:
     def get_effective_user_name() -> str:
         return pwd.getpwuid(LocalShell.geteuid()).pw_name
 
+    def __init__(self, additional_environ: typing.Dict[str, str] = None):
+        self.additional_environ = additional_environ
+        self.preserve_env = additional_environ.keys() if additional_environ is not None else None
+
     def can_run_as(self, user: str):
         return self.geteuid() == 0 or self.get_effective_user_name() == user
 
     def su_subprocess_run(
             self,
-            user: str,
+            user: typing.Optional[str],
             cmd: str,
             tty=False,
-            preserve_env: typing.Optional[typing.List[str]] = None,
             **kwargs,
     ):
         """Call subprocess.run as another user.
@@ -111,24 +115,30 @@ class LocalShell:
         This variant is the most flexible one as it pass unknown kwargs to the underlay subprocess.run. However, it
         accepts only cmdline but not argv, as the argv is used internally to switch user.
         """
-        if self.get_effective_user_name() == user:
+        if user is None or self.get_effective_user_name() == user:
             args = ['/bin/sh', '-c', cmd]
         elif 0 == self.geteuid():
             args = ['su', user, '--login', '-c', cmd]
             if tty:
                 args.append('--pty')
-            if preserve_env:
+            if self.preserve_env:
                 args.append('-w')
-                args.append(','.join(preserve_env))
+                args.append(','.join(self.preserve_env))
         else:
             raise AuthorizationError(
                 cmd, None, user,
                 f"non-root user '{self.get_effective_user_name()}' cannot switch to another user"
             )
-        logger.debug('su_subprocess_run: %s, %s', args, kwargs)
-        return subprocess.run(args, **kwargs)
+        if not self.additional_environ:
+            logger.debug('su_subprocess_run: %s, %s', args, kwargs)
+            return subprocess.run(args, **kwargs)
+        else:
+            logger.debug('su_subprocess_run: %s, env=%s, %s', args, self.additional_environ, kwargs)
+            env = dict(os.environ)
+            env.update(self.additional_environ)
+            return subprocess.run(args, env=env, **kwargs)
 
-    def get_rc_stdout_stderr_raw(self, user: str, cmd: str, input: typing.Optional[bytes] = None):
+    def get_rc_stdout_stderr_raw(self, user: typing.Optional[str], cmd: str, input: typing.Optional[bytes] = None):
         result = self.su_subprocess_run(
             user, cmd,
             input=input,
@@ -137,13 +147,13 @@ class LocalShell:
         )
         return result.returncode, result.stdout, result.stderr
 
-    def get_rc_stdout_stderr(self, user: str, cmd: str, input: typing.Optional[str] = None):
+    def get_rc_stdout_stderr(self, user: typing.Optional[str], cmd: str, input: typing.Optional[str] = None):
         rc, stdout, stderr = self.get_rc_stdout_stderr_raw(user, cmd, input.encode('utf-8') if input is not None else None)
         return rc, Utils.decode_str(stdout).strip(), Utils.decode_str(stderr).strip()
 
     def get_rc_and_error(
             self,
-            user: str,
+            user: typing.Optional[str],
             cmd: str,
     ) -> typing.Tuple[int, typing.Optional[str]]:
         """Run a command for its side effects. Returns (rc, error_message)
@@ -151,17 +161,8 @@ class LocalShell:
         If the return code is 0, outputs from the command will be ignored and (0, None) is returned.
         If the return code is not 0, outputs from the stdout and stderr is combined as a single message.
         """
-        if self.get_effective_user_name() == user:
-            args = ['/bin/sh', '-c', cmd]
-        elif self.geteuid() == 0:
-            args = ['su', user, '--login', '-c', cmd]
-        else:
-            raise AuthorizationError(
-                cmd, None, user,
-                f"non-root user '{self.get_effective_user_name()}' cannot switch to another user"
-            )
-        result = subprocess.run(
-            args,
+        result = self.su_subprocess_run(
+            user, cmd,
             stdin=subprocess.DEVNULL,
             stdout=subprocess.PIPE,
             stderr=subprocess.STDOUT,
@@ -174,7 +175,7 @@ class LocalShell:
 
     def get_stdout_or_raise_error(
             self,
-            user: str,
+            user: typing.Optional[str],
             cmd: str,
             success_exit_status: typing.Optional[typing.Set[int]] = None,
     ):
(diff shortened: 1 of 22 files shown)


liangxin1300:
After using ssh-agent to setup a cluster, some commands will fail without adding `SSH_AUTH_SOCK`:
- crm cluster run "cmd"
- crm report
- crm cluster stop --all
Can crmsh be aware this cluster had been set up via ssh-agent? So that crmsh can give a warning to hint use how to do?

nicholasyang2022:
> Can crmsh be aware this cluster had been set up via ssh-agent? So that crmsh can give a warning to hint use how to do?

@liangxin1300 `ClusterShell` has the ability to catch an ssh call with 255 exit code and raise an `AuthorizationError` exception. It is easy to add diagnose messages for that exception.

However, the old `get_stdout_or_raise_error` method does not handle 255 specially. To avoid breaking things, I made `ClusterShell` to raise `AuthorizationError` only when requested explicitly.

I will try enabling that exception by default.

liangxin1300:
@nicholasyang2022 I've added some of my feedback, please take a look when you have time

Thanks!

liangxin1300:
> @nicholasyang2022 I've added some of my feedback, please take a look when you have time
> 
> Thanks!

Forgot to submit my review...

crmsh/utils.py
```diff
@@ -3200,13 +3191,22 @@ def save_remote(self, remote_hosts: typing.Iterable[str]):
...
     def get(self, host):
         return self._hosts_users[host]
 
     def add(self, user, host):
         self._hosts_users[host] = user
 
+    def set_no_generating_ssh_key(self, value: bool):
+        self._no_generating_ssh_key = value
+
+    def get_no_generating_ssh_key(self) -> bool:
```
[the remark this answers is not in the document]
nicholasyang2022: They are not the same. `HostUserConfig` serves to update ssh related configuration in batch. Values are buffered in memory until `save()` is called and `get_no_generating_ssh_key()` is the getter of  the in-memory state of a `HostUserConfig` instance.

While `UserOfHost` serves to provide proper users pair to create ssh connection and uses a different strategy for buffering/caching.

crmsh/utils.py
```diff
@@ -3168,9 +3151,14 @@ class HostUserConfig:
     """
     def __init__(self):
```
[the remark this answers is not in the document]
nicholasyang2022: As I commented previously, `HostUserConfig` is an abstraction to read/write configuration files, while `UserOfHost` is an abstraction to calculate ssh parameters. I think they do not belongs to the same module, even though both of them depend on configuration `core.hosts`.

crmsh/ui_corosync.py
```diff
@@ -62,7 +62,7 @@ def do_status(self, context, status_type="ring"):
         '''
         Quick cluster health status. Corosync status or QNetd status
         '''
-        if not ServiceManager(sh.LocalOnlyClusterShell(sh.LocalShell())).service_is_active("corosync.service"):
+        if not ServiceManager(sh.ClusterShellAdaptorForLocalShell(sh.LocalShell())).service_is_active("corosync.service"):
```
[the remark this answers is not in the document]
nicholasyang2022: `ServiceManager` is able to operate on both local and remote nodes. However, in these early stage of bootstrap, I'd like to make it explicit and static to indicate ClusterShell is not ready to use and these instance of ServiceManager can only operate on localhost, so that on some day in the future, when we head back to these code, we can tell these codes are unrelated to ssh, directly from the call and interface,  without inspecting the implementation of `ServiceManager`.

This is a manual dependency injection. Instead of making `ServiceManager` depends directly on `ClusterShell`, it depends on the interface of `ClusterShell`, and its caller provides it a `LocalShell` adapted to that interface.

test/features/ssh_agent.feature
```diff
@@ -0,0 +1,80 @@
...
+    Then    Expected "Environmental variable SSH_AUTH_SOCK does not exist." not in stderr
+
+  Scenario: Errors are reported when there are no keys in ssh-agent
+    Given   ssh-agent is started at "/root/ssh-auth-sock" on nodes ["hanode1", "hanode2", "hanode3"]
+    When    Try "SSH_AUTH_SOCK=/root/ssh-auth-sock crm cluster init --use-ssh-agent -y" on "hanode1"
+    Then    Expected "ssh-add" in stderr
+
+  Scenario: Skip creating ssh key pairs with --use-ssh-agent
+    Given   Run "mkdir /root/ssh_disabled && mv /root/.ssh/id_* /root/ssh_disabled" OK on "hanode1,hanode2,hanode3"
+    When    Run "SSH_AUTH_SOCK=/root/ssh-auth-sock ssh-add /root/ssh_disabled/id_rsa" on "hanode1,hanode2,hanode3"
```
liangxin1300, line 20: The `qnetd-node` seems not configured with ssh-agent?

liangxin1300:
```
# SSH_AUTH_SOCK=/root/ssh-auth-sock crm cluster init --use-ssh-agent -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
WARNING: No NTP service found.
INFO: Use public keys from ssh-agent:
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDf/Lj5aAT+8zJrukNGGqGi6seRbUwahL6UnAL9Z+D33ivueBpQt6CFzzz4mcRKXhovHMgA20GJxedV5rOcymeFHLf1KcNxPW5f+VKDnnI39R0++/KEyulxY9ZKDjEehMwYFlYZ1q6KlYa8VxwjSt+s3EdmlR2zKEi2gVHmV+UX+1jwPCsgAB5vFuu+hFO4tBc00xg7F5vSTu+RBJYck/VB9i1BMqnyZnJxe57sScsxrlvJQ26N9cjutaKJAwfo/qxtBCukjgK2jVhtFh3ZmtB+h4o74ig9ozGceONzzDhx5KCR0UqZDiZcI75aUrhhYRNGuc/oZdbanqEWx5Bwq8t9CxE2gN/GX2fSrYlR5DhEjoDFIbYv26kglyaOoHgv6EtAXcSssAvTi4MhsHfUDjlScQQyKsmMsVtDFYrbNAoI9nhdLII+wRW01IbKYygupHvSZfP8Quho0MIwDXUd3CwodHwI4iGfZMxawfsZEPPJlhmkmzOronwJ1FJhDNIcG+U= root@e72d101d1e96
```
Seems no need to show the key content?

liangxin1300:
```
INFO: Merging known_hosts
Warning: Permanently added 'hanode3,2001:db8:10::2' (ECDSA) to the list of known hosts.
```
And this warning is not necessary, either

liangxin1300:
`SSH_AUTH_SOCK=/root/ssh-auth-sock crm report` will hang
It's better to add crm report case under ssh-agent secnario

test/features/ssh_agent.feature
nicholasyang2022: Programs on qnetd-node do not create ssh session to other nodes. There is no need for ssh-agent.

crmsh/ssh_key.py
```diff
@@ -16,6 +18,32 @@ def __init__(self, msg: str):
...
+    def __init__(self, msg):
+        super().__init__(f'{msg}{self.diagnose()}')
+
+    @staticmethod
+    def diagnose() -> str:
+        with StringIO() as buf:
+            if 'SSH_AUTH_SOCK' not in os.environ:
+                buf.write(' Environmental variable SSH_AUTH_SOCK does not exists.')
+                if 'SUDO_USER' in os.environ:
+                    buf.write(' Please make sure environmental variables are preserved across sudo calls.')
```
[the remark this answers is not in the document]
nicholasyang2022: Yeah. It is covered when non-root scenarios are added to functional tests.

liangxin1300:
Compare with current crmsh-4.6 code:
```
# crm cluster init -y

```
Above 2 lines are missing while not using ssh-agent;
And
```
```
Above line is missing while using ssh-agent (but hacluster's key still be generated locally?)

liangxin1300:
> `SSH_AUTH_SOCK=/root/ssh-auth-sock crm report` will hang, please check
> 
> And, it's better to add crm report case under ssh-agent scenario

Works now, thanks!

zzhou1:
> Compare with current crmsh-4.6 code:
> 
> ```
> # crm cluster init -y
> 

...

> ```
> INFO: SSH key for hacluster does not exist, hence generate it now
> ```
> 
> Above line is missing while `using` ssh-agent (but hacluster's key still be generated locally? I mean don't we consider hacluster also using ssh-agent? @zzhou1 )

Agree, "hacluster" ssh-key should be inside the cluster node since hawk might call for it.

crmsh/sh.py
```diff
@@ -44,15 +46,26 @@ def __init__(self, msg, cmd):
...
         self.user = user
 
+    @staticmethod
+    def diagnose() -> str:
+        if user_of_host.instance().use_ssh_agent():
+            with StringIO() as buf:
+                if 'SSH_AUTH_SOCK' not in os.environ:
+                    buf.write('Environment variable SSH_AUTH_SOCK does not exist.')
+                    if 'SUDO_USER' in os.environ:
+                        buf.write(' Please make sure environment variables are preserved across sudo calls.')
```
zzhou1, line 66:
```suggestion
                        buf.write(' Probably could consider, eg. sudo --preserve-env=SSH_AUTH_SOCK .')
```

nicholasyang2022: Lacking `--preserve-env=SSH_AUTH_SOCK` is only one of the possible causes of `SSH_AUTH_SOCK` not found in environment variables. How about ` Please check whether an ssh-agent is available and consider using "sudo --preserve-env=SSH_AUTH_SOCK"`

liangxin1300: `Please check whether ssh-agent is available and consider using "sudo --preserve-env=SSH_AUTH_SOCK"`

liangxin1300 (approved):

PR #1262, opened 2023-10-25
Fix Improper Method Call (Close file descriptors to prevent leaks)

fazledyn-or:
## Details
While triaging your project, we found that your project had used `mkstemp()` method in several places to create temporary files. However, it came to our attention that the file descriptors that were generated by the method weren't being closed at all. This can lead to a [file descriptor leak](https://cwe.mitre.org/data/definitions/775.html) which should be prevented. It is suggested that file descriptors should be closed after use.

### Proof of Concept
For the file descriptor leak related proof, one can execute the following script:
```py
from tempfile import mkstemp

for i in range(10000):
    print(f"Opening file descriptor for {i}-th time")
    fd, name = mkstemp()
    print(f"FD: {fd}")
```
During execution, the program is going to face an error/exception, since in Linux, each processes are allowed to have 1024 file descriptors associated with them. Among them, file descriptor 0, 1, 2 are assigned to `stdin`, `stdout`, and `stderr` respectively.

## Changes
- As a prevention, I have used `os.close()` method to close the file descriptors after their use.

Suggestions related to these changes are welcomed.

## CLA Requirements
*This section is only relevant if your project requires contributors to sign a Contributor License Agreement (CLA) for external contributions.*

All contributed commits are already automatically signed off.

> The meaning of a signoff depends on the project, but it typically certifies that committer has the rights to submit this work under the same license and agrees to a Developer Certificate of Origin (see [https://developercertificate.org/](https://developercertificate.org/) for more information).
\- [Git Commit SignOff documentation](https://developercertificate.org/)

## Sponsorship and Support
This work is done by the security researchers from OpenRefactory and is supported by the [Open Source Security Foundation (OpenSSF)](https://openssf.org/): [Project Alpha-Omega](https://alpha-omega.dev/). Alpha-Omega is a project partnering with open source software project maintainers to systematically find new, as-yet-undiscovered vulnerabilities in open source code - and get them fixed – to improve global software supply chain security.

The bug is found by running the Intelligent Code Repair (iCR) tool by OpenRefactory and then manually triaging the results.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -679,6 +679,7 @@ def str2tmp(_str, suffix=".pcmk"):
     if not s.endswith('\n'):
         f.write("\n")
     f.close()
+    os.close(fd)
     return tmp
 
 
@@ -1235,6 +1236,7 @@ def run_ptest(graph_s, nograph, scores, utilization, actions, verbosity):
     if config.core.dotty and not nograph:
         fd, dotfile = mkstemp()
         ptest = "%s -D %s" % (ptest, dotfile)
+        os.close(fd)
     else:
         dotfile = None
     # ptest prints to stderr
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -679,7 +679,6 @@ def str2tmp(_str, suffix=".pcmk"):
     if not s.endswith('\n'):
         f.write("\n")
     f.close()
-    os.close(fd)
     return tmp
 
 
--- a/crmsh/cibstatus.py
+++ b/crmsh/cibstatus.py
@@ -237,6 +237,7 @@ class CibStatus(object):
         if config.core.dotty and not nograph:
             fd, dotfile = mkstemp()
             cmd = "%s -D %s" % (cmd, dotfile)
+            os.close(fd)
         else:
             dotfile = None
         rc = ext_cmd(cmd % self.source_file())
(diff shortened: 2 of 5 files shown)


crmsh/utils.py
```diff
@@ -679,6 +679,7 @@ def str2tmp(_str, suffix=".pcmk"):
     if not s.endswith('\n'):
         f.write("\n")
     f.close()
+    os.close(fd)
```
liangxin1300, line 682: CI failed after this line since that file needs to return and continue to be used

fazledyn-or: Removed it.

Even though I don't understand why it'd fail- since I'm just closing the FD and returning the file name.

fazledyn-or: Uh, my bad. Understood.
```
f = os.fdopen(fd, "w")
```
is being used. So, `f.close()` is enough. No need to close the FD manually.

fazledyn-or:
Hello @liangxin1300 , sorry for bothering you. Can please review again?
Thanks

liangxin1300:
> Hello @liangxin1300 , sorry for bothering you. Can please review again? Thanks

Your commits looks good, thanks!
But we need to wait since one of a codecov checking not passed, maybe need to adjust the check rules for those 0% diff hit  cases

nicholasyang2022:
Leaving these file descriptors open does not cause problems. As crmsh is a one-shot utility tool instead of a daemon, all file descriptors are closed when the process exits.

PR #1264, opened 2023-10-27
Dev: ui_corosync: Improve corosync.show and corosync.edit subcommands

liangxin1300:
- corosync.show: validate corosync.conf before showing it
- corosync.edit: edit a temporary copy of corosync.conf, validate it before saving it,
    then give the user hints on how to sync to other nodes
- Add corosync_ui.feature to collect test cases for crm corosync ui

### DIFF
--- a/crmsh/ui_corosync.py
+++ b/crmsh/ui_corosync.py
@@ -121,7 +121,11 @@ class Corosync(command.UI):
         '''
         cfg = corosync.conf()
         try:
-            utils.edit_file_ext(cfg, template='')
+            rc = utils.edit_file_ext(cfg, corosync.is_valid_corosync_conf)
+            if rc and len(utils.list_cluster_nodes()) > 1:
+                logger.warning(f"\"{cfg}\" has changed, should be synced with other nodes")
+                logger.info("Use \"crm corosync diff\" to show the difference")
+                logger.info("Use \"crm corosync push\" to sync")
         except IOError as e:
             context.fatal_error(str(e))
 
@@ -129,10 +133,9 @@ class Corosync(command.UI):
         '''
         Display the corosync configuration.
         '''
-        cfg = corosync.conf()
-        if not os.path.isfile(cfg):
-            context.fatal_error("No corosync configuration found on this node.")
-        utils.page_string(open(cfg).read())
+        if not corosync.is_valid_corosync_conf():
+            return False
+        utils.page_file(corosync.conf())
 
     def do_log(self, context):
         '''
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1230,31 +1230,27 @@ def edit_file(fname):
     return ext_cmd_nosudo("%s %s" % (config.core.editor, fname))
 
 
-def edit_file_ext(fname, template=''):
+def edit_file_ext(fname: str, validator: typing.Callable[[typing.IO], bool] = None) -> bool:
     '''
     Edit a file via a temporary file.
     Raises IOError on any error.
+
+    returns True if the file was changed
     '''
-    if not os.path.isfile(fname):
-        s = template
-    else:
-        s = open(fname).read()
-    filehash = hash(s)
-    tmpfile = str2tmp(s)
-    try:
-        try:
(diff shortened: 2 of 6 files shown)


liangxin1300:
- [ ] How to loosen the limitation of codecov to allow such interactive cases which have 0% diff hit?

crmsh/utils.py
```diff
@@ -1221,31 +1221,22 @@
...
-            f2 = open(fname, 'w')
-            f2.write(s)
-            f2.close()
-        finally:
-            os.unlink(tmpfile)
-    except OSError as e:
-        raise IOError(e)
+    with create_tempfile() as tmpfile:
+        source_data = read_from_file(fname)
+        str2file(source_data, tmpfile)
```
nicholasyang2022, line 1232: `str2file` uses `open_atomic`. It does not make sense to it for writing a tempfile. Something like `shutil.copyfile` is enough.

crmsh/utils.py
```diff
@@ -1221,31 +1221,22 @@ def edit_file(fname):
     return ext_cmd_nosudo("%s %s" % (config.core.editor, fname))
 
 
-def edit_file_ext(fname, template=''):
+@contextmanager
+def edit_file_ext(fname):
```
nicholasyang2022, line 1225: The interface looks strange. `edit_file_ext` does not finished the job of editing a file. How about `def edit_file_ext(fname: str, validator: typing.Callable[typing.IO, bool])`

crmsh/ui_corosync.py
```diff
@@ -121,18 +121,29 @@ def do_edit(self, context):
...
         '''
         Display the corosync configuration.
         '''
+        if not corosync.is_valid_corosync_conf():
+            return False
         cfg = corosync.conf()
-        if not os.path.isfile(cfg):
-            context.fatal_error("No corosync configuration found on this node.")
-        utils.page_string(open(cfg).read())
+        utils.page_string(utils.read_from_file(cfg))
```
nicholasyang2022, line 146: No need to read it into memory and send to pager. `page_file` is enough.

crmsh/utils.py
```diff
@@ -1221,31 +1221,26 @@ def edit_file(fname):
...
+    with create_tempfile() as tmpfile:
+        shutil.copyfile(fname, tmpfile)
+        if edit_file(tmpfile) != 0:
+            raise IOError(f"Cannot edit file \"{fname}\"")
+        changed_data = read_from_file(tmpfile)
+        source_data = read_from_file(fname)
+        if hash(changed_data) != hash(source_data):
+            if validator and not validator(tmpfile):
+                return False
+            shutil.copyfile(tmpfile, fname)
```
nicholasyang2022, line 1240: The original file needs to be replaced atomically.  Either create the temp file in the same directory, fsync and use `os.rename` to replace the original file, or use `str2file`.

liangxin1300: Changed, please take a look
Thanks!

crmsh/utils.py
```diff
@@ -1230,31 +1230,27 @@ def edit_file(fname):
...
-            os.unlink(tmpfile)
-    except OSError as e:
-        raise IOError(e)
+    with create_tempfile() as tmpfile:
+        shutil.copyfile(fname, tmpfile)
+        if edit_file(tmpfile) != 0:
+            raise IOError(f"Cannot edit file \"{fname}\"")
+        changed_data = read_from_file(tmpfile)
+        source_data = read_from_file(fname)
+        if hash(changed_data) != hash(source_data):
```
nicholasyang2022, line 1246: `hash(a) != hash(b)` is not a necessary condition for `a != b`.

nicholasyang2022 (changes requested):
1. 2 pieces of data shares a same hash value does not means those data are the same.
2. Please update `after_n_builds` in `codecov.yml` as a new test job is added.

crmsh/utils.py
liangxin1300: Changed as comparing them directly

nicholasyang2022 (approved):

PR #1265, opened 2023-11-01
Dev: ui_utils: Supports '=' when setting node/resource attributes

liangxin1300:
Now supports both:

    crm node attribute <node> set <attr> <value>
    crm node attribute <node> set <attr>=<value>
    
    crm resource meta <rsc> set <attr> <value>
    crm resource meta <rsc> set <attr>=<value>
    
    crm resource param <rsc> set <param> <value>
    crm resource param <rsc> set <param>=<value>

### DIFF
--- a/test/features/configure_bugs.feature
+++ b/test/features/configure_bugs.feature
@@ -15,8 +15,11 @@ Feature: Functional test for configure sub level
     And     Online nodes are "hanode1 hanode2"
 
     # mask password by default
-    When    Run "crm node utilization hanode1 set password qwertyui" on "hanode1"
+    When    Run "crm node utilization hanode1 set password=qwertyui" on "hanode1"
     When    Try "crm configure show|grep password|grep qwertyui"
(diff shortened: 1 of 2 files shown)


test/features/configure_bugs.feature
```diff
@@ -15,7 +15,7 @@ Feature: Functional test for configure sub level
     And     Online nodes are "hanode1 hanode2"
 
     # mask password by default
-    When    Run "crm node utilization hanode1 set password qwertyui" on "hanode1"
+    When    Run "crm node utilization hanode1 set password=qwertyui" on "hanode1"
```
zzhou1, line 18: Instead of change the current test case, which is our preference conforms to the help text, I prefer to add one more test case. In the end, both are tested.

liangxin1300: Changed

zzhou1 (approved):

PR #1271, opened 2023-11-15
[crmsh-4.4] Fix: log: Redirect debug messages into stderr (bsc#1208991)

liangxin1300:
If keep debug and info messages all in stdout, then some scripts that need to handle or analyze the output of the crm command might be messy.

Also redirect warning messages into stderr

backport from #1127

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -19,12 +19,12 @@ class ConsoleCustomHandler(logging.StreamHandler):
     """
     A custom handler for console
 
-    Redirect ERROR message to sys.stderr
-    Redirect INFO/WARNING/DEBUG message to sys.stdout
+    Redirect ERROR/WARNING/DEBUG message to sys.stderr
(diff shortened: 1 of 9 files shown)


nicholasyang2022:
How does [Dev: behave: Add 'sudo' before most of commands](https://github.com/ClusterLabs/crmsh/pull/1271/commits/7ebd01e4cbf6b3cf98d75a0830d30252333a9a4f) have anything to do with this bug?

liangxin1300:
> How does [Dev: behave: Add 'sudo' before most of commands](https://github.com/ClusterLabs/crmsh/pull/1271/commits/7ebd01e4cbf6b3cf98d75a0830d30252333a9a4f) have anything to do with this bug?

No, just see the conflicts when backporting, and cherry-pick the related part, should no harm

nicholasyang2022 (approved):

PR #1280, opened 2023-12-01
[crmsh-4.6] Unify version definition

nicholasyang2022:
There are 2 version strings used in crmsh:

1. The package metadata defined in `setup.py`. It was hardcoded in `setup.py`.
2. The version file in source root `./version`, which will be installed to `/usr/shar/crmsh/version`, used by `crm --version`. It was hardcoded in `AC_INIT` macro in `configure.ac`.

This pull request contains 2 changes:

1. Make `setup.py` read version from `./version`, so that the version string is unified between python package metadata and `crm --version`.
2. Fix the `--with-version` option in autoconf, so that it can override the version string in generated `./version` file. This allow a packager to override the version string.

With these change, a packager can use `git describe` and `./configure --with-version` to generate version string automatically from git tags.


liangxin1300 (approved):
LGTM

Thanks!

PR #1283, opened 2023-12-08
[crmsh-4.6] Dev: log: log the backtrace of ValueError in log file and refactoring

nicholasyang2022:
In previous versions, crmsh does not print backtrace for most  of exceptions to make the error message more readable. However, backtrace is important information to debug a problem happened in a production environment.

This pull request enable crmsh to print the backtrace in logfile and suppress it in console in the same time, by implement a custom LogFormatter for console.

And this pull request also contains these refactoring and improvements:

* Avoid modifying module `logging` in the standard library. Instead, subclass the classes in it to customize.
* Avoid modifying `LogRecord` in handlers and formatters as it is shared by multiple handlers and formatters.
* Use `levelno` instead `levelname` to filter the logs for better performance.
* Remove `ConsoleReportFormatter` as it does nothing different to `ConsoleCustomFormatter`.
* Remove `FileCustomFormatter` as it does nothing different to `logging.Formatter`.
* Move the implementation of `lineno` used in regression tests, from handler to logger, as the number is controlled manually by the caller of logger.

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -12,8 +12,17 @@ from contextlib import contextmanager
 from . import options
 from . import constants
 
-
+DEBUG2 = logging.DEBUG + 5
 CRMSH_LOG_FILE = "/var/log/crmsh/crmsh.log"
+
+
+class DEBUG2Logger(logging.Logger):
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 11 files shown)


liangxin1300 (approved):
Nice work!!

Thanks!

PR #1289, opened 2023-12-21
[crmsh-4.6] Dev: bootstrap: Assign hosts with _context.node_list_in_cluster in jo…

liangxin1300:
…in_ssh_merge

Instead of invoking `utils.list_cluster_nodes`, consider the following:
If the cluster service on the current node is not running, it will read
from `cib.xml`. If `cib.xml` contains an original node that is no longer
available, the join process at ssh_merge stage will fail;
    
`_context.node_list_in_cluster` is a list of node names that are fetched
from the init node, queried by `crm_node -l`

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1914,10 +1914,7 @@ def join_ssh_merge(cluster_node, remote_user):
     """
     logger.info("Merging known_hosts")
 
-    hosts = utils.list_cluster_nodes()
-    if hosts is None:
-        hosts = list()
-    hosts.append(cluster_node)
+    hosts = _context.node_list_in_cluster
 
     shell = sh.cluster_shell()
     # create local entry in known_hosts


nicholasyang2022:
What is the difference between parsing nodes from `cibadmin -Q` and `crm_node -l`? In what cases will they return different results?

liangxin1300:
> What is the difference between parsing nodes from `cibadmin -Q` and `crm_node -l`? In what cases will they return different results?

For the running cluster, the results should be the same;

For this case, node1 is the init node, the service is running;
node2 is the join node, the service is not running yet (at the time join_ssh_merge);
At this point, if node2 is calling `utils.list_cluster_nodes`, it will parse the file cib.xml, which might contains `unrelated` node (like node3 from origin cluster); if node2 using `_context.node_list_in_cluster`, it's the remote result of `crm_node -l`, from init node node1, that's the correct node list for the current cluster.

nicholasyang2022 (approved):

Issue #826, opened 2021-06-11
[Question] Usage for altering operation parameters using crm config set


asidique:
Hello,

I'm trying to understand the behaviour for `crm config set` when using it to change a resource's operation parameters. For an example, I have the following resource configuration: 
<pre><code>primitive my_resource my_host \
        params instance=instance1 hostname=host_name \
        op monitor timeout=50s <b>interval=10s</b> on-fail=ignore \
        op start interval=0s timeout=900s \
        op stop interval=0s timeout=900s \
        meta migration-threshold=0 multiple-active=block is-managed=true
</pre></code>

Then, I run `crm config set my_resource.monitor.10s.interval 50s` and it successfully changes the configuration:
<pre><code>primitive my_resource my_host \
        params instance=instance1 hostname=host_name \
        op monitor timeout=50s <b>interval=50s</b> on-fail=ignore \
        op start interval=0s timeout=900s \
        op stop interval=0s timeout=900s \
        meta migration-threshold=0 multiple-active=block is-managed=true
</pre></code>
However, when I want to change the monitor operation again, I'm expecting to use the following command:
`crm config set my_resource.monitor.50s.interval 100s`, however; this results in the following error:
`ERROR: configure.set: Operation "monitor" interval "50s" not found for resource my_resource`. 

Specifying the interval 10s again does work:
`crm config set my_resource.monitor.10s.interval 100s`

Lastly, I know the `crm config set` command doesn't require an interval but I'm interested in understanding the behaviour of the command when using the interval. I tried to also run `crm config commit` and `crm resource refresh` to ensure the configuration is updated but the behaviour is the same.

liangxin1300:
Hi @asidique ,

When first time configured, the `id` for your monitor operation should be `my_resource-monitor-10s`,
After change the interval value as 50s, I think the `id` should update as `my_resource-monitor-50s`, but it not updated
I think this should be a bug.

Thanks for report!

PR #1292, opened 2024-01-02
[crmsh-4.6] Update the operation id after the monitor interval changed

liangxin1300:
The monitor operation's ID is constructed as
'<resource_id>-monitor-<interval_value>'. Therefore, when the monitor
interval is altered by configure.set, the ID should also be updated to
reflect the latest interval. If not, configure.set cannot update the
value based on the current path, which is identified by
id.[op_type.][interval.]name.
    
See the issue https://github.com/ClusterLabs/crmsh/issues/826

### DIFF
--- a/crmsh/parse.py
+++ b/crmsh/parse.py
@@ -12,7 +12,7 @@ from .ra import disambiguate_ra_type, ra_type_validate
 from . import schema
 from .utils import keyword_cmp, verify_boolean, lines2cli
 from .utils import get_boolean, olist, canonical_boolean
-from .utils import handle_role_for_ocf_1_1, compatible_role
+from .utils import handle_role_for_ocf_1_1, compatible_role, add_time_unit_if_needed
 from . import xmlutil
 from . import log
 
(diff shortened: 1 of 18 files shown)


nicholasyang2022:
This patch looks good. However, when I try to use it against `admin-ip`, I got

```
suse@ha-1-1:~> sudo crm configure show admin-ip
primitive admin-ip IPaddr2 \
	params ip=192.168.122.17 \
	op monitor interval=10 timeout=20
suse@ha-1-1:~> sudo crm configure set admin-ip.monitor.10s.interval 60s
ERROR: configure.set: Operation "monitor" interval "10s" not found for resource admin-ip
suse@ha-1-1:~> sudo crm configure set admin-ip.monitor.10.interval 60
WARNING: (unpack_config) 	warning: Blind faith: not fencing unseen nodes
```

This monitor id for admin ip generated in cluster initialization is inconsistent with others. Is this a bug?

liangxin1300:
> This patch looks good. However, when I try to use it against `admin-ip`, I got
> 
> ```
> suse@ha-1-1:~> sudo crm configure show admin-ip
> primitive admin-ip IPaddr2 \
> 	params ip=192.168.122.17 \
> 	op monitor interval=10 timeout=20
> suse@ha-1-1:~> sudo crm configure set admin-ip.monitor.10s.interval 60s
> ERROR: configure.set: Operation "monitor" interval "10s" not found for resource admin-ip
> suse@ha-1-1:~> sudo crm configure set admin-ip.monitor.10.interval 60
> WARNING: (unpack_config) 	warning: Blind faith: not fencing unseen nodes
> ```
> 
> This monitor id for admin ip generated in cluster initialization is inconsistent with others. Is this a bug?

I raised an issue https://github.com/ClusterLabs/crmsh/issues/1304
And if the existing CIB already contains interval without 's' unit, the `crm configure set` should consider both 10 and 10s cases

liangxin1300:
Fix #1304 in https://github.com/ClusterLabs/crmsh/pull/1292/commits/8bf489c2051e6218ef3a1574e78835fac9e51ae1

PR #1295, opened 2024-01-03
[crmsh-4.6] Fix: report: Unable to gather log files that are in the syslog format (bsc#1218491)

liangxin1300:
Logs in syslog format lack year information, making it challenging to determine the year a line was recorded. Therefore, when encountering a future timestamp, we must estimate the year

### DIFF
--- a/crmsh/report/utils.py
+++ b/crmsh/report/utils.py
@@ -260,12 +260,12 @@ def find_files_in_timespan(context: core.Context, target_dir_list: List[str]) ->
     return file_list
 
 
-def find_first_timestamp(data: List[str]) -> float:
+def find_first_timestamp(data: List[str], log_file: str) -> float:
     """
     Find the first timestamp in the given list of log line
     """
     for line in data:
-        timestamp = get_timestamp(line)
+        timestamp = get_timestamp(line, log_file)
         if timestamp:
             return timestamp
     return None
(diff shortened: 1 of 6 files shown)


crmsh/report/utils.py
```diff
@@ -362,7 +362,29 @@ def get_pkg_mgr() -> str:
     return ""
 
 
-def get_timestamp(line: str) -> float:
+def get_timestamp_from_time_line(time_line: str, stamp_type: str, log_file: str):
+    timestamp = crmutils.parse_to_timestamp(time_line, quiet=True)
+    # got a timestamp in the future
+    if timestamp and timestamp > datetime.datetime.now().timestamp() and stamp_type == "syslog":
```
nicholasyang2022, line 368: It is better for performance to check `stamp_type` before calling `datetime.now()`.

crmsh/report/utils.py
```diff
@@ -362,7 +362,29 @@ def get_pkg_mgr() -> str:
...
+        # syslog doesn't have year info, so we need to guess it
+        mtime = os.path.getmtime(log_file)
+        mtime = datetime.datetime.fromtimestamp(mtime)
+        this_year = datetime.datetime.now().year
```
nicholasyang2022, line 372: Instead of calling `datatime.now()` repeatly, the datetime instance can be cached as a local variable.

nicholasyang2022 (approved):

PR #1297, opened 2024-01-05
Dev: coverage: inject coverage with sitecustomize.py in feature tests

nicholasyang2022:
We used to inject the coverage in the program entrance /usr/sbin/crm. However, crm.report uses multiprocessing and the child processes does not execute code in /usr/sbin/crm. So its coverage cannot be caculated.

This patch use an alternative method, by using sitecustomize.py, to inject the coverage. It should work for all python processes.


liangxin1300:
report module already be ignored in 5f1fb4db
Maybe at higher python3 version we have better solution?

PR #1312, opened 2024-01-22
[crmsh-4.6] Fix: ui_cluster: Can't start cluster with --all option if no cib (bsc#1219052)

liangxin1300:
## Issue
When setting up cluster via yast2-cluster,
- Can't start the cluster with --all option:
```
15sp5-1:~ # crm cluster start --all
ERROR: cluster.start: Cannot get the node list from cluster
```
- Can't start cluster on peer node
```
15sp5-1:~ # crm cluster start 15sp5-2
ERROR: cluster.start: Cannot get the node list from cluster
```
- Can't start cluster on local node
```
15sp5-1:~ # crm cluster start 15sp5-1
ERROR: cluster.start: Cannot get the node list from cluster
```
## Solution
- The --all option try to query the node list from `utils.list_cluster_nodes`, which should grab the address list from corosync.conf when there is no cib
- For the peer node case, stop compare with the return results from list_cluster_nodes, since it's hard to tell if the given name is not the alias name
- For the local node case, return itself immediately

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1766,6 +1766,14 @@ def print_cluster_nodes():
         print("{}\n".format(out))
 
 
+def get_address_list_from_corosync_conf():
+    """
+    Return a list of addresses configured in corosync.conf
+    """
+    from . import corosync
+    return corosync.get_values("nodelist.node.ring0_addr")
+
+
 def list_cluster_nodes(no_reg=False):
     '''
     Returns a list of nodes in the cluster.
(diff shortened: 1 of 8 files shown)


crmsh/utils.py
```diff
@@ -1796,9 +1796,9 @@ def list_cluster_nodes(no_reg=False):
     node_list = []
     for node in cib.xpath(constants.XML_NODE_PATH):
         name = node.get('uname') or node.get('id')
+        # exclude remote node
         if node.get('type') == 'remote':
-            srv = cib.xpath("//primitive[@id='%s']/instance_attributes/nvpair[@name='server']" % (name))
-            if srv:
+            if cib.xpath(f"//primitive/instance_attributes/nvpair[@name='server' and @value='{name}']"):
```
nicholasyang2022, line 1801: The attributes of `primitive` node should be check to ensure it is a `ocf::pacemaker:remote`.

nicholasyang2022 (changes requested):

nicholasyang2022 (approved):

PR #1321, opened 2024-02-05
[crmsh-4.6] Fix: bootstrap: clear stall data about ssh users left possiblely from previous setups (bsc#1219476)

nicholasyang2022:
/root/.config/crm/crm.conf may contain `core.hosts` and `core.no_generating_ssh_key` from previous cluster setup. They should be clear on `crm cluster init` and `crm cluster join`.


nicholasyang2022:
The regression test for [bsc#1183654](https://bugzilla.suse.com/show_bug.cgi?id=1183654) fails. There are 2 problems:

1. `/etc/hosts` are not consistent across the cluster. The test adds the alias `HANODE1` in the `/etc/hosts` only on hanode1, so it is hard for the code running on hanode2 to figure out `HANODE1` is an alias of `hanode1`.
2. libc wrapper provided in python standard library, particularly `gethostbyaddr` and `getnameinfo`, does not provide all features available in libc. `gethostbyaddr` does not accept address family, so it cannot return all aliases in IPv4/IPv6 dual stack environment. And `getnameinfo` simply is not capable to return multiple results.

These problems prevents us from handling hostname aliases properly. We have to stick to use  canonical hostname.

liangxin1300 (approved):

Issue #1324, opened 2024-02-13
Incompatible with Python 3.12


Hyask:
Hi,

It seems `crmsh` is using `distutils` [at least here](https://github.com/ClusterLabs/crmsh/blob/master/crmsh/utils.py#L37), and `imp` [here](https://github.com/ClusterLabs/crmsh/blob/master/test/unittests/test_utils.py#L27).
Unfortunately, `distutils` has been [deprecated](https://docs.python.org/3.12/whatsnew/3.10.html#distutils-deprecated) in Python 3.10 and removed completely in Python 3.12. `imp` has also been [removed](https://docs.python.org/3.12/whatsnew/3.12.html) in Python 3.12.

Please see [this Debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1061794) and [this Ubuntu bug](https://bugs.launchpad.net/ubuntu/+source/booth/+bug/2052018) for example of issues with the removal of `distutils`.

PR #1325, opened 2024-02-14
Python 3.12 compatibility

Hyask:
See #1324 for description of the problem.

I don't know what solution you would prefer regarding the `LooseVersion` problem:
* there is this current implement that uses a kinda hidden code in `setuptools`
  *pros*: `setuptools`  is widely distributed everywhere, so that code is already broadly available
  *cons*: as it's a kinda of "hidden" part of the API, I don't know how this will be maintained in the future
* another solution is to use the specific [looseversion](https://github.com/effigies/looseversion) package
  *pros*: that package's only purpose is to provide the `LooseVersion` API, so it won't change even in the distant future
  *cons*: that package is recent and not broadly available yet. It's available in Fedora, but not in Debian, for example, making the packaging of new version of `crmsh` in distributions more difficult.

### DIFF
--- a/crmsh/ra.py
+++ b/crmsh/ra.py
@@ -49,7 +49,7 @@ def crm_resource(opts):
 
 @utils.memoize
 def can_use_lrmadmin():
-    from distutils import version
+    from setuptools._distutils import version
     # after this glue release all users can get meta-data and
     # similar from lrmd
     minimum_glue = "1.0.10"
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -34,7 +34,7 @@ from . import userdir
 from . import constants
 from . import options
 from . import term
-from distutils.version import LooseVersion
+from setuptools._distutils.version import LooseVersion
 from .constants import SSH_OPTION
 from . import log
 from .prun import prun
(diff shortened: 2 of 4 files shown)


Hyask:
It would be kind of you to trigger the CI on this branch, as I've only been testing this with some particular Debian and Ubuntu autopkgtest setup.

crmsh/ra.py
```diff
@@ -49,7 +49,7 @@ def crm_resource(opts):
 
 @utils.memoize
 def can_use_lrmadmin():
```
liangxin1300, line 51: I doubt the codes related to lrmadmin already deprecated
Let me confirm that
If yes, then we can skip this issue

liangxin1300: Another solution https://github.com/readthedocs/readthedocs-sphinx-ext/pull/111/files and https://github.com/readthedocs/readthedocs-sphinx-ext/issues/106

liangxin1300:
Hi @Hyask 

I created another PR in #1327 , please take a look

Thanks!

Hyask:
> Hi @Hyask
> 
> I created another PR in #1327 , please take a look
> 
> Thanks!

That other PR is great, I'll close this one. Many thanks for handling that quickly, it's good to see a project well maintained :-)

PR #1327, opened 2024-02-18
Python 3.12 compatibility

liangxin1300:
To address #1324 , changes include:
- Remove dead codes about lrmadmin, which was using distutils.version
- Enable Python 3.12 in CI
- Replace LooseVersion with packaging.version, and add 'packaging' dependency
- Use 'importlib' instead of 'imp' for Python 3.12, cherry pick from #1325 
- Dockerfile and test script adjustment due to docker image upgrade

Thanks @Hyask for helping!


Hyask:
This looks very good! There are some bits that I've no context on, but the 3.12 compat parts are great.

crmsh/utils.py
```diff
@@ -26,6 +26,7 @@
 from contextlib import contextmanager, closing
 from stat import S_ISBLK
 from lxml import etree
+from packaging import version
```
nicholasyang2022, line 29: Pacemaker is not a python package. Python packaging tools should not be used to parse it.

test/unittests/test_utils.py
```diff
@@ -24,7 +24,7 @@ def setup_function():
     utils._ip_for_cloud = None
     # Mock memoize method and reload the module under test later with imp
     mock.patch('crmsh.utils.memoize', lambda x: x).start()
-    imp.reload(utils)
+    importlib.reload(utils)
```
nicholasyang2022, line 27: It is better to use `functools.wraps` and patch `crmsh.utils.memoize.__wrapped__`, to avoid relying on reloading the module.

test_container/Dockerfile
```diff
@@ -3,11 +3,11 @@ MAINTAINER Xin Liang <XLiang@suse.com>
...
-RUN zypper refresh && \
-    zypper -n install systemd \
-        make autoconf automake vim which libxslt-tools mailx iproute2 iputils bzip2 openssh tar file glibc-locale-base firewalld libopenssl1_1 dos2unix iptables cpio \
-        python3 python3-pip python3-lxml python3-python-dateutil python3-setuptools python3-PyYAML python3-curses python3-behave \
-        csync2 libglue-devel corosync corosync-qdevice pacemaker booth corosync-qnetd
+RUN zypper refresh
+RUN zypper -n install systemd
+RUN zypper -n install make autoconf automake vim which libxslt-tools mailx iproute2 iputils bzip2 openssh tar file glibc-locale-base firewalld libopenssl1_1 dos2unix iptables cpio
+RUN zypper -n install python3 python3-pip python3-lxml python3-python-dateutil python3-setuptools python3-PyYAML python3-curses python3-behave
+RUN zypper -n install csync2 libglue-devel corosync corosync-qdevice pacemaker booth corosync-qnetd
```
nicholasyang2022, line 10: Splitting it into multiple steps result in more layers in container images and slows down the container. And running refresh before installing package help to ensure all the packages is latest.

test/run-functional-tests
```diff
@@ -16,6 +16,7 @@ HA_NETWORK_V6_ARRAY[0]="2001:db8:10::/64"
 HA_NETWORK_V6_ARRAY[1]="2001:db8:20::/64"
 BEHAVE_CASE_DIR="$(dirname $0)/features/"
 BEHAVE_CASE_EXCLUDE="sbd|ocfs2"
+SUDOERS="/usr/etc/sudoers"
```
nicholasyang2022, line 19: Instead of modifying `/usr/etc/sudoers`, we should add configurations in `/etc/sudoers`.

From https://en.opensuse.org/openSUSE:Packaging_UsrEtc:

> All distribution-provided configuration files are located below /usr/etc and are not modifiable by the admin (with a read-only root filesystem, this would not even be possible)

crmsh/utils.py
liangxin1300: This is recommended by PEP632 https://peps.python.org/pep-0632/#migration-advice

test_container/Dockerfile
liangxin1300: Changed

test/run-functional-tests
liangxin1300: Changed

crmsh/utils.py
nicholasyang2022: Yeah, it is recommended to replace distutils. However, both of `distutils` and `packaging` and designed for python packages, not an unrelated project written in C.

test/unittests/test_utils.py
liangxin1300: Changed

crmsh/utils.py
liangxin1300: I don't get your point here.
We are trying to compare the `string` like "pacemaker-3.7" and "pacemaker-3.6"

nicholasyang2022: Python's [Version Specifier Specification](https://packaging.python.org/en/latest/specifications/version-specifiers/) is quite complex. `packaging` is designed to handle it. It is not the expected use case of `packaging` to parse some version string from random, non-python projects. Using it to parse the version string of pacemaker causes confusions by implying pacemaker will follow python's specification.

liangxin1300: Then, what's your suggestion? any good alternative?

liangxin1300: I think `utils.is_larger_than_min_version` is not only for parsing Pacemake's version. It's a common function

nicholasyang2022 (approved):

PR #1335, opened 2024-02-22
[crmsh-4.6] Fix: report: use ClusterShell for ssh (bsc#1220170)

nicholasyang2022:
The current ssh implementation in crm.report will use a wrong user when using root + ssh-agent + sudo.

This patch change to use try to ClusterShell first, providing consistent behavior with other parts of crmsh. And fallback to interactive authentication when ClusterShell does not work.

### DIFF
--- a/crmsh/report/core.py
+++ b/crmsh/report/core.py
@@ -5,16 +5,19 @@
 import argparse
 import multiprocessing
 import os
-import re
+import subprocess
 import sys
 import shutil
 import json
 import ast
+import typing
 from inspect import getmembers, isfunction
 from typing import List
 
+import crmsh.sh
+import crmsh.report.sh
 from crmsh import utils as crmutils
-from crmsh import config, log, userdir, corosync, tmpfiles, ui_cluster, sh
+from crmsh import config, log, tmpfiles, ui_cluster
 from crmsh.sh import ShellUtils
 
 
@@ -39,7 +42,7 @@ class Context:
         self.single: bool= config.report.single_node
         self.sensitive_regex_list: List[str] = []
         self.regex_list: List[str] = "CRIT: ERROR: error: warning: crit:".split()
-        self.ssh_askpw_node_list: List[str] = []
+        self.passwordless_shell_for_nodes = dict()
         self.me = crmutils.this_node()
         self.pe_dir: str
         self.cib_dir: str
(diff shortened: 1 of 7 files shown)


crmsh/report/core.py
```diff
@@ -441,6 +432,31 @@ def load_context_attributes(context: Context) -> None:
     context.cores_dir_list.extend([constants.COROSYNC_LIB] if os.path.isdir(constants.COROSYNC_LIB) else [])
 
 
+def load_context_trace_dir_list(context: Context) -> None:
+    log_contents = ""
```
liangxin1300, line 436: Better to move these two comment lines (https://github.com/ClusterLabs/crmsh/blob/crmsh-4.6/crmsh/report/collect.py#L192-L193) into here to explain why need to do like this

crmsh/report/core.py
```diff
@@ -441,6 +432,31 @@ def load_context_attributes(context: Context) -> None:
...
+            logger.info('Creating ssh connection to %s...', node)
+            cmd = cmd.replace('"', '\\"')
+            cmd = 'ssh {} {} "{}{}"'.format(
+                constants.SSH_OPTS,
+                node,
+                '' if context.ssh_user is None or context.ssh_user == 'root' else 'sudo ',
+                cmd,
+            )
+            shell = crmsh.sh.LocalShell()
+            ret = shell.su_subprocess_run(None, cmd, tty=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
liangxin1300, line 454: Is the same code as line 264-275?

nicholasyang2022: Extracted as a function `_interactive_ssh_run_as_root`.

zzhou1 (approved):
LGTM

crmsh/report/core.py
```diff
@@ -441,6 +422,38 @@ def load_context_attributes(context: Context) -> None:
...
+            # This is a last effort when passwordless ssh is not available
+            cmd = cmd.replace('"', '\\"')
+            ret = _interactive_ssh_run_as_root(node, context.ssh_user, cmd)
+        if ret.returncode == 0:
+            log_contents += ret.stdout.decode('utf-8')
+    context.trace_dir_list = list(set(re.findall("written to (.*)/.*", log_contents)))
+
+
+def _interactive_ssh_run_as_root(host: str, ssh_user: str, cmd: str):
+    # cmd MUST be escaped to make sure it works as an arugment of ssh command
```
liangxin1300, line 444: What does this comment mean for "escaped", in this function scope?

nicholasyang2022: The cmd is concatenated to `ssh` as string. This string is firstly parsed by bash and later by ssh. This requires special characters in it to be escaped 2 times. And I am not sure how to write a general function to escape arguments for ssh. So for now, let the callers to escape it case by case.

PR #1339, opened 2024-02-29
Dev: unittest: disable python 3.10 in test matrix

nicholasyang2022:
As ALP is now targeting 3.11.


liangxin1300:
The tox.ini needs to change, too

liangxin1300 (approved):

PR #1342, opened 2024-03-01
Dev: workflows: Remove the redundant stage condition

liangxin1300:
Since the condition to check the branch is already in the 'integration' stage (https://github.com/ClusterLabs/crmsh/blob/4a0937c33154b631a341d16c69880d1ceedd8731/.github/workflows/crmsh-cd.yml#L19), we don't need to check it again in the subsequent stages.

### DIFF
--- a/.github/workflows/crmsh-cd.yml
+++ b/.github/workflows/crmsh-cd.yml
@@ -20,7 +20,6 @@ jobs:
     uses: ./.github/workflows/crmsh-ci.yml
 
   delivery:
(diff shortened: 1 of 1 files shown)


liangxin1300:
And the redundant `if` statement in the `submit` stage also break the yaml syntax https://github.com/ClusterLabs/crmsh/actions/runs/8108869283/workflow

nicholasyang2022 (approved):

Issue #1337, opened 2024-02-28
crm cluster init example on the start-guide page


mpagot:
https://github.com/ClusterLabs/crmsh/blob/e3d1170e72bc399c92ee8ac7f639a1e2058a7a0f/doc/website-v1/start-guide.adoc?plain=1#L120

The example reported is like

```
crm cluster init --name demo-cluster --nodes "alice bob"
```

Is it a valid syntax in crm --version 4.5.0 ? Should not be as reported by `crm cluster help init`

```
crm cluster init --name demo-cluster -N alice -N bob
```

liangxin1300:
Hi @mpagot 

Please use `--node "alice bob"`

And the start-guide might be outdated, should be updated


Thanks!

PR #1343, opened 2024-03-01
Fix cluster init example on the start guide.

mpagot:
Fix for https://github.com/ClusterLabs/crmsh/issues/1337


liangxin1300 (approved):
Thanks!

PR #1351, opened 2024-03-06
Minor refactoring for code duplication

liangxin1300:
- Dev: bootstrap: Refactor qdevice user parsing and finding
- Dev: ssh_key: Extract duplicate code about list keys from ssh-agent to a function


crmsh/bootstrap.py
```diff
@@ -108,6 +108,7 @@ def __init__(self):
         self.ipv6 = None
         self.qdevice_inst = None
         self.qnetd_addr = None
+        self.qnetd_addr_orig = None
```
nicholasyang2022, line 111: We need a better variable name. It is hard to tell what is a _original addr_.

liangxin1300:
Additional commit:  Dev: Remove the unused codes from crmsh/bootstrap.py and crmsh/utils.py

nicholasyang2022:
Does removing `vgfs` requires a PED?

liangxin1300:
> Does removing `vgfs` requires a PED?

I don't think so
Seems no user case
@zzhou1 Please take a look

liangxin1300:
One more commit to keep code clean:
```
 Dev: bootstrap: Remove unused ssh_remote related code
    
    Oringinal init_ssh_remote is used for adding init node's key to authorized_keys,
    currently this has been done by ssh_key.AuthorizedKeyManager, at init_ssh function
```

zzhou1:
> > Does removing `vgfs` requires a PED?
> 
> I don't think so Seems no user case @zzhou1 Please take a look

It was deprecated in the past. It's time to drop it. I do agree. Thx!

"""
logger.warning("vgfs stage was deprecated and is an alias of ocfs2 stage now")
"""

zzhou1:
> Does removing `vgfs` requires a PED?

No.

nicholasyang2022 (approved):

PR #1352, opened 2024-03-07
[crmsh-4.6] Fix: utils: A workaround for the truncating issue (bsc#1205925)

liangxin1300:
When python program calling cibsecret with a small terminal width, the command `ps -ef | grep '[p]acemaker-controld'` will return 1

>>> cmd = "ps -ef | grep '[p]acemaker-controld' >/dev/null"
>>> # When terminal width is small
>>> subprocess.call(cmd, shell=True)
1
>>> # When terminal is big enough
>>> subprocess.call(cmd, shell=True)
0

Set a large terminal width could be a workaround

See also: https://github.com/ClusterLabs/pacemaker/pull/3384

### DIFF
--- a/crmsh/constants.py
+++ b/crmsh/constants.py
@@ -499,6 +499,7 @@ SCHEMA_MIN_VER_SUPPORT_OCF_1_1 = "pacemaker-3.7"
 REJOIN_COUNT = 60
 REJOIN_INTERVAL = 10
 DC_DEADTIME_DEFAULT = 20
+TERMINAL_MAX_WIDTH = '500'
 
 ADVISED_ACTION_LIST = ['monitor', 'start', 'stop', 'promote', 'demote']
 ADVISED_KEY_LIST = ['timeout', 'interval', 'role']
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -779,7 +779,9 @@ def ext_cmd(cmd, shell=True):
     if options.regression_tests:
         print(".EXT", cmd)
     logger.debug("invoke: %s", cmd)
-    return subprocess.call(cmd, shell=shell)
+    env = os.environ.copy()
+    env["COLORTERM"] = constants.TERMINAL_MAX_WIDTH
+    return subprocess.call(cmd, shell=shell, env=env)
 
 
 def ext_cmd_nosudo(cmd, shell=True):


nicholasyang2022:
When I test it further, I found that using `subprocess.call(cmd, shell=True)` in a python file, without using interactive interpreter, does not make `ps -ef` to cut its output to terminal width.

It seems that crmsh messes up tty states and makes `ps` not to work as expected. I'd like to find the root cause.

nicholasyang2022:
Further inspection show that environment variable `COLUMNS` is set, making `ps` to limit its output width.

```patch
diff --git a/crmsh/utils.py b/crmsh/utils.py
index b3e3232d..ed7123c5 100644
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -777,6 +777,12 @@ def show_dot_graph(dotfile, keep_file=False, desc="transition graph"):
 
 
 def ext_cmd(cmd, shell=True):
+    print('--------')
+    for k, v in os.environ.items():
+        print(f'{k}={v}')
+    print('--------')
+    subprocess.call('env', shell=shell)
+    print('--------')
     cmd = add_sudo(cmd)
     if options.regression_tests:
         print(".EXT", cmd)
```

```
--------
LANG=en_US.UTF-8
COLORTERM=1
TERM=xterm-256color
MAIL=/var/mail/root
PATH=/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin:/usr/local/sbin:/usr/lib/pacemaker
LOGNAME=root
USER=root
HOME=/root
SHELL=/bin/bash
SUDO_COMMAND=/usr/sbin/crm resource secret stonith_ipmi_CN2 stash passwd
SUDO_USER=suse
SUDO_UID=1000
SUDO_GID=100
OCF_ROOT=/usr/lib/ocf
--------
LANG=en_US.UTF-8
SUDO_GID=100
COLORTERM=1
SUDO_COMMAND=/usr/sbin/crm resource secret stonith_ipmi_CN2 stash passwd
USER=root
PWD=/home/suse
LINES=51
HOME=/root
SUDO_USER=suse
OCF_ROOT=/usr/lib/ocf
SUDO_UID=1000
COLUMNS=81
MAIL=/var/mail/root
SHELL=/bin/bash
TERM=xterm-256color
SHLVL=1
LOGNAME=root
PATH=/usr/sbin:/usr/bin:/sbin:/bin:/usr/local/bin:/usr/local/sbin:/usr/lib/pacemaker
_=/usr/bin/env
--------
```

However, this behavior of `subprocess.run` cannot be reproduced with a minimum python script `test.py`:

```python
#!/usr/bin/python3

import subprocess
subprocess.call('env', shell=True)
```

`./test.py | grep COLUMNS` prints nothing.

liangxin1300:
So, seems unset COLUMNS or setting it a big enough one is the workaround, until this bug got fixed?

nicholasyang2022:
> So, seems unset COLUMNS or setting it a big enough one is the workaround, until this bug got fixed?

Unset `COLUMNS` and use `subprocess.call(..., env=os.environ)`, or `readline` will modify environment variables when `subprocess.call` is invoked.

PR #1356, opened 2024-03-13
[crmsh-4.6] Fix: sh: pass env to child process explicitly (bsc#1205925)

nicholasyang2022:
A bug in python stardard library, python/cpython#100516, make
environment variables, `COLUMNS` and `ROWS` to be added at fork. These
environment variables may make some of the commands called by crmsh to
truncate its output even if those commands are not writing to the
terminal.

Passing argument `env` explicitly is a workaround for the python bug.

### DIFF
--- a/crmsh/report/sh.py
+++ b/crmsh/report/sh.py
@@ -69,6 +69,7 @@ class Shell:
         return subprocess.run(
             args,
             input=cmd.encode('utf-8'),
+            env=os.environ,  # bsc#1205925
             **kwargs,
         )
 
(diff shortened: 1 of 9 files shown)


nicholasyang2022:
[Further experiment](https://gist.github.com/nicholasyang2022/ae36c7f2695b086c0fa224f3325838cd) shows that the behavior of modifying environ `COLUMNS` and `ROWS` exists in `libreadline`, instead of the python wrapper.

nicholasyang2022:
CPython backtrace: https://gist.github.com/nicholasyang2022/ce30ffc409ef3069f8b5628fa4b3d429

nicholasyang2022:
The backtrace shows that `COLUMNS` is set when `rl_initialize()` is called from `import readline`. I had believed the environ is set at fork. However, this is not true. I use `/proc/$pid/environ` to determine whether `COLUMNS` is set. This is also unreliable as it is not updated instantly when `setenv(3)` is called.

PR #1366, opened 2024-03-21
Dev: bootstrap: Give a warning when detecting $SSH_AUTH_SOCK but not use --use-ssh-agent option

liangxin1300:


crmsh/bootstrap.py
```diff
@@ -143,7 +143,6 @@ def __init__(self):
                 COROSYNC_AUTH, "/var/lib/heartbeat/crm/*", "/var/lib/pacemaker/cib/*",
                 "/var/lib/corosync/*", "/var/lib/pacemaker/pengine/*", PCMK_REMOTE_AUTH,
                 "/var/lib/csync2/*", "~/.config/crm/*"]
-        self.use_ssh_agent = False
```
nicholasyang2022, line 146: Do not abuse dynamic language features. Adding fields dynamically make code unreadable.

liangxin1300: Some `crm cluster` subcommands don't have `--use-ssh-agent`, like `crm cluster remove`
I made this change to distinguish if the user used this option or not.

liangxin1300: Or, we can add a flag to mark which option is from argparse while calling
```
@classmethod
    def set_context(cls, options):
        ctx = cls()
        for opt in vars(options):
            setattr(ctx, opt, getattr(options, opt))
        ctx.initialize_user()
        return ctx

```

nicholasyang2022: `None` is enough for marking a field as not set.

nicholasyang2022 (approved):

PR #1369, opened 2024-03-25
[crmsh-4.6] Fix: utils: Consider both hostname and IP address when removing node …

liangxin1300:
…from a node list

Since utils.list_cluster_nodes might return an IP list

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1844,9 +1844,22 @@ def list_cluster_nodes_except_me():
     node_list = list_cluster_nodes()
     if not node_list:
         raise ValueError("Failed to get node list from cluster")
+    return remove_local_from_node_list(node_list)
+
+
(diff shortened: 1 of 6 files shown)


crmsh/utils.py
```diff
@@ -1844,9 +1844,22 @@ def list_cluster_nodes_except_me():
...
+def remove_local_from_node_list(node_list: typing.List[str]) -> typing.List[str]:
+    """
+    Remove local node from node list
+    Consider both hostname and IP address cases
+    """
     me = this_node()
-    if me in node_list:
+    if me in node_list[:]:
         node_list.remove(me)
+    else:
```
liangxin1300, line 1858: Don't know why this `else` block not be covered by the below behave tests
Function `remove_local_from_node_list` was called by `fix_cluster`, which seems be running directly, not under subprocess like the crm report

PR #1398, opened 2024-04-22
[crmsh-4.6] Dev: adds a new command $ crm check migration

aleksei-burlakov:
This new command checks if the migration from sle15 to sle16 is safe.
(ref: PED-8270)


nicholasyang2022 (commented):
It is not important to check whether these supported packages are installed before os migration. For example, we don't need libknet1 until corosync is upgraded to 3, and upgrading corosync to 3 will pull in libknet1 automatically as a dependency.

It is more important to make sure whether unsupported packages are not  installed or not used. For example, the old SAP resource agent should not be installed or used in cib.

liangxin1300:
Close this
The final implementation is https://github.com/ClusterLabs/crmsh/pull/1629

PR #1405, opened 2024-04-26
Fix: bootstrap: open corosync ports in firewalld

liangxin1300:
- Drop SuSEfirewall2 support
- bootstrap: open corosync ports in firewalld


nicholasyang2022:
Can you cherry-pick #1406? It will not pass patch coverage check without calling `firewall_open_corosync_ports`.

nicholasyang2022 (commented):
The supported features are not aligned in bootstrap and crash_test, including:

1. bootstrap supports both firewalld and ufw, and crash_test supports only firewalld
2. bootstrap allows firewalld installed but not started, while crash_test requires firewalld running if it is installed.

liangxin1300:
> The supported features are not aligned in bootstrap and crash_test, including:
> 
> 1. bootstrap supports both firewalld and ufw, and crash_test supports only firewalls

I think it's worth adding an issue.

> 2. bootstrap allows firewalld installed but not started, while crash_test requires firewalld running if it is installed.

crash_test requires stonith enabled, not firewalld running

PR #1406, opened 2024-04-27
Fix: corosync: ConfigParser.get_all(nonexsistent) should return an empty list

nicholasyang2022:


liangxin1300 (approved):
Works for me
Thanks!

nicholasyang2022:
Ah! I just added another commit and I have not added any unit test.

PR #1408, opened 2024-04-28
Dev: ui_context: make help subcommands to exit with 0

nicholasyang2022:
port:

* #1375

revert:

* 9148684b980b30291de91d9e203980ee08a610d9 which is merged by mistake

and

* make `ui_clsuter.do_start` to parse cmdline args before doing any other check


liangxin1300 (approved):
Thanks!

PR #1410, opened 2024-04-28
[crmsh-4.5] Fix: healthcheck: KeyError when local nodename not found in cib (bsc#1223438)

nicholasyang2022:
Log a warning and continue, as join_ssh_impl should work even if local host is included in the node list.


nicholasyang2022:
I cannot cover this in functional tests. `upgrade_if_needed` aborts when it `utils.can_ask` returns False. The result of `utils.can_ask` is unreliable and I cannot make it return True in new test cases anymore. This problem needs to be addressed with another pull request.

liangxin1300 (approved):

PR #1414, opened 2024-05-06
Dev: utils: Add info when property is newly added

liangxin1300:
case 1
```
# first node init
no 'priority' property

# then the second node join
set 'priority=1' with no warning/info to show this behavior
```
case 2
```
# two-node cluster running
# remove the second node
show:
WARNING: "priority" in rsc_defaults is set to 0, it was 1
```
case 3
```
# removed node re-join
show:
WARNING: "priority" in rsc_defaults is set to 1, it was 0
```

This PR adds info when the property is newly added, for case 1


nicholasyang2022 (commented):
Please also change line 2803 to use `%s` instead of `{}`.

nicholasyang2022 (approved):

PR #1418, opened 2024-05-08
Dev: Change the RA name string format

liangxin1300:
OCF resource agents will be displayed as ocf:PROVIDER:AGENT rather than ocf::PROVIDER:AGENT in crm_mon and crm_simulate output and pacemaker-schedulerd logs


nicholasyang2022:
Please add PED reference to the title line of the commit message so that it will get picked up in changelogs.

nicholasyang2022 (approved):

PR #1423, opened 2024-05-20
Dev: Hide lsb and service resource agent class type from UI and doc

liangxin1300:


crmsh/ra.py
```diff
@@ -50,8 +50,9 @@ def ra_classes():
         return cache.retrieve("ra_classes")
     if can_use_crm_resource():
         l = crm_resource("--list-standards")
+        l = [x for x in l if x not in ("lsb", "service")]
     else:
-        l = ["heartbeat", "lsb", "ocf", "stonith", "systemd"]
```
nicholasyang2022, line 54: How is `heartbeat` related to `lsb`?

nicholasyang2022 (commented):
Is there a PED to be referenced?

zzhou1:
> Is there a PED to be referenced?

Let's wait few moment till the Impl. Task PED get created.

crmsh/ra.py
zzhou1: "heartbeat" is used by the current ocf1.0 and 1.1 standard and located at
/usr/lib/ocf/resource.d/heartbeat

liangxin1300:
> > Is there a PED to be referenced?
> 
> Let's wait few moment till the Impl. Task PED get created.

Now I write jsc#PED-5577 in commit message

crmsh/ra.py
liangxin1300:
```
# crm_resource --list-standards
ocf
lsb
service
systemd
stonith
```
Just synced with the output of --list-standards

nicholasyang2022 (approved):

PR #1425, opened 2024-05-27
Dev: log an error when fencing node without stonith device configured and running

liangxin1300:
When running `crm node fence` or `crm cluster crash_test --fence-node`, should check if stonith device configured and running, if not, log an error and return False


nicholasyang2022 (approved):
LGTM

PR #1452, opened 2024-06-13
Doc improvement

liangxin1300:
- Remove 'moon' from doc, since it has been deprecated since pacemaker commit 283ab7079
- Adjust doc for deprecated multi-rule within a location constraint

### DIFF
--- a/doc/crm.8.adoc
+++ b/doc/crm.8.adoc
@@ -824,7 +824,6 @@ duration|date_spec ::
          | weeks=<value>
          | years=<value>
          | weekyears=<value>
-         | moon=<value>
 ..............
 
 [[topics_Lifetime,Lifetime parameter format]]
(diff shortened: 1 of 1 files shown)


doc/crm.8.adoc
```diff
@@ -3110,7 +3111,7 @@ load xml push smallcib.xml
 ==== `location`
 
 `location` defines the preference of nodes for the given
-resource. The location constraints consist of one or more rules
+resource. The location constraints consist of one rule
 which specify a score to be awarded if the rule matches.
```
nicholasyang2022, line 3115:
```suggestion
resource. The location constraints are defined by a single rule
which specifies a score to be awarded if the rule matches.
```

nicholasyang2022 (approved):

PR #1453, opened 2024-06-13
[crmsh-4.5] Fix: upgradeutil: refine error handling (bsc#1226147, bsc#1223371)

nicholasyang2022:
Getting upgrade seq from remote node may fails for various reasons. Previous code handles SSH problems gracefully but other problems are note handled and turn into misleading error messages.

This commit handles those problems in the same way in which SSH problems are handled: output detail information to debug log and skip upgrading.


liangxin1300 (approved):
LGTM

Will this be ported to crmsh-4.6 and master?

PR #1454, opened 2024-06-13
Fix: cibconfig: Disable auto complete advised operation values when adding a rsc_template

### LINKED ISSUE #1438: There is an error when configuring primitive resource with rsc_template
Run
```
crm configure rsc_template dummy_template ocf:pacemaker:Dummy op monitor interval=12s
```

Then
```
crm configure primitive d1 @dummy_template params passwd=123
```

Got error
```
ERROR: None:None: got no meta-data, does this RA exist? 
```

liangxin1300:
Run
```
crm configure rsc_template dummy_template ocf:pacemaker:Dummy op monitor interval=12s
```

Then
```
crm configure primitive d1 @dummy_template params passwd=123
```

Got error
```
ERROR: None:None: got no meta-data, does this RA exist? 
```

Solution: When adding a rsc_template as a primitive resource, should disable auto complete advised operation values

Fix #1438

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -793,7 +793,7 @@ def id_for_node(node, id_hint=None):
     return obj_id
 
 
-def postprocess_cli(node, oldnode=None, id_hint=None, complete_advised=False):
+def postprocess_cli(node, oldnode=None, id_hint=None, promotable_default_meta=False):
     """
     input: unprocessed but parsed XML
     output: XML, obj_type, obj_id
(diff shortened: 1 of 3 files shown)


nicholasyang2022:
The regression tests fails. Do you need to rebase this?

nicholasyang2022:
This problem happens when executing the command instead of activating the completion. I think completion related code should not be called here. This patch does not fix the root cause.

liangxin1300:
> This problem happens when executing the command instead of activating the completion. I think completion related code should not be called here. This patch does not fix the root cause.

No, it has no relation with completion here.
Maybe the var term `complete_advised` confuses?
That means to complete RA's advised operation values if user not specify them

nicholasyang2022:
> Maybe the var term complete_advised confuses?

Yes, appearing together with "cli" and "parse", it is so confusing. Can you change to a clearer variable name?

liangxin1300:
> > Maybe the var term complete_advised confuses?
> 
> Yes, appearing together with "cli" and "parse", it is so confusing. Can you change to a clearer variable name?

Already changed the name of `complete_advised`
Regarding "cli" and "parse", since they are largely used and beyond the scope of this PR, I suggest to open an issue to trace?

nicholasyang2022 (approved):

PR #1458, opened 2024-06-17
Dev: coverage: inject coverage with sitecustomize.py in feature tests

nicholasyang2022:
With python-coverage updated to 7.5.3 from 4.x, code called with `multiprocessing` is able to be measured now.

This is an refreshed version of #1297.


liangxin1300 (approved):
Nice!

PR #1461, opened 2024-06-19
Dev: ui_configure: Complete required parameters first

liangxin1300:
When adding an RA, if that RA has required parameters, 
sort the completion results by required parameters and then alphabetically
```
crm(live/alp-1)configure# primitive vip IPaddr2 params 
ip=                        arp_count_refresh=         broadcast=                 flush_routes=              lvs_ipv6_addrlabel_value=  network_namespace=         noprefixroute=             send_arp_opts=
arp_bg=                    arp_interval=              cidr_netmask=              iflabel=                   lvs_support=               nic=                       preferred_lft=             unique_clone_address=
arp_count=                 arp_sender=                clusterip_hash=            lvs_ipv6_addrlabel=        mac=                       nodad=                     run_arping=                
crm(live/alp-1)configure# primitive fs Filesystem params 
device=             fstype=             force_clones=       kill_signals=       run_fsck=           statusfile_prefix=  
directory=          fast_stop=          force_unmount=      options=            signal_delay=       term_signals=
```

### DIFF
--- a/crmsh/command.py
+++ b/crmsh/command.py
@@ -8,6 +8,8 @@
 
 import inspect
 import re
+import ctypes
+import ctypes.util
 from . import help as help_module
 from . import ui_utils
 from . import log
@@ -569,6 +571,10 @@ class ChildInfo(object):
         '''
         ret = []
         if self.completer is not None:
+            if sort_completion_inst is not None and sort_completion_inst.value == 0:
+                # Restore the original value of rl_sort_completion_matches
+                # before calling the completer again
+                sort_completion_inst.value = orig_sort_completion_value
             specs = inspect.getfullargspec(self.completer)
             if 'context' in specs.args:
                 ret = self.completer([self.name] + args, context)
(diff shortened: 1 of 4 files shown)


liangxin1300:
@zzhou1 @nicholasyang2022 
I tried to show all parameters at once with the required ones sorted in the beginning
But it seems Python readline module always sorts the completion results in alphabetical order
So, I think this current solution is not bad and we can move on

liangxin1300:
> > readline module always sorts the completion results in alphabetical order
> 
> This is configurable. See https://stackoverflow.com/questions/31229708/python-gnu-readline-bindings-keep-my-sort-order.

Thanks! I will take a look

crmsh/utils.py
```diff
@@ -3107,4 +3110,32 @@ def time_value_with_unit(time_value):
...
+        if readline_path:
+            # Inspired by
+            # https://stackoverflow.com/questions/31229708/python-gnu-readline-bindings-keep-my-sort-order
+            rl = ctypes.cdll.LoadLibrary(readline_path)
+            sort = ctypes.c_ulong.in_dll(rl, 'rl_sort_completion_matches')
+            sort.value = 0
+            result = func(*args, **kwargs)
+            # Restore the default sort order later,
+            # after the completion matches are done.
+            timer = threading.Timer(1, lambda: setattr(sort, 'value', 1))
```
nicholasyang2022, line 3135: This looks unreliable. Cab we just disable sort at readline level globally, and control whether to sort in our completers?

nicholasyang2022: Can we pass a flag representing _whether to sort_ along together with completion candidates, and change option `sort` accordingly just before the completion starts?

liangxin1300: > Can we pass a flag representing _whether to sort_ along together with completion candidates, and change option `sort` accordingly just before the completion starts?

Not get your point yet.
Using this decorator means disabling the default readline sort behavior

nicholasyang2022: Let's forget the decorator. We are calling readline to get some input from user. So the most suitable way to control whether to sort the candidates, is to set the option just before we call readline, and restore it after readline returns. What we need to do is pass this option down where readline is called, which should be in the RPEL I think.

liangxin1300: > restore it after readline returns
Don't clear how to do that, so I reset it before calling completer again

nicholasyang2022 (approved):

Issue #1460, opened 2024-06-18
'crm configure delete' don't complete existing id like before


liangxin1300:
It should be a regression issue which was introduced by #1403

Issue #1466, opened 2024-06-20
crmsh lacks a prompt asking whether to commit the changes


liangxin1300:
Which should be a regression issue

on 15SP5
```
crm(live/15sp5-1)configure# primitive d Dummy
crm(live/15sp5-1)configure# up
There are changes pending. Do you want to commit them (y/n)? n
crm(live/15sp5-1)# configure
crm(live/15sp5-1)configure# edit
crm(live/15sp5-1)configure# up
There are changes pending. Do you want to commit them (y/n)?
```
For master code
```
crm(live/alp-2)configure# primitive d Dummy
crm(live/alp-2)configure# up
crm(live/alp-2)# configure
crm(live/alp-2)configure# edit
crm(live/alp-2)configure# up
```

Issue #1355, opened 2024-03-11
`crm configure help XXX` is broken if the cluster is not running


zzhou1:
# crm configure help show
ERROR: running cibadmin -Ql: Could not connect to the CIB: Transport endpoint is not connected

Indeed, "--help" works as expected.
# crm configure show --help

liangxin1300:
Fix in #1300

PR #1469, opened 2024-06-21
Fix: ui_context: crm configure delete autocompletion #1403 #1460

aleksei-burlakov:
If you type
  $ crm configure delete
of do it interactively
  $ crm configure
  crm(live/node1)configure# delete
and then double press tab it, should propose possible resource options. And this commit makes it working again.

### DIFF
--- a/crmsh/ui_context.py
+++ b/crmsh/ui_context.py
@@ -11,6 +11,7 @@ from . import userdir
 from . import constants
 from . import log
 from . import main
+from .service_manager import ServiceManager
 
 
 logger = log.setup_logger(__name__)
@@ -255,6 +256,9 @@ class Context(object):
         self._in_transit = True
 
         entry = level()
+        if ServiceManager().service_is_active("pacemaker.service"):
+            if 'requires' in dir(entry) and not entry.requires():
+                self.fatal_error("Missing requirements")
         self.stack.append(entry)
         self.clear_readline_cache()
 


crmsh/ui_context.py
```diff
@@ -255,6 +255,8 @@ def enter_level(self, level):
         self._in_transit = True
 
         entry = level()
+        if 'requires' in dir(entry) and not entry.requires():
+            self.fatal_error("Missing requirements")
```
liangxin1300, line 259: No, this will bring the issue https://github.com/ClusterLabs/crmsh/issues/1355 back, which was fixed in https://github.com/ClusterLabs/crmsh/pull/1300

crmsh/ui_context.py
```diff
@@ -255,6 +255,9 @@ def enter_level(self, level):
         self._in_transit = True
 
         entry = level()
+        if self.command_name == 'configure' and self.command_args and self.command_args[0] in ['delete']:
```
liangxin1300, line 258: I found issue https://github.com/ClusterLabs/crmsh/issues/1466 might have the same root cause
It works after running `configure refresh`

So here we should think a common solution, not only for the `delete` command

liangxin1300: Yes, https://github.com/ClusterLabs/crmsh/pull/1403 cause this and #1466 both

aleksei-burlakov: Hi @liangxin1300 , could you please check the update?

liangxin1300 (approved):
Thanks, @aleksei-burlakov !

liangxin1300:
Please backport this to crmsh-4.6

liangxin1300:
And also take a look at issue #1466 when you have time
Thanks!

Issue #1472, opened 2024-06-24
Codecov: the coverage data of merge commits are not uploaded


nicholasyang2022:
For a pull request, codecov compares its coverage to its merge base. It should be the commit from which the pull request forks, typically a merge commits in master branch or release branches (e.g. crmsh-4.6).

However, as the coverage data from merge commits are not uploaded, codecov is only able to use the coverage data from the last pull request. If multiple pull requests interleaves, these data are not complete, making the coverage diff inaccurate.

For example, [#1458](https://app.codecov.io/gh/ClusterLabs/crmsh/pull/1458) is compared to 8f4ac8f instead of 30a5a20d.

PR #1473, opened 2024-06-25
Fix: codecov: use environ to pass CODECOV_TOKEN (#1472)

nicholasyang2022:
As shown in https://github.com/ClusterLabs/crmsh/actions/runs/9612091155/job/26512047071, codecov reports: Error: Codecov token not found. Please provide Codecov token with -t flag.

So try a different way to provide the token.


liangxin1300 (approved):
Thanks!

nicholasyang2022:
Does not work either. Maybe the value in action secrets is wrong.  I will update the secret and try again.

PR #1475, opened 2024-06-26
DC lost during wait

freishutz:
The initial error that we faced during our online storage upgrade:
```
Command: x Exascaler Install: apply_lustre_params,create_udev_rules,email,emf_agent,emf_node_manager,ha,hosts,ipmi,kdump,logging,lustre,lvm,mdt_backup,modprobe,nics,ntp,os,ost_pools,restart_network,serial,start_cluster,sync_exa_toml (Config ver. 1) failed
User: api

  Job: x es-install --steps start_cluster on node5 failed

    Step: x Run config-pacemaker on node5 failed (took: 12s 534ms 171us 586ns)
    Result (Error):
      Bad Exit Code: 1.
    Started: 2024-02-07T03:26:16.158Z
    Ended: 2024-02-07T03:26:28.692Z
    Stdout:
      Running Command: config-pacemaker --unmanaged-emf
    Stderr:
      x Command has failed.
      Code: exit status: 1
      Stdout: INFO: cib.commit: committed '5e8558de-1ceb-46c2-bd70-1ab4d8504c9f' shadow CIB to the cluster

      Stderr: WARNING: DC lost during wait
```

Basically, the source of our problems below (case 3 - DC election or voiting during cluster recalculation):

```
[root@es-1-virt1 ~]# crmadmin -D -t 1; echo $?
Designated Controller is: es-2-virt1
0

[root@es-1-virt1 ~]# crm cluster stop
INFO: The cluster stack stopped on es-1-virt1
[root@es-1-virt1 ~]# crmadmin -D -t 1; echo $?
error: Could not connect to controller: Connection refused
error: Command failed: Connection refused
102

[root@es-1-virt1 ~]# crm cluster start
INFO: The cluster stack started on es-1-virt1
[root@es-1-virt1 ~]# crmadmin -D -t 1; echo $?
error: No reply received from controller before timeout (1000ms)
error: Command failed: Connection timed out
124
```
Potentially, we have a deadloop in dc_waiter, but it also means that pacemaker in the same state and in worst case the amount of time should not be more than 'dc-deadtime'.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -991,12 +991,12 @@ def append_file(dest, src):
 
 def get_dc(peer=None):
     cmd = "crmadmin -D -t 1"
-    _, out, _ = sh.cluster_shell().get_rc_stdout_stderr_without_input(peer, cmd)
+    rc, out, _ = sh.cluster_shell().get_rc_stdout_stderr_without_input(peer, cmd)
     if not out:
-        return None
+        return (None, rc)
     if not out.startswith("Designated"):
-        return None
-    return out.split()[-1]
+        return (None, rc)
+    return (out.split()[-1], rc)
 
 
 def wait4dc(what="", show_progress=True):
@@ -1022,8 +1022,21 @@ def wait4dc(what="", show_progress=True):
     There's no timeout, as we expect the DC to eventually becomes
     idle.
     '''
-    dc = get_dc()
-    if not dc:
+    def dc_waiter():
+        while True:
+            dc, rc = get_dc()
+            if rc == 0:
+                return dc
+            if rc == 102 or rc == 1:
+                logger.warning("Could not connect to the controller: Connection refused")
+                return None
+            if rc == 124:
+                logger.warning("No reply received from the controller before timeout")
+                continue
+            logger.warning("Unknown return code from crmadmin: %d", rc)
+            return None
+
+    if not dc_waiter:
         logger.warning("can't find DC")
         return False
     cmd = "crm_attribute -Gq -t crm_config -n crmd-transition-delay 2> /dev/null"
@@ -1039,7 +1052,7 @@ def wait4dc(what="", show_progress=True):
     max_sleep = 1.00
     sleep_time = init_sleep
     while True:
-        dc = get_dc()
+        dc = dc_waiter()
         if not dc:
             logger.warning("DC lost during wait")
             return False
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -989,14 +989,16 @@ def append_file(dest, src):
         return False
 
 
-def get_dc(peer=None):
+def get_dc(peer=None, show_rc=False):
     cmd = "crmadmin -D -t 1"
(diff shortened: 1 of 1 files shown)


freishutz:
> 1. Can you describe your use case from crmsh perspective? I don't known what `config-pacemaker --unmanaged-emf` does.
> 
>     2. `get_dc` is called from several other places. We cannot simply change its return value.

1. Inside of config-pacemaker we apply CIB rules via crmsh. The problem here not in config-pacemaker, but in a moment of time when we start the cluster and we are waiting DC, but crmsh fails since wait4dc is not smart enough and relies on timings. Config-pacemaker simply tries to apply and failed due to crmsh error.
2.  I see next things:
```
anovik@anovikpc:~/repo/crmsh-whamcloud$ grep -r get_dc .
./crmsh/report/collect.py:    node = crmutils.get_dc()
./crmsh/ui_history.py:        if not utils.get_dc():
./crmsh/ui_cluster.py:            utils.check_function_with_timeout(utils.get_dc, wait_timeout=dc_timeout, peer=node)
./crmsh/utils.py:def get_dc(peer=None):
./crmsh/utils.py:            dc, rc = get_dc()
./test/features/steps/step_implementation.py:        if crmutils.get_dc():
./test/unittests/test_report_collect.py:    @mock.patch('crmsh.utils.get_dc')
./test/unittests/test_report_collect.py:    def test_dump_runtime_state(self, mock_run, mock_str2file, mock_debug, mock_get_dc, mock_this_node, mock_real_path):
./test/unittests/test_report_collect.py:        mock_get_dc.return_value = "node1"
```
So, I didn't use `crm report`, so don't know how it will react.

Maybe the fastest way to fix and don't reak will be the second flag in the function like:
```
def get_dc(peer=None, show_rc=False):
....
  if show_rc:
     return (None, rc)
  else:
     return None
```

We are using patched version of crmsh (with current patch) for last couple of months, so at least it helps in our case.

nicholasyang2022:
Thank you. I get the problem. `wait4dc` traces DC transition until it gets stable. However, at certain steps during the transition, there will be no DCs in the cluster. `wait4dc` does not handle this situation and fails. This function name is misleading in a way. I think `wait_dc_stable`/`wait_dc_idle` is better.

And there is a related issue fixed in #1232. It provides `crmsh.ui_cluster.Cluster._wait_for_dc` doing a similar thing like what `dc_waiter` in the patch does. We can replace it with a more general version.

So IMO, we can:

1. Rename `crmsh.utils.wait4dc` to `crmsh.utils.wait_dc_stable`.
2. Keep `get_dc()` untouched to avoid potential regressions.
3. Combine `dc_waiter()` in the patch and `crmsh.ui_cluster.Cluster._wait_for_dc()`, and move them to a toplevel function `crmsh.utils.wait_for_dc`.
4. When testing the exit status of `crmadmin`, define constants (see https://github.com/ClusterLabs/pacemaker/blob/main/include/crm/common/results.h) instead of using magic numbers. (Or should we generates these constants directly from pacemaker headers?)

@liangxin1300 What do you think?

PR #1479, opened 2024-07-01
Dev: bootstrap: Use the existing function to query cluster property

liangxin1300:
The command `crm_attribute --attr-name no-quorum-policy --delete-attr` didn't work, will get
```
crm_attribute: Missing argument for --delete-attr
```

And adding a new function to delete the cluster property.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2779,6 +2779,17 @@ def get_property(name, property_type="crm_config", peer=None):
     return stdout if rc == 0 else None
 
 
+def delete_property(name, property_type="crm_config") -> bool:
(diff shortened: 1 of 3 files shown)


crmsh/utils.py
```diff
@@ -2779,6 +2779,15 @@ def get_property(name, property_type="crm_config", peer=None):
     return stdout if rc == 0 else None
 
 
+def delete_property(name, property_type="crm_config"):
```
nicholasyang2022, line 2782: Should provide a feedback about whether the operation succeeds. Please add a return value or raise an exception when it fails.

nicholasyang2022 (approved):

PR #1481, opened 2024-07-04
Fix: ui_context: crm configure up prompt #1466

aleksei-burlakov:
The regression was introduced in beb26f3e
Before beb26f3e it was like
```
crm(live/15sp5-1)configure# primitive d Dummy
crm(live/15sp5-1)configure# up
There are changes pending. Do you want to commit them (y/n)?
```
After beb26f3e there is no prompt `There are changes pending. Do you want to commit them (y/n)?`
This change brings the prompt back.

### DIFF
--- a/crmsh/ui_context.py
+++ b/crmsh/ui_context.py
@@ -326,7 +326,7 @@ class Context(object):
         '''
         ok = True
         if len(self.stack) > 1:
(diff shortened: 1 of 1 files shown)


liangxin1300 (approved):
Thanks!

Please backport to crmsh-4.6

nicholasyang2022:
It seems that this patch breaks regression tests.

PR #1483, opened 2024-07-08
Fix: DC lost during wait

nicholasyang2022:
This is a refined version of #1475.

`crmsh.utils.wait4dc` traces DC transition until it gets stable. However, at certain steps during the transition, there will be no DCs in the cluster. wait4dc does not handle this situation and fails.

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -991,12 +991,12 @@ def append_file(dest, src):
 
 def get_dc(peer=None):
     cmd = "crmadmin -D -t 1"
-    _, out, _ = sh.cluster_shell().get_rc_stdout_stderr_without_input(peer, cmd)
+    rc, out, _ = sh.cluster_shell().get_rc_stdout_stderr_without_input(peer, cmd)
     if not out:
-        return None
+        return (None, rc)
     if not out.startswith("Designated"):
-        return None
-    return out.split()[-1]
+        return (None, rc)
+    return (out.split()[-1], rc)
 
 
 def wait4dc(what="", show_progress=True):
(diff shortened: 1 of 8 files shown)


crmsh/constants.py
```diff
@@ -454,4 +454,9 @@
 
 # Commands that are deprecated and hidden from UI
 HIDDEN_COMMANDS = {'ms'}
+
+# pacemaker crm/common/results.h
```
freishutz, line 458: Just hit another one:
```
      Stderr: �[33mWARNING�[0m: Unknown return code from crmadmin: 113
      �[33mWARNING�[0m: DC lost during wait
```
which is  
`CRM_EX_UNSATISFIED          = 113, //!< Requested item does not satisfy constraints`
and it seems like this case mapped on `./lib/common/results.c` in pacemaker:
```
        case EAGAIN:
        case EBUSY:
            return CRM_EX_UNSATISFIED;
```
So, worth to ignore that RC and give another try.

crmsh/utils.py
```diff
@@ -999,7 +999,34 @@ def get_dc(peer=None):
...
 
-def wait4dc(what="", show_progress=True):
+def wait_for_dc(node: str = None):
+    """
+    Wait for the cluster's DC to become available
+    """
+    if not ServiceManager().service_is_active("pacemaker.service", remote_addr=node):
+        raise ValueError("Pacemaker is not running. No DC.")
+    dc_deadtime = get_property("dc-deadtime", peer=node) or str(constants.DC_DEADTIME_DEFAULT)
+    dc_timeout = int(dc_deadtime.strip('s')) + 5
```
freishutz, line 1009: `dc-deadtime` value  takes `time` and it could be `s|m|h` and maybe even `d` days.

crmsh/constants.py
nicholasyang2022: So, just reusing the original version of `get_dc`, ignoring the exit status, is good enough.

freishutz (approved):
I think that if nothing happen in `dc_timeout = crm_msec(dc_deadtime) // 1000 + 5` then this should work as expected.

liangxin1300 (approved):
Thanks!

PR #1486, opened 2024-07-10
Dev: refine change detection for `crm configure` (#1466)

nicholasyang2022:
#1300 tried to remove unnecessary dependencies on a running pacemaker service when crmsh is not actually going to call the service. However, it removed the call to `end_game()` from `up()` and `quit()` incorrectly. This made sublevel `configure` not to have a chance to check uncommitted changes anymore, leading to a regression described in #1466.

#1481 tried to fix this regression by adding the call to `end_game()` back to `up()`, and check if pacemaker service is running before calling `end_game()`. This is not optimal as `up()` is a common method used for all sublevels and the status of pacemaker service is only related sublevel `configure`. Checking the status of pacemaker service when using other sublevels does not make sense.

This pull request uses a more straightforward method to fix the problem: improve how to detect uncommitted changes in `has_cib_changes()`.

The original implementation is to check if there is any elements in change queue. This requires the in memory representation of cib to be populated, leading to an indirect dependency to pacemaker service. However, we can take advantage of a property: the in memory cib needs to be populated before doing any changes. So in the implementation can be optimized: if the in memory cib is not populated, we will know there is not any change.

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -2769,14 +2769,18 @@ class CibFactory(object):
                 # no need to warn, user can see the object displayed as XML
                 logger.debug("object %s cannot be represented in the CLI notation", obj.obj_id)
 
-    def initialize(self, cib=None):
+    def initialize(self, cib=None, no_side_effects=False):
         if self.cib_elem is not None:
             return True
         if cib is None:
-            cib = read_cib(cibdump2elem)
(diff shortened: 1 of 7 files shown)


liangxin1300 (approved):
LGTM!
Thanks!

liangxin1300:
Please backport to crmsh-4.6

PR #1488, opened 2024-07-11
Fix: hahealth.py script

aleksei-burlakov:
There are two changes:
1) use tar gz (now reports are tar.gz)
2) remove -D flag from the crm_report (obsolete)


liangxin1300 (approved):

liangxin1300:
Thanks!

PR #1489, opened 2024-07-16
Fix: qnetd fails to start when TLS is disabled (bsc#1227649)

nicholasyang2022:
When TLS for qdevice/qnetd communication is disabled, the current implementation changes only the configuration on the client side (qdevice), and the server side (qnetd) still uses default value: "on". As a result, qnetd needs to be prepared for accepting incoming TLS connections. This requires CA and server certificates to be generated.

This pull request ensures CA and server certificates present on qnetd node and qnetd service started before starting to configure qdevice.

### DIFF
--- a/crmsh/qdevice.py
+++ b/crmsh/qdevice.py
@@ -97,28 +97,12 @@ def qnetd_lock_for_multi_cluster(func):
 
 
 class QDevice(object):
-    """
-    Class to manage qdevice configuration and services
+    """Class to manage qdevice configuration and services
 
-    Whole certification process:
-    For init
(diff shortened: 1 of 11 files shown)


crmsh/qdevice.py
```diff
@@ -102,7 +102,6 @@ class QDevice(object):
 
     Whole certification process:
     For init
-    Step 1:  init_db_on_qnetd
```
liangxin1300, line 105: Then the docstring here needs to be adjusted, starts with 'Step 2' looks strange

liangxin1300 (approved):
Thanks!

PR #1490, opened 2024-07-16
Dev: github-actions: actions/setup-python@v4 is deprecated

nicholasyang2022:


liangxin1300 (approved):

liangxin1300:
Is there any alias name to always use the latest one so that we don't need to care about the version?

Issue #1442, opened 2024-06-03
Add an option to disable auto-config advised resource actions


liangxin1300:
Which was introduced from #1038

PR #1494, opened 2024-07-24
Add an option 'add_advised_op_values' to disable adding advised op values

liangxin1300:
This PR is for issue #1442 

The default value for `add_advised_op_values` is 'yes'.
When set to 'no', the advised operation values will not be added automatically if the user has not specified them


nicholasyang2022 (approved):
LGTM

zzhou1 (approved):

PR #1497, opened 2024-07-29
Dev: crm_rpmcheck: use ansible to get package versions

aleksei-burlakov:
This PR introduces `ansible` and this change is just the right place. Instead of repeatedly calling `rpm -q ...` it calls ansible only once and gets all the package information.

### DIFF
--- a/utils/crm_rpmcheck.py
+++ b/utils/crm_rpmcheck.py
@@ -6,7 +6,7 @@ import os
 import sys
 import json
 import subprocess
-
+import shutil
 
 def run(cmd):
     proc = subprocess.Popen(cmd,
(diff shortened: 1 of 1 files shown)


utils/crm_rpmcheck.py
```diff
@@ -23,6 +23,11 @@ def package_data(pkg):
     """
     Gathers version and release information about a package.
     """
+    if os.path.isfile('/bin/ansible'):
```
nicholasyang2022, line 26: Should use `shutil.which` instead, as it works better with environ `PATH` and also checks for 'x' permission.

utils/crm_rpmcheck.py
```diff
@@ -31,6 +36,37 @@ def package_data(pkg):
 
     return {'name': pkg, 'error': "unknown package manager"}
 
+packages = None
```
nicholasyang2022, line 39: Private variables, which is not intended to be accessed directly from other modules, should start with `_`.

utils/crm_rpmcheck.py
aleksei-burlakov: My idea was to make the code uniform. So that it looked like the rest. Like 
`if os.path.isfile('/bin/rpm'):`
and
`if os.path.isfile('/usr/bin/dpkg'):`
look all the same.
I could use the `shutil.which` indeed, but that would logically require refactoring the `/bin/rpm` and `/bin/dpkg` cases.

nicholasyang2022: So this is a good time to refactor them.

liangxin1300 (approved):
LGTM, thanks!

PR #1498, opened 2024-07-30
Fix: ui_context: enter_level() should not check requirements for non-functional subcommands

nicholasyang2022:
Fix a regression introduced in ac812f17b1fd94ff489e1f0ca17e968019a08417.


liangxin1300 (approved):
Thanks!

PR #1506, opened 2024-08-05
[crmsh-4.6] Fix: Don't add time units to values for existing CIB (bsc#1228817)

liangxin1300:
## Problem
Commit 7b2cfb23 automatically appends 's' as the default time unit for timeout and interval. This causes inconsistency in the existing CIB.

## How to reproduce
1. Run `crm configure primitive vip IPaddr2 params ip=10.10.10.123 op monitor interval=10`
2. Run `crm configure show vip`, we can see `interval=10s`
    ```
    primitive vip IPaddr2 \
        params ip=10.10.10.123 \
        op monitor interval=10s timeout=20s \
        op stop timeout=20s interval=0s
    ```
3. Run `crm configure edit xml vip`, change `interval=10s` as `interval=10`, then run `crm configure show vip`
    ```
    xml <primitive id="vip" class="ocf" provider="heartbeat" type="IPaddr2"> \
        <instance_attributes id="vip-instance_attributes"> \
          <nvpair name="ip" value="10.10.10.123" id="vip-instance_attributes-ip"/> \
        </instance_attributes> \
        <operations> \
          <op name="monitor" interval="10" timeout="20s" id="vip-monitor-10s"/> \
          <op name="start" timeout="20s" interval="0s" id="vip-start-0s"/> \
          <op name="stop" timeout="20s" interval="0s" id="vip-stop-0s"/> \
        </operations> \
    </primitive>
    ````
    The above means for those old existing CIB which time value didn't have time unit, crmsh still tries to add the time unit on it, which causes CIB inconsistency
## Root cause
crmsh will compare parsed cib(auto add time unit 's') with original cib(without time unit 's'), if not equal, it will show the XML directly
## Fix
This commit renames the variable to reflect that it handles both adding advised operation values and time units, but only when the input is from a new(see above reproduce step 1) CLI command. Don't add time units to values for existing CIB.
```
# crm configure show vip
primitive vip IPaddr2 \
        params ip=10.10.10.123 \
        op monitor interval=10 timeout=20s \
        op stop timeout=20s interval=0s
```

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -855,7 +855,8 @@ def parse_cli_to_xml(cli, oldnode=None):
     output: XML, obj_type, obj_id
     """
     node = None
-    advised_op_values = False
+    # Flag to auto add adviced operation values and time units
+    auto_add = False
     default_promotable_meta = False
     comments = []
     if isinstance(cli, str):
@@ -865,12 +866,12 @@ def parse_cli_to_xml(cli, oldnode=None):
     else:  # should be a pre-tokenized list
         utils.auto_convert_role = True
         if len(cli) >= 3 and cli[0] == "primitive" and cli[2].startswith("@"):
-            advised_op_values = False
+            auto_add = False
             default_promotable_meta = False
         else:
-            advised_op_values = config.core.add_advised_op_values
+            auto_add = config.core.add_advised_op_values
             default_promotable_meta = True
-        node = parse.parse(cli, comments=comments, ignore_empty=False, add_advised_op_values=advised_op_values)
+        node = parse.parse(cli, comments=comments, ignore_empty=False, auto_add=auto_add)
     if node is False:
         return None, None, None
     elif node is None:
--- a/crmsh/parse.py
+++ b/crmsh/parse.py
@@ -170,13 +170,13 @@ class BaseParser(object):
         self.begin(cmd, min_args=min_args)
         return self.match_dispatch(errmsg="Unknown command")
 
-    def do_parse(self, cmd, ignore_empty, add_advised_op_values):
+    def do_parse(self, cmd, ignore_empty, auto_add):
         """
         Called by CliParser. Calls parse()
         Parsers should pass their return value through this method.
         """
         self.ignore_empty = ignore_empty
-        self.add_advised_op_values = add_advised_op_values
+        self.auto_add = auto_add
         out = self.parse(cmd)
         if self.has_tokens():
             self.err("Unknown arguments: " + ' '.join(self._cmd[self._currtok:]))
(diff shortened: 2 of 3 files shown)


nicholasyang2022:
Please refine your pull request description. From it I cannot get why parsing cli parameters affects the behavior of `crm configure show`.

nicholasyang2022 (commented):
This implementation is hard to understand. It depends on `if isinstance(cli, str)` at cibconfig.py:865 to call `parse(..., auto_add=False)`.

```
> /usr/lib/python3.11/site-packages/crmsh/cibconfig.py(865)parse_cli_to_xml()
-> node = parse.parse(s, comments=comments)
(Pdb) list
860  	    default_promotable_meta = False
861  	    comments = []
862  	    if isinstance(cli, str):
863  	        utils.auto_convert_role = False
864  	        for s in lines2cli(cli):
865  ->	            node = parse.parse(s, comments=comments)
866  	    else:  # should be a pre-tokenized list
867  	        utils.auto_convert_role = True
868  	        if len(cli) >= 3 and cli[0] == "primitive" and cli[2].startswith("@"):
869  	            auto_add = False
870  	            default_promotable_meta = False
```

When calling `parse_cli_to_xml`, there are not any explicit parameters to tell the difference between: "Oh, we need to add advised operations and convert 600 to 600s as this is a user input." and "Oh, we should not do these as this is loaded from CIB". What `parse_cli_to_xml` will do depends on whether the cli input is tokenized. This too implicit and hard to understand and unmaintainable.

liangxin1300:
> parse_cli_to_xml

liangxin1300:
> This implementation is hard to understand. It depends on `if isinstance(cli, str)` at cibconfig.py:865 to call `parse(..., auto_add=False)`.
> 
> ```
> > /usr/lib/python3.11/site-packages/crmsh/cibconfig.py(865)parse_cli_to_xml()
> -> node = parse.parse(s, comments=comments)
> (Pdb) list
> 860  	    default_promotable_meta = False
> 861  	    comments = []
> 862  	    if isinstance(cli, str):
> 863  	        utils.auto_convert_role = False
> 864  	        for s in lines2cli(cli):
> 865  ->	            node = parse.parse(s, comments=comments)
> 866  	    else:  # should be a pre-tokenized list
> 867  	        utils.auto_convert_role = True
> 868  	        if len(cli) >= 3 and cli[0] == "primitive" and cli[2].startswith("@"):
> 869  	            auto_add = False
> 870  	            default_promotable_meta = False
> ```
> 
> When calling `parse_cli_to_xml`, there are not any explicit parameters to tell the difference between: "Oh, we need to add advised operations and convert 600 to 600s as this is a user input." and "Oh, we should not do these as this is loaded from CIB". What `parse_cli_to_xml` will do depends on whether the cli input is tokenized. This too implicit and hard to understand and unmaintainable.

Good point. Thanks!
But I don't have much time to do the refactoring in recent. I opened a new issue for tracking #1517
And append a behave test case to make sure we don't introduce regression for future refactoring

If the customer does have positive feedback, I think we can move on this time

nicholasyang2022 (approved):

PR #1513, opened 2024-08-12
[crmsh-4.5] Fix: utils: group check failure when os.getgroups() returns empty (bsc#1229030)

liangxin1300:
As written in getgroups(2):
It is unspecified whether the effective group ID of the calling process
is included in the returned list. (Thus, an application should also call
getegid(2) and add or remove the resulting value.)
    
Also use numeric 'haclient' gid instead of text group name, as it
performs less access to the group database

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -21,6 +21,7 @@ import ipaddress
 import argparse
 import random
 import string
+import pwd
 import grp
 from pathlib import Path
 from contextlib import contextmanager, closing
@@ -532,12 +533,10 @@ def chown(path, user, group):
     if isinstance(user, int):
         uid = user
     else:
-        import pwd
         uid = pwd.getpwnam(user).pw_uid
     if isinstance(group, int):
         gid = group
     else:
-        import grp
         gid = grp.getgrnam(group).gr_gid
     try:
         os.chown(path, uid, gid)
@@ -3523,7 +3522,7 @@ def in_haclient():
     """
     Check if current user is in haclient group
     """
-    return constants.HA_GROUP in [grp.getgrgid(g).gr_name for g in os.getgroups()]
+    return grp.getgrnam(constants.HA_GROUP).gr_gid in (os.getgroups() + [os.getegid()])
 
 
 def check_user_access(level_name):
--- a/test/unittests/test_utils.py
+++ b/test/unittests/test_utils.py
@@ -1864,15 +1864,16 @@ def test_has_sudo_access(mock_run):
     mock_run.assert_called_once_with("sudo -S -k -n id -u")
 
 
-@mock.patch('grp.getgrgid')
+@mock.patch('grp.getgrnam')
+@mock.patch('os.getegid')
 @mock.patch('os.getgroups')
-def test_in_haclient(mock_group, mock_getgrgid):
-    mock_group.return_value = [90, 100]
-    mock_getgrgid_inst1 = mock.Mock(gr_name=constants.HA_GROUP)
-    mock_getgrgid_inst2 = mock.Mock(gr_name="other")
-    mock_getgrgid.side_effect = [mock_getgrgid_inst1, mock_getgrgid_inst2]
+def test_in_haclient(mock_getgroups, mock_getegid, mock_getgrnam):
+    mock_getgroups.return_value = [90]
+    mock_getegid.return_value = 90
+    mock_getgrnam_inst = mock.Mock(gr_gid=90)
+    mock_getgrnam.return_value = mock_getgrnam_inst
     assert utils.in_haclient() is True
-    mock_group.assert_called_once_with()
+    mock_getgrnam.assert_called_once_with(constants.HA_GROUP)
 
 
 @mock.patch('crmsh.utils.in_haclient')


nicholasyang2022:
So, this is checking static user group instead of effective gid. Is this safe? What is the root cause for the problem?

liangxin1300:
> So, this is checking static user group instead of effective gid. Is this safe? What is the root cause for the problem?

An alert shell script running `/usr/sbin/crm --force node fence` command under hacluster, `[grp.getgrgid(g).gr_name for g in os.getgroups()]` is []

nicholasyang2022:
> An alert shell script running /usr/sbin/crm --force node fence command under hacluster, [grp.getgrgid(g).gr_name for g in os.getgroups()] is []

So this script is running in a context with dropped privilege, having only uid=hacluster, but not gid=haclient. Adding the gid back does not make scene.

`crm node fence <node>` calls `stonith_admin -F <node>`. The root problem is what privilege `stonith_admin -F <node>` needs. Does it requires `gid=haclient`? Or does it just need `uid=hacluster or gid=haclient`?

If the actual privilege needed is `uid=hacluster or gid=haclient`, I think the correct fix is to change `in_haclient` to `is_hacluster_or_in_haclient`.

liangxin1300:
`crm node fence <node>` will call `crm_attribute -t status -N '<node>' -n terminate -v true`
which requires gid=haclient
```
15sp5-1:/opt # id user1
uid=1001(user1) gid=90(haclient) groups=90(haclient)
15sp5-1:/opt # su - user1 
user1@15sp5-1:~> /usr/sbin/crm_attribute -t status -N '15sp5-2' -n terminate -v true                                                                                                               
user1@15sp5-1:~> echo $?
0

15sp5-1:/opt # id user2
uid=1002(user2) gid=100(users) groups=100(users)
15sp5-1:/opt # su - user2
user2@15sp5-1:~> /usr/sbin/crm_attribute -t status -N '15sp5-2' -n terminate -v true
crm_attribute: Could not connect to the CIB: Transport endpoint is not connected
user2@15sp5-1:~> echo $?
102
```

From the debug info that the alert script given, the running user is hacluster, group is haclient, but os.getgroups() return [], so we should change another way to check the user group

nicholasyang2022:
```
15sp5-1:/opt # id user2
uid=1002(user2) gid=100(users) groups=100(users)
```

You should check a process with uid=hacluster and groups!=haclient. And again, runtime credentials and privileges are associated with a particular process instead of a user. You should use `id` without any arguments to check the privileges hold by the current process.

liangxin1300:
> ```
> 15sp5-1:/opt # id user2
> uid=1002(user2) gid=100(users) groups=100(users)
> ```
> 
> You should check a process with uid=hacluster and groups!=haclient. And again, runtime credentials and privileges are associated with a particular process instead of a user. You should use `id` without any arguments to check the privileges hold by the current process.

Changed by using `id` command output

nicholasyang2022:
I added some debug logging in `in_haclient()` and got these output:

```text
id: uid=90(hacluster) gid=90(haclient) groups=90(haclient)
os.getegid(): 90
os.getgroups(): []
```

As written in `getgroups(2)`:

> It is unspecified whether the effective group ID of the calling process is included in the returned list.  (Thus, an application should also call getegid(2)  and  add  or  remove  the  resulting value.)

So, the fix should be:

```diff
-    return constants.HA_GROUP in [grp.getgrgid(g).gr_name for g in os.getgroups()]
+    return constants.HA_GROUP in [grp.getgrgid(g).gr_name for g in os.getgroups() + [os.getegid()]]
```

crmsh/utils.py
```diff
@@ -3523,7 +3522,7 @@ def in_haclient():
     """
     Check if current user is in haclient group
     """
-    return constants.HA_GROUP in [grp.getgrgid(g).gr_name for g in os.getgroups()]
+    return constants.HA_GROUP_NUM in list(set(os.getgroups() + [os.getegid()]))
```
nicholasyang2022, line 3525: 1. Do not assume the group id of `hacluster` is a constant.
2. No need to convert between set and list.

```suggestion
    return grp.getgrnam(constants.HA_GROUP).gr_gid in (os.getgroups() + [os.getegid()])
```

crmsh/constants.py
```diff
@@ -519,6 +519,7 @@
 LVMLOCKD_RA = "ocf::heartbeat:lvmlockd"
 HA_USER = "hacluster"
 HA_GROUP = "haclient"
+HA_GROUP_NUM = 90
```
nicholasyang2022, line 522:
```suggestion
```

nicholasyang2022 (changes requested):

crmsh/utils.py
liangxin1300: > 1. Do not assume the group id of `hacluster` is a constant.
> 2. No need to convert between set and list.

Still need set to remove duplication

nicholasyang2022: We only need to check whether the gid of hacluster is in the list. Duplication in a list does not make a difference in the search result.

nicholasyang2022 (approved):

PR #1515, opened 2024-08-14
[crmsh-4.6] Fix: bootstrap: ssh public key should be copied to qnetd node when ssh-agent feature is not enabled (bsc#1228950)

nicholasyang2022:
# Problems

When running `crm cluster init qdevice --qnetd-hostame <qnetd-node>` with an environ `SSH_AUTH_SOCK`, even if `--enable-ssh-agent` is not specified, crmsh will use that ssh-agent for `check_ssh_passwd_need`. If the ssh-agent provides a key enabling passwordless ssh authentication to the `<qnetd-node>`, the ssh public key of the `<init-node>` will not be added to the `authorized_keys` of `<qnetd-node>`.

This makes the ssh authentication between `<init-node>` and `<qnetd-node>` to depend on ssh-agent. It is unexpected and causes problems. For example, when a new node joins without ssh-agent, the new node needs the help of `<init-node>` to copy its ssh public key to `<qnetd-node>` (by running `crm cluster init qnetd-remote <new-node>` on `<init-node>`). This will fail as the `<init-node>` cannot get ssh access to `<qnetd-node>` without ssh-agent.

# Fixes

When ssh-agent support is not enabled, drop environ `SSH_AUTH_SOCK` before checking whether passwordless ssh is available.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -885,14 +885,15 @@ def init_ssh_impl(local_user: str, ssh_public_keys: typing.List[ssh_key.Key], us
                 logger.info("Adding public keys to authorized_keys on %s@%s", user, node)
                 for key in ssh_public_keys:
                     authorized_key_manager.add(node, local_user, key)
-                if user != 'root' and 0 != shell.subprocess_run_without_input(
-                        node, user, 'sudo true',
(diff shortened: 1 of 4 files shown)


liangxin1300 (approved):
LGTM, thanks!

Please port to master

PR #1518, opened 2024-08-23
[crmsh-4.6] Fix: report: crm report will hang if CIB contains invalid configuraions (bsc#1229686)

liangxin1300:
## Problem
Command `crm report` will hang if CIB contains invalid configurations, like lack of `stonith-enabled=false` when there is no stonith device configured.
## Root cause
When running `crm report`, crmsh tries to collect the output of `crm_verify`, by calling `get_stdout_or_raise_error`.
`crm_verify` will return non 0 value, then `get_stdout_or_raise_error` will raise custom exception `CommandFailure`, which will be wrongly pickled between processes.
## Solution
1. Implementing the `__reduce__` method to provide custom pickling
    support. This method returns a tuple specifying how to reconstruct
    instances of CommandFailure during the unpickling process, ensuring all
    necessary arguments (`cmd`, `host`, `user`, `msg`) are correctly passed
    to the `__init__` method
2. When there is invalid configuration in cib as crm_verify complains, the
    error output should be recorded in the report result instead of just
    raising an exception

### DIFF
--- a/crmsh/report/collect.py
+++ b/crmsh/report/collect.py
@@ -375,8 +375,9 @@ def consume_cib_in_workdir(workdir: str) -> None:
         crmutils.str2file(out, os.path.join(workdir, constants.CONFIGURE_SHOW_F))
 
         cmd = f"crm_verify -V -x {cib_in_workdir}"
-        out = cluster_shell_inst.get_stdout_or_raise_error(cmd)
-        crmutils.str2file(out, os.path.join(workdir, constants.CRM_VERIFY_F))
+        _, _, err = cluster_shell_inst.get_rc_stdout_stderr_without_input(None, cmd)
+        if err:
+            crmutils.str2file(err, os.path.join(workdir, constants.CRM_VERIFY_F))
 
 
 def collect_config(context: core.Context) -> None:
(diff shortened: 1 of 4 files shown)


nicholasyang2022:
1. Why does unpickling failure lead to hangs?
2. Please add unit tests for pickling `CommandFailure`.

liangxin1300:
> 1. Why does unpickling failure lead to hangs?

Here is an example:
```
import multiprocessing
import pickle

class CustomException(Exception):
    def __init__(self, cmd, host, user, msg):
        super().__init__(f"Command: {cmd}, Host: {host}, User: {user}, Message: {msg}")
        self.cmd = cmd
        self.host = host
        self.user = user
        self.msg = msg

def child_task():
    raise CustomException("ls", "localhost", "root", "Permission denied")

def main():
    pool = multiprocessing.Pool(1)
    result = pool.apply_async(child_task)
    try:
        result.get()
    except Exception as e:
        print(f"Caught exception: {e}")
    finally:
        pool.close()
        pool.join()

if __name__ == "__main__":
    main()
```
I can't exactly tell the reasons, just guess the parent process can't recognize such unpickling failure?

> 2. Please add unit tests for pickling `CommandFailure`.

Done

nicholasyang2022 (approved):

nicholasyang2022:
> I can't exactly tell the reasons, just guess the parent process can't recognize such unpickling failure?

Looks like a bug in `multiprocessing`. We should reproduce it in the latest cpython release and report to upstream if it still exists.

PR #1522, opened 2024-08-26
[crmsh-4.6] Fix: hawk2 cannot call `crm script` as user hacluster (bsc#1228271)

nicholasyang2022:
After the non-root feature is introduced, most of the subprocess calling rountines, including local ones and remote ones, try to switch to root before calling the specified command. This is good for command line usages. However, hawk2 also calls `crm script`, and as a webservice with less privileges, runs as user hacluster and should not escalate to root.

This pull request uses the interceptor interface provided by `prun` module to modify its behaviors. When `crm script` runs as hacluster, subprocesses will also be created as hacluster.

Also fixes a minor problem which stops script `hahealth` from parsing outputs correctly. The problem is caused by module `completers`, which try to call `crm_mon` before setting up the `PATH` environment variable.

### DIFF
--- a/crmsh/completers.py
+++ b/crmsh/completers.py
@@ -70,8 +70,8 @@ def primitives(args):
 
 
 nodes = call(xmlutil.listnodes)
-online_nodes = call(xmlutil.CrmMonXmlParser().get_node_list, "online")
-standby_nodes = call(xmlutil.CrmMonXmlParser().get_node_list, "standby")
+online_nodes = call(lambda x: xmlutil.CrmMonXmlParser().get_node_list(x), "online")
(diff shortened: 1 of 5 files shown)


liangxin1300:
> Also fixes a minor problem which stops script `hahealth` from parsing outputs correctly. The problem is caused by module `completers`, which try to call `crm_mon` before setting up the `PATH` environment variable.

Yeah, that's the issue #1525 
Then PR #1527 and #1526 can be closed

liangxin1300 (approved):
Now run `crm script run health` under hacluster works for me

Thanks!

PR #1531, opened 2024-08-29
[crmsh-4.6] Fix: Add a new option 'has_fa_advised_op' (bsc#1228858)

liangxin1300:
This option is used to automatically add the advised operation values
for fencing agents. Default value is 'no', which means doesn't add the
advised operation values for fencing agents.

Other changes:
- Dev: Add a new internal flag to control auto add time units on operation
- Dev: Rename the option 'add_advised_op_values' to 'has_ra_advised_op'

### DIFF
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -857,6 +857,8 @@ def parse_cli_to_xml(cli, oldnode=None):
     node = None
     # Flag to auto add adviced operation values and time units
     auto_add = False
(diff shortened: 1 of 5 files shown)


zzhou1 (approved):

nicholasyang2022 (commented):
Config/variable names like `ra_advised_op_values` and `fa_advised_op_values` do not sound like boolean type.

PR #1534, opened 2024-08-30
Dev: ui_node: Improve command 'clearstate <node>'

liangxin1300:
Log an error and return False if the node is not found in CIB, avoid IndexError for subsequent code


nicholasyang2022 (approved):
LGTM

PR #1536, opened 2024-08-30
Fix: utils: Stop providing the detailed and precise sudoer rules to "hack" the privilege (bsc#1229093)

liangxin1300:
As the title, to avoid misleading the user for the security risk, crmsh just raises the general error message. That says, let the user figure out the best practice according to the security policy of their organization.


zzhou1 (approved):
LGTM

Issue #1499, opened 2024-07-30
Allow `crm config show` without a running cluster


zzhou1:
Even without a running cluster,  it could give a bit more convenience to user to run `crm config show` directly with an expected WARNING message, something like:

tumbleweed:~ # crm config show
WARNING: no cluster running, this implies `CIB_file=/var/lib/pacemaker/cib/cib.xml crm config show`

liangxin1300:
Not only `crm config show`, but also for other read only commands

liangxin1300:
```
# crm configure show
WARNING: Cluster is not running, loading CIB_file from environment variable
node 1: alp-1
node 2: alp-2
property cib-bootstrap-options: \
        have-watchdog=false \
        dc-version="2.1.7+20240711.239cba384-1.1-2.1.7+20240711.239cba384" \
        cluster-infrastructure=corosync \
        cluster-name=hacluster \
        stonith-enabled=false
rsc_defaults build-resource-defaults: \
        resource-stickiness=1 \
        migration-threshold=3 \
        priority=1
op_defaults op-options: \
        timeout=600 \
        record-pending=true
```

```
# crm configure get_property stonith-enabled 
WARNING: Cluster is not running, loading CIB_file from environment variable
false
```

PR #1540, opened 2024-09-03
Dev: utils: Load CIB_file env before some readonly commands

liangxin1300:
When cluster is not running, load CIB_file environment variable before some readonly commands to avoid the error message. (issue #1499 )

Current affected commands:
- crm configure show
- crm configure get_property
- crm configure verify
- crm node show
- crm node server

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -3171,4 +3171,15 @@ def ssh_command():
     if config.core.no_ssh:
         raise NoSSHError(constants.NO_SSH_ERROR_MSG)
     return "ssh"
+
+
+def load_cib_file_env():
+    if options.regression_tests or ServiceManager().service_is_active("pacemaker.service"):
+        return
+    cib_file = os.environ.setdefault('CIB_file', constants.CIB_RAW_FILE)
+    logger.warning("Cluster is not running, loading the CIB file from %s", cib_file)
+    if not os.path.exists(cib_file):
+        raise ValueError(f"Cannot find cib file: {cib_file}")
+
+
 # vim:ts=4:sw=4:et:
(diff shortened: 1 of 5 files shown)


crmsh/utils.py
```diff
@@ -3171,4 +3171,15 @@ def ssh_command():
     if config.core.no_ssh:
         raise NoSSHError(constants.NO_SSH_ERROR_MSG)
     return "ssh"
+
+
+def load_cib_file_env():
+    if options.regression_tests or ServiceManager().service_is_active("pacemaker.service"):
+        return
+    logger.warning("Cluster is not running, loading CIB_file from environment variable")
+    cib_file = os.environ.setdefault('CIB_file', constants.CIB_RAW_FILE)
```
zzhou1, line 3180: Not 100% sure, keep in mind the existing use case:
`CIB_file="/path/to/cib.xml" crm configure show`

I wonder the safer code should be like this 

```suggestion
    cib_file = os.getenv('CIB_file')
    if cib_file is None:
        cib_file = os.environ.setdefault('CIB_file', constants.CIB_RAW_FILE)
    logger.warning(f"Cluster is not running, loading the CIB file from {cib_file}")
```

zzhou1 (changes requested):

crmsh/utils.py
liangxin1300: https://docs.python.org/3/library/stdtypes.html#dict.setdefault
> If key is in the dictionary, return its value. If not, insert key with a value of default and return default.

So, it's OK for the existing one, just check if the file exists

zzhou1: Great! More clean code then! Even so, I still suggest

```suggestion
    cib_file = os.environ.setdefault('CIB_file', constants.CIB_RAW_FILE)
    logger.warning(f"Cluster is not running, loading the CIB file from {cib_file}")
```

zzhou1 (approved):

PR #1471, opened 2024-06-24
Dev: corosync: add subcommands 'crm corosync link ...' for managing multi-links in knet (jsc#PED-8083)

nicholasyang2022:
Knet transport in corosync provides flexible configurations of multi-link connectivity. This pull request adds some subcommands to manage these configurations, including:

1. `crm corosync link show` to list configured links
2. `crm corosync link update <linknumber> [<nodename>=<new_address> ... ] [options <option>=[value] ...]` to modify an existing link
3. `crm corosync link add <nodename>=<new_address> ... [options <option>=[value] ...]` to add a new link.
4. `crm corosync link remove <linknumber>` to remove an existing link.

PR #1542, opened 2024-09-04
Dev: help: support multi level subcommands

nicholasyang2022:
#1471 adds 3rd level subcommands to crmsh. However, the `help` module is hardcoded to uses 2 levels of subcommands. This pull request modifies the `help` module to support arbitrary levels of subcommands.

It consists of 2 parts:

1. Replace `_`, the subcommand level separators used in `crm.8.adoc`, with `.`, to avoid ambiguity with `_` used in subcommand names (such as `geo_init`).
2. Use tree structure to store help data in memory, so that it works with arbitrary levels of subcommands.

### DIFF
--- a/crmsh/help.py
+++ b/crmsh/help.py
@@ -26,6 +26,7 @@ Help for the level itself is like this:
 '''
 import dataclasses
 import functools
+import io
 import os
 import re
 import typing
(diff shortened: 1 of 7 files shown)


crmsh/help.py
```diff
@@ -129,6 +125,48 @@ def __repr__(self):
...
+            self,
+            short_help: str,
+            cmd_args: typing.Sequence[str],
+            alias_for=None, generated=False,
+    ):
+        super().__init__(short_help, '', alias_for, generated)
+        self._cmd_args = cmd_args
+
+    @functools.cached_property
+    def long(self):
```
liangxin1300, line 139: rename as `long_help` might look more descriptive

crmsh/help.py
```diff
@@ -129,6 +125,48 @@ def __repr__(self):
...
+        args = ['crm']
+        args.extend(self._cmd_args)
+        args.append('--help-without-redirect')
+        rc, stdout = ShellUtils.get_stdout(args, shell=False)
```
liangxin1300, line 143: use `get_stdout_stderr` should be better for storing error msg

crmsh/help.py
```diff
@@ -129,6 +125,48 @@ def __repr__(self):
         return str(self)
 
 
+class LazyHelpEntryFromCli(HelpEntry):
```
liangxin1300, line 128: doc string needed

crmsh/help.py
```diff
@@ -129,6 +125,48 @@ def __repr__(self):
...
+
+
+@dataclasses.dataclass
+class SubcommandTreeNode:
+    name: str
+    help: HelpEntry
+    children: dict[str, typing.Self]
+
+    def is_leaf(self) -> bool:
+        return bool(self.children)
```
liangxin1300, line 157: Typically, a leaf node is a node with, or without children?

crmsh/help.py
```diff
@@ -129,6 +125,48 @@ def __repr__(self):
...
+        args.append('--help-without-redirect')
+        rc, stdout = ShellUtils.get_stdout(args, shell=False)
+        return stdout
+
+    def paginate(self):
+        page_string('{}\n\n{}'.format(self.short, self.long))
+
+
+@dataclasses.dataclass
+class SubcommandTreeNode:
```
liangxin1300, line 151: docstring needed for this new class

crmsh/help.py
```diff
@@ -163,37 +204,34 @@ def get_max_width(dict_):
     return max_width
 
 
+def _render_commmad_tree(out: io.StringIO, node: SubcommandTreeNode, indent: int):
```
liangxin1300, line 207: typo here `commmad`

crmsh/help.py
nicholasyang2022: I removed this method as it is not used. In recursions, it is more straightforward to loop over `children` directly without checking whether it is a leaf node.

liangxin1300 (approved):
Great work! Thanks!

PR #1555, opened 2024-09-18
[crmsh-4.5] Dev: ui_cluster: add subcommand 'crm cluster health hawk2 [--fix]' (bsc#1228899)

nicholasyang2022:
for checking and fixing paswordless ssh authentication for user hacluster.

### DIFF
--- a/.github/workflows/crmsh-ci.yml
+++ b/.github/workflows/crmsh-ci.yml
@@ -357,6 +357,22 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
         flags: integration
 
+  functional_test_healthcheck:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 40
+    steps:
+    - uses: actions/checkout@v4
+    - name: functional test for healthcheck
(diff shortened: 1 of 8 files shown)


liangxin1300:
For that cluster never configured passwordless for root, run command will get the error:
```
# crm cluster health hawk2 --fix
ERROR: cluster.health: Can not create ssh session from 15sp5-1 to 15sp5-2.
```

nicholasyang2022:
> For that cluster never configured passwordless for root, run command will get the error:
> 
> ```
> # crm cluster health hawk2 --fix
> ERROR: cluster.health: Can not create ssh session from 15sp5-1 to 15sp5-2.
> ```

Now, it is changed to

```
suse@ha-1-1:~> sudo crm cluster health hawk2 --fix
INFO: Passwordless ssh is not initialized. Use `crm cluster init ssh` and `crm cluster join ssh -c <init-node>` to set it up.
ERROR: Cannot fix automatically.
```

liangxin1300 (approved):

PR #1561, opened 2024-09-24
Minor improve for cluster health

liangxin1300:


crmsh/healthcheck.py
```diff
@@ -187,7 +187,7 @@ def check_local(self, nodes: typing.Iterable[str]) -> bool:
             return False
 
     def fix_local(self, nodes: typing.Iterable[str], ask: typing.Callable[[str], None]) -> None:
-        logger.warning('Passwordless ssh is not initialized. Use `crm cluster init ssh` and `crm cluster join ssh -c <init-node>` to set it up.')
+        logger.warning('Passwordless ssh is not initialized. Use `crm cluster init ssh` (on init node) and `crm cluster join ssh -c <init-node>` (on join node) to set it up.')
```
nicholasyang2022, line 190: Maybe `on one node` and `on other nodes` is better, as  any one of the nodes can be the init node.

nicholasyang2022 (approved):

Issue #1551, opened 2024-09-10
Issue when ICMP blocked but SSH is allowed


liangxin1300:
```
# crm cluster join -c 15sp6-1 -y
WARNING: chronyd.service is not configured to start at system boot.
ERROR: cluster.join: host "15sp6-1" is unreachable:
```
```
# crm cluster stop --all
WARNING: host "15sp6-1" is unreachable: 
INFO: The cluster stack stopped on 15sp6-2
```
```
# crm report
15sp6-1: ERROR: host "15sp6-2" is unreachable: 

15sp6-1: INFO: The report is saved in ./crm_report-Tue-10-Sep-2024.tar.gz
15sp6-1: INFO: Report timespan: 2024-09-10 09:29:16 - 2024-09-10 21:29:16
15sp6-1: INFO: Including nodes: 15sp6-1
15sp6-1: INFO: Thank you for taking time to create this report
```

All above cases will work when I comment out `utils.ping_node`

PR #1563, opened 2024-09-24
Dev: utils: Check node is reachable by using both ping and ssh

liangxin1300:
If both ping and ssh fail, the node is considered unreachable.

To fix issue #1551

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -26,6 +26,7 @@ import gzip
 import bz2
 import lzma
 import json
+import socket
 from pathlib import Path
 from contextlib import contextmanager, closing
 from stat import S_ISBLK
@@ -2108,15 +2109,21 @@ def check_ssh_passwd_need(local_user, remote_user, host, shell: sh.LocalShell = 
     return rc != 0
 
 
-def check_port_open(ip, port):
-    import socket
-
-    family = socket.AF_INET6 if IP.is_ipv6(ip) else socket.AF_INET
-    with closing(socket.socket(family, socket.SOCK_STREAM)) as sock:
-        if sock.connect_ex((ip, port)) == 0:
-            return True
-        else:
-            return False
+def check_port_open(host, port, timeout=3) -> bool:
+    """
+    Check whether the port is open on the host
+    Use getaddrinfo to support both IPv4 and IPv6
+    """
+    try:
+        addrinfo = socket.getaddrinfo(host, port, socket.AF_UNSPEC, socket.SOCK_STREAM)
+        af, socktype, proto, canonname, sa = addrinfo[0]
+        with closing(socket.socket(af, socktype, proto)) as sock:
+            sock.settimeout(timeout)
+            if sock.connect_ex(sa) == 0:
+                return True
+        return False
+    except socket.error:
+        return False
 
 
 def valid_port(port):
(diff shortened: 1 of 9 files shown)


crmsh/utils.py
```diff
@@ -2460,13 +2461,25 @@ def package_is_installed(pkg, remote_addr=None):
...
-    rc, _, err = ShellUtils().get_stdout_stderr("ping -c 1 {}".format(node))
-    if rc != 0:
-        raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+    rc, _, _ = ShellUtils().get_stdout_stderr(f"ping -c {ping_count} {node}")
+    if rc == 0:
+        return True
+    # ping failed, try to connect to ssh port by socket
+    error_msg = f"host \"{node}\" is unreachable"
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
```
nicholasyang2022, line 2474: Does not work with IPv6. Please use `getaddrinfo` first.

crmsh/utils.py
```diff
@@ -2460,13 +2461,25 @@ def package_is_installed(pkg, remote_addr=None):
...
-def ping_node(node):
+def node_reachable_checking(node, ping_count=1, socket_port=22, socket_timeout=3):
     """
-    Check if the remote node is reachable
+    Check if node is reachable by using ping and socket to ssh port
     """
-    rc, _, err = ShellUtils().get_stdout_stderr("ping -c 1 {}".format(node))
-    if rc != 0:
-        raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+    rc, _, _ = ShellUtils().get_stdout_stderr(f"ping -c {ping_count} {node}")
```
nicholasyang2022, line 2468: Please add `-n` to avoid triggering a reverse DNS lookup.

crmsh/utils.py
liangxin1300: I found we already have `utils.check_port_open`, should be reused this time

crmsh/utils.py
```diff
@@ -2460,13 +2467,18 @@ def package_is_installed(pkg, remote_addr=None):
...
-def ping_node(node):
+def node_reachable_checking(node, ping_count=1, port=22, timeout=3):
     """
-    Check if the remote node is reachable
+    Check if node is reachable by using ping and socket to ssh port
     """
-    rc, _, err = ShellUtils().get_stdout_stderr("ping -c 1 {}".format(node))
-    if rc != 0:
-        raise ValueError("host \"{}\" is unreachable: {}".format(node, err))
+    rc, _, _ = ShellUtils().get_stdout_stderr(f"ping -n -c {ping_count} {node}")
```
nicholasyang2022, line 2474: `timeout` is not applied to the ping command.

crmsh/utils.py
```diff
@@ -2460,13 +2467,18 @@ def package_is_installed(pkg, remote_addr=None):
     return rc == 0
 
 
-def ping_node(node):
+def node_reachable_checking(node, ping_count=1, port=22, timeout=3):
```
nicholasyang2022, line 2470: "check" is a noun.

```suggestion
def node_reachable_check(node, ping_count=1, port=22, timeout=3):
```

nicholasyang2022 (approved):

PR #1564, opened 2024-09-24
Fix: report: find_shell should accept hacluster user (bsc#1228899)

nicholasyang2022:
1. Fix `crm.report.sh.Shell.subprocess_run_without_input`. It is expected to run commands as a specified user on a specified host. The user and host is carried in `Shell` instance, so that callers do not need to specify one when calling method `subprocess_run_without_input`.
2. Modify `crm.report.sh.Shell. _try_create_report_shell` to make it work for 2 different senerios:
    1. When the specified user is `root` or `hacluster`, returns a  `crmsh.sh.SSHShell` wrapped in an adaptor `crm.report.sh.SSHShell`, so that commands runs directly as the specified user.
    2. When the specified user is not `root` or `hacluster`, try to run `sudo` with the `SSHShell`. If
        1. sudo works, returns a `crmsh.report.sh.SSHSudoShell`, which prepends a `sudo` to the command and runs it with `crmsh.SSHShell`.
        2. sudo does not work, return None, failing to create a usable instance of `crm.report.sh.ShellShell`.

### DIFF
--- a/crmsh/report/core.py
+++ b/crmsh/report/core.py
@@ -11,7 +11,6 @@ import sys
 import shutil
 import json
 import ast
-import typing
 from inspect import getmembers, isfunction
 from io import StringIO
 from typing import List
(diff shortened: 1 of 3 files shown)


liangxin1300:
Hi Nicholas,

Please add more PR descriptions to explain the background or context

Thanks!

nicholasyang2022:
> Hi Nicholas,
> 
> Please add more PR descriptions to explain the background or context
> 
> Thanks!

It is updated. Please check if the descriptions are clear enough.

liangxin1300 (approved):

PR #1570, opened 2024-09-29
Dev: scripts: Filter out unreachable nodes in cluster health

liangxin1300:
So that the cluster health check can be continued even if some nodes are unreachable.


nicholasyang2022:
How does this affect other scripts? Do they work as expected if some unreachable nodes are filtered?

liangxin1300:
> How does this affect other scripts? Do they work as expected if some unreachable nodes are filtered?

Don't clear yet, but in the future we might drop the whole scripts part

liangxin1300:
> How does this affect other scripts? Do they work as expected if some unreachable nodes are filtered?

Since no user uses the scripts in these years, I have no exact answer for this, here are my points:
- I think should be no harm to filter out the unreachable nodes
- The whole scripts part(crmsh-scripts package) might be droppd in the future(SLE16 SP1+)

@zzhou1 what do you think?

crmsh/ui_cluster.py
```diff
@@ -849,12 +854,31 @@ def do_health(self, context, *args):
...
+                return False
+            logger.info("Dumping analysis from the report:")
+            try:
+                with tarfile.open(tar_candidates[0], 'r:*') as tar:
+                    analysis_member = tar.getmember("report/analysis.txt")
+                    with tar.extractfile(analysis_member) as analysis_file:
+                        for line in analysis_file:
+                            print(line.decode('utf-8'), end='')
+            except (tarfile.TarError, FileNotFoundError, KeyError) as e:
+                logger.error("Failed to extract analysis from the report: %s", e)
```
nicholasyang2022: Log the exception type and backtrace into `crmsh.log`:

```suggestion
                logger.error("Failed to extract analysis from the report: %s", e, exc_info=True)
```

Issue #1465, opened 2024-06-20
After rename cluster's name, the cluster lose connection with qnetd server


liangxin1300:
```
crm(live/hanode1)# corosync status qnetd
1 hanode1 member
2 hanode2 member

Cluster "hacluster":
    Algorithm:          Fifty-Fifty split (KAP Tie-breaker)
    Tie-breaker:        Node with lowest node ID
    Node ID 1:
        Client address:         2001:db8:10::3:57288
        HB interval:            8000ms
        Configured node list:   1, 2
        Ring ID:                1.e
        Membership node list:   1, 2
        Heuristics:             Undefined (membership: Undefined, regular: Undefined)
        TLS active:             Yes (client certificate verified)
        Vote:                   ACK (ACK)
    Node ID 2:
        Client address:         2001:db8:10::2:56814
        HB interval:            8000ms
        Configured node list:   1, 2
        Ring ID:                1.e
        Membership node list:   1, 2
        Heuristics:             Undefined (membership: Undefined, regular: Undefined)
        TLS active:             Yes (client certificate verified)
        Vote:                   No change (ACK)

crm(live/hanode1)# cluster rename newcluster
INFO: cluster.rename: Copy cluster config file to "hanode2"
INFO: hanode2
INFO: cluster.rename: To apply the change, restart the cluster service at convenient time
crm(live/hanode1)# cluster restart --all
INFO: The cluster stack stopped on hanode1
INFO: The cluster stack stopped on hanode2
WARNING: Qdevice's vote is 0, which simply means Qdevice can't talk to Qnetd(qnetd-node) for various reasons.
INFO: The cluster stack started on hanode1
INFO: The cluster stack started on hanode2
crm(live/hanode1)# corosync status qnetd
1 hanode1 member
2 hanode2 member

None
```

PR #1573, opened 2024-10-08
Fix: ui_cluster: Stop renaming cluster name when using qdevice

liangxin1300:
Give user a hint that the cluster name should not be renamed when using qdevice, should remove qdevice first.

And also automatically restart cluster when there is no any resource running, when renaming the cluster name.

Fix issue #1465


nicholasyang2022 (approved):
LGTM

PR #1384, opened 2024-04-10
Setup bootstrap stages dependency

liangxin1300:
## Changes
- Stages dependency on init side
```
15sp5-1:~ # crm cluster init cluster -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
ERROR: cluster.init: Please run 'ssh' stage first

15sp5-1:~ # crm cluster init ssh -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: A new ssh keypair is generated for user root.
INFO: A new ssh keypair is generated for user hacluster.
INFO: Done (log saved to /var/log/crmsh/crmsh.log)

15sp5-1:~ # crm cluster init cluster -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
ERROR: cluster.init: Please run 'csync2' stage first

15sp5-1:~ # crm cluster init csync2 -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Configuring csync2
INFO: Starting csync2.socket service on 15sp5-1
INFO: BEGIN csync2 checking files
INFO: END csync2 checking files
INFO: Done (log saved to /var/log/crmsh/crmsh.log)

15sp5-1:~ # crm cluster init cluster -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
ERROR: File "/etc/corosync/corosync.conf" not exist
ERROR: cluster.init: Please run 'corosync' stage first

15sp5-1:~ # crm cluster init corosync -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Configuring corosync(knet)
INFO: Done (log saved to /var/log/crmsh/crmsh.log)

15sp5-1:~ # crm cluster init cluster -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Hawk cluster interface is now running. To see cluster status, open:
INFO:   https://192.168.122.29:7630/
INFO: Log in with username 'hacluster', password 'linux'
[...]

PR #1575, opened 2024-10-09
[crmsh-4.5] Dev: bootstrap: more robust implementation for ssh_merge (bsc#1230530)

nicholasyang2022:
* Do not use legacy `write_remote_file`, which does not work well with `_guess_user_for_ssh`.
* Implement deduplication when merging known_hosts

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -26,15 +26,13 @@ import typing
 
 import yaml
 import socket
-from tempfile import mktemp
 from string import Template
 from lxml import etree
 
-import crmsh.parallax
 from . import config, constants
 from . import utils
 from . import xmlutil
-from .cibconfig import mkset_obj, cib_factory
+from .cibconfig import cib_factory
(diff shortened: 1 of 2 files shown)


liangxin1300:
> ## How to reproduce
> We need to emulate a cluster with key-based ssh authentication set up in advance, and ensure there are at least 2 existing nodes in the cluster, so that `ssh_merge` has some host keys to merge.
> 
> 1. on node1: `crm cluster init -N node2 -y`
> 2. on node3:
>    
>    1. `ssh-keygen`
>    2. `ssh-copy-id node2`
>    3. `crm cluster init ssh_merge -c node2 -y`

I can reproduce following above steps.
btw, the third step should be `crm cluster join ssh_merge -c node2 -y`

liangxin1300:
As the user still using stages to setup cluster, I suggest we backport https://github.com/ClusterLabs/crmsh/pull/1384 to crmsh-4.5 and crmsh-4.4

Issue #1587, opened 2024-10-21
Failed to re-init cluster on interactive mode


liangxin1300:
Reproduce steps:
1. On the first node, run `crm cluster init -y`
2. On the second node, run `crm cluster join -c <first_node> -y`
3. On the first node, run `crm cluster stop --all`
4. On the first node, run `crm cluster init`
```
# crm cluster init
...
/etc/corosync/corosync.conf already exists - overwrite (y/n)? n
...
INFO: BEGIN Waiting for cluster
...........                                                                                                                                                                                        
INFO: END Waiting for cluster
ERROR: cluster.init: Joined existing cluster - will not reconfigure.
```
The first node is online, but without configuring cluster property like `stonith-enabled`

PR #1588, opened 2024-10-21
Fix: bootstrap: Remove the unnecessary check

liangxin1300:
Fix issue #1587

At this point, the local node must already online, otherwise the previous init_cluster_local will raise exceptions. And this is not related to the join process.


nicholasyang2022:
I think the original behavior is correct. You do not overwrite corosync.conf, and you will not get a new cluster.

PR #1589, opened 2024-10-21
Fix: hahealth: return fail when report failed (bsc#1231840)

aleksei-burlakov:
This will help external applications (e.g. hawk) better understand the result of the 'crm script run health' execution.

### DIFF
--- a/scripts/health/hahealth.py
+++ b/scripts/health/hahealth.py
@@ -22,7 +22,7 @@ def create_report():
 
 
 if not create_report():
-    crm.exit_ok({'status': 'Failed to create report'})
+    crm.exit_fail({'status': 'Failed to create report'})
 
 
 def extract_report():
@@ -31,7 +31,7 @@ def extract_report():
 
 
 if not extract_report():
-    crm.exit_ok({'status': 'Failed to extract report'})
+    crm.exit_fail({'status': 'Failed to extract report'})
 
 analysis = ''
 if os.path.isfile('health-report/analysis.txt'):


scripts/health/hahealth.py
```diff
@@ -22,7 +22,7 @@ def create_report():
 
 
 if not create_report():
-    crm.exit_ok({'status': 'Failed to create report'})
+    crm.exit_fail({'status': 'Failed to create report'})
```
nicholasyang2022, line 25: [`exit_fail`](https://github.com/ClusterLabs/crmsh/blob/76fb6a0be7c121e971907d98af7e4b3bbc6f9548/utils/crm_script.py#L60) does not accept a dictonary.

```suggestion
    crm.exit_fail('Failed to create report')
```

nicholasyang2022 (commented):
And should we also change [this line](https://github.com/ClusterLabs/crmsh/blob/6fa784dcb26be957e02118b7b14303c3f31a5398/scripts/health/hahealth.py#L34)?

scripts/health/hahealth.py
aleksei-burlakov: > And should we also change [this line](https://github.com/ClusterLabs/crmsh/blob/6fa784dcb26be957e02118b7b14303c3f31a5398/scripts/health/hahealth.py#L34)?

That's right.

> [exit_fail](https://github.com/ClusterLabs/crmsh/blob/76fb6a0be7c121e971907d98af7e4b3bbc6f9548/utils/crm_script.py#L60) does not accept a dictionary.

Actually it does. I think, it was in the older versions of python, you couldn't pass a dictionary to a `print` function. Now it's a valid code `print({'status': 'Failed to create report'})`

nicholasyang2022 (approved):

nicholasyang2022:
@liangxin1300 We have test failures.

PR #1601, opened 2024-11-03
Fix: Python 3.12: SyntaxWarning: invalid escape sequence

vvidic:


nicholasyang2022 (approved):
It's great! Thank you.

And we should setup a linter to check similar problems.

PR #1604, opened 2024-11-05
Dev: github-actions: rebuild the test container image weekly

nicholasyang2022:
so that we will catch up with the updates from tumbleweed


liangxin1300 (approved):
Nice!

PR #1605, opened 2024-11-05
[crmsh-4.6] Fix: report.utils: Fix the performance issue (bsc#1232821)

liangxin1300:
The function `extract_critical_log` can be very slow when processing large log files. The main issue is the inefficiency of the regular expression, which combines multiple wildcards (.*) with alternation (|), leading to excessive backtracking. Additionally, the function reads the entire file into memory, which is not optimal for large files.

To improve performance, it is better to use `grep` with the -F option to search for fixed strings.

And for a sequence of archived log files, check the modify time. No need to check the rest of the files if the from time is greater than the modify time of the file.

### DIFF
--- a/crmsh/report/constants.py
+++ b/crmsh/report/constants.py
@@ -3,7 +3,7 @@
 BIN_CRM = "/usr/sbin/crm"
 BIN_COLLECTOR = f"{BIN_CRM} report __collector"
 COMPRESS_DATA_FLAG = "COMPRESS CRM_REPORT DATA:::"
-LOG_PATTERNS = "CRIT: ERROR: error: warning: crit:"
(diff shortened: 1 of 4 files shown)


crmsh/report/constants.py
```diff
@@ -3,7 +3,7 @@
 BIN_CRM = "/usr/sbin/crm"
 BIN_COLLECTOR = f"{BIN_CRM} report __collector"
 COMPRESS_DATA_FLAG = "COMPRESS CRM_REPORT DATA:::"
-LOG_PATTERNS = "CRIT: ERROR: error: warning: crit:"
+LOG_PATTERNS = "CRIT: ERROR: WARNING: crit: error: warning:"
```
nicholasyang2022, line 6: Instead of defining these patterns as space-separated string, we can define it directly as a list of strings.

nicholasyang2022 (approved):

PR #1608, opened 2024-11-11
Fix: log: The report DEBUG log message is not displayed in the log file

liangxin1300:
Combine "filter_report" and "filter" into one filter

Fix issue #1607


nicholasyang2022 (approved):
LGTM

PR #1616, opened 2024-11-22
Fix: behave: systemd bpf-restrict-fs fails in opensuse tumbleweed

nicholasyang2022:
Use apparmor=unconfined when the distribution does not provide a proper profile.


zzhou1 (approved):
Works in my TW environment. Thanks!

liangxin1300 (approved):

PR #1600, opened 2024-11-01
Dev: utils: Avoid hardcoding the ssh key type as RSA

liangxin1300:
## Problem
- The join process will fail if the existing key type is not RSA. See: https://github.com/ClusterLabs/crmsh/issues/1504#issuecomment-2443330157
- If the init and join nodes are already set up for passwordless access using an ed25519 key, an RSA key pair is still generated during the init/join process.

## Changes include:
- Avoid hardcoding the ssh key type as RSA
- Introduced a new function `ssh_key.fetch_public_key_list` to fetch public keys from local or remote, return as public key path list or public key content list
- In KeyFileManager, use class variable to store the key type instead of hardcoding it as RSA
- Improve shell script in generate_ssh_key_pair_on_remote to avoid hardcoding the key type as RSA
- Remove unused code

PR #1617, opened 2024-11-25
[crmsh-4.6] Dev: utils: Avoid hardcoding the ssh key type as RSA

liangxin1300:
backport #1600

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1037,10 +1037,8 @@ def change_user_shell(user, remote=None):
 
 def configure_ssh_key(user):
     """
-    Configure ssh rsa key on local or remote
-
-    If <home_dir>/.ssh/id_rsa not exist, generate a new one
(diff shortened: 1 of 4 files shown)


nicholasyang2022:
I think we should not add this new feature to 4.6.

liangxin1300:
> I think we should not add this new feature to 4.6.

Do you think this is a feature or issue fix? this PR doesn't change the default key as the RSA
@zzhou1

zzhou1:
> > I think we should not add this new feature to 4.6.
> 
> Do you think this is a feature or issue fix? this PR doesn't change the default key as the RSA @zzhou1

I think I agree with @nicholasyang2022 as a master branch feature to adapt to the latest Tumbleweed. I don't see the reasonable justification in 4.6 as a bug fix.

liangxin1300:
Let's reopen this when this is needed

PR #1620, opened 2024-11-26
Fix: report: Fix a regression for irregular file collection

liangxin1300:
Single irregular file like crm.conf and profiles.yml are not collected, which was introduced by commit 210661b3.

Should determine if the file does have archived logs first.


nicholasyang2022 (approved):
LGTM

PR #1622, opened 2024-11-27
Collect ~/.config/crm/crm.conf in crm report result

liangxin1300:
Changes include:
1. The file `~/.config/crm/crm.conf` is now collected in crm report result

2. To manage files that have the same basename (e.g., `crm.conf`), the PR implements a new directory. When multiple files share the same basename, a directory named `{basename}.d` is created in the report result. All files with the same basename are then placed inside this directory. For instance, the files `/etc/crm/crm.conf` and `/root/.config/crm/crm.conf` will be placed inside a directory named `crm.conf.d`

### DIFF
--- a/crmsh/report/collect.py
+++ b/crmsh/report/collect.py
@@ -96,28 +96,6 @@ def collect_journal_logs(context: core.Context) -> None:
         logger.debug(f"Dump jounal log for {item} into {utils.real_path(_file)}")
 
 
-def dump_D_process() -> str:
-    """
-    Dump D-state process stack
-    """
-    out_string = ""
-
-    sh_utils_inst = ShellUtils()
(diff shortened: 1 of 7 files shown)


crmsh/report/collect.py
```diff
@@ -65,9 +65,15 @@ def collect_ha_logs(context: core.Context) -> None:
     Collect pacemaker, corosync and extra logs
     """
     log_list = [get_pcmk_log(), get_corosync_log()] + context.extra_log_list
-    for log in log_list:
+    log_list = [
+        os.path.expanduser(log)
+        if log.startswith("~") else log
```
nicholasyang2022, line 70: No need to test `startwith("~")`. `os.path.expanduser()` handles paths without leading `~` correctly.

nicholasyang2022 (approved):

PR #1624, opened 2024-12-05
Dev: ui_resource: Refine 'do_failcount' function

liangxin1300:
Changes include:
- Enable set failcount for resource operation when there is no failcount
      entry
- Improve readability of code
- Add functional test cases to increase coverage

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -2,6 +2,7 @@
 # Copyright (C) 2013-2018 Kristoffer Gronlund <kgronlund@suse.com>
 # See COPYING for license information.
 
+import re
 from . import command
 from . import completers as compl
 from . import constants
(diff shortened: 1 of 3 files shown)


crmsh/constants.py
```diff
@@ -44,6 +44,7 @@
                 "notify",
                 "reload")
 ra_operations = tuple(["probe"] + list(op_cli_names))
+ATTR_OPERATION_CMDS = ('delete', 'set', 'show')
```
nicholasyang2022, line 47: This is only used by subcommand `resource`. It should not be put here.

crmsh/ui_resource.py
```diff
@@ -491,20 +492,23 @@ def do_failcount(self, context, rsc, cmd, node, value=None, operation=None, inte
         failcount <rsc> delete <node>
         failcount <rsc> show <node>"""
         def sec_to_ms(s):
-            return s + '000'
+            return '0' if s == '0' else f'{s}000'
```
nicholasyang2022, line 495: `s + '000'` will be more efficient.

crmsh/ui_resource.py
```diff
@@ -491,20 +492,23 @@ def do_failcount(self, context, rsc, cmd, node, value=None, operation=None, inte
...
 
         def ms_to_sec(m):
-            return m[:len(m)-3]
+            return '0' if m == '0' else f'{int(m) // 1000}'
```
nicholasyang2022, line 498: `str(int(m) // 1000)` will be more efficient.

nicholasyang2022 (approved):

PR #1639, opened 2024-12-22
Fix: bootstrap: Local joining node should be included when merging known_hosts (bsc#1229419)

liangxin1300:
Otherwise, the local joining node's known_hosts entry will not be synced to the other nodes.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1748,7 +1748,7 @@ def join_ssh_merge(cluster_node, remote_user):
     """
     logger.info("Merging known_hosts")
 
-    hosts = _context.node_list_in_cluster
+    hosts = _context.node_list_in_cluster + [utils.this_node()]
 
     shell = sh.cluster_shell()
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -1757,7 +1757,6 @@ def join_ssh_merge(cluster_node, remote_user):
     known_hosts_new: set[str] = set()
 
     cat_cmd = "[ -e ~/.ssh/known_hosts ] && cat ~/.ssh/known_hosts || true"
-    #logger_utils.log_only_to_file("parallax.call {} : {}".format(hosts, cat_cmd))
```
zzhou1, line 1760: This is the chance to make lines unique before saving `known_hosts_new` to the disk. Currently, it easily gets duplicate lines.

liangxin1300: Please try again
I think this `duplicate` issue already gone

nicholasyang2022 (approved):

PR #1643, opened 2024-12-25
Dev: help: Support '--help' option for cluster properties

liangxin1300:
By using `crm configure property <property_name> --help`, print the help of property on non-interactive mode.

Also, for `crm configure property` command, print all properties if no property is specified


zzhou1 (approved):
LGTM

PR #1660, opened 2025-01-09
Make minor changes based on Pacemaker-3.0's changelog (jsc#PED-8231)

liangxin1300:
Changed:
- Dev: bootstrap: Drop 'record-pending' operation option
- Dev: utils: Make change since 'crmd-transition-delay' already renamed to 'transition-delay'
- Dev: constants: Drop deprecated "crmd-integration-timeout" and "crmd-finalization-timeout"
- Dev: crash_test: Drop deprecated 'poweroff' value of 'stonith-action' option
- Dev: Remove deprecated cibadmin --local option
- Dev: Drop unsupported 'moon' in date_spec
- Dev: constants: Drop deprecated 'restart-type' resource option
- Dev: Improve options of 'crm status' command
    
    - Reduce duplicated options definition
    - Drop deprecated '-s/--simple-status' option

### DIFF
--- a/crmsh/xmlutil.py
+++ b/crmsh/xmlutil.py
@@ -84,7 +84,7 @@ def compressed_file_to_cib(s):
     return cib_elem
 
 
-cib_dump = "cibadmin -Ql"
+cib_dump = "cibadmin -Q"
 
 
 def sudocall(cmd):
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1458,7 +1458,7 @@ def init_cluster():
     logger.info("Loading initial cluster configuration")
 
     crm_configure_load("update", """property cib-bootstrap-options: stonith-enabled=false
-op_defaults op-options: timeout=600 record-pending=true
+op_defaults op-options: timeout=600
 rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
 
--- a/crmsh/cmd_status.py
+++ b/crmsh/cmd_status.py
@@ -41,6 +41,22 @@ _ERRORS = ['not running',
            'Unknown',
            'OFFLINE',
            'Failed actions']
+CRM_MON_OPTIONS_MAP = {
+        "bynode": "-n",
+        "inactive": "-r",
+        "ops": "-o",
+        "timing": "-t",
+        "failcounts": "-f",
+        "verbose": "-V",
+        "quiet": "-Q",
+        "html": "--output-as html",
+        "xml": "--output-as xml",
+        "tickets": "-c",
+        "noheaders": "-D",
+        "detail": "-R",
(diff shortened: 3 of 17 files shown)


zzhou1:
Do you want this PR scope in "Drop support for name with id-ref attribute" ?

liangxin1300:
> Do you want this PR scope in
> 
> * ["Drop support for name with id-ref attribute"](https://github.com/ClusterLabs/pacemaker/commit/c08aa7) aka. "Attribute reference"

Done

liangxin1300:
Thinking: Considering the user might have a chance to `downgrade` the schema, is this PR suitable to drop these codes?

crmsh/xmlutil.py
```diff
@@ -1450,14 +1450,12 @@ def nvpair_id(nvpairid, name, value):
     return new("nvpair", id=nvpairid, name=name, value=value)
 
 
-def nvpair_ref(idref, name=None):
+def nvpair_ref(idref):
     """
-    <nvpair id-ref=<idref> [name=<name>]/>
+    <nvpair id-ref=<idref>/>
```
gao-yan, line 1455: Indeed pacemaker-3.0 dropped the support for nvpair name with id-ref from the XML schema "pacemaker-4.0". But in case it's existing when an old version syntax of CIB is used, pacemaker interprets it internally.

It makes sense to drop the usage from the manpage/helptext. But crmsh should better still be able to show it in case it is existing.

gao-yan:
> Thinking: Considering the user might have a chance to `downgrade` the schema, is this PR suitable to drop these codes?

No need to consider `downgrade` since it's not supported. But we still should be able to support users' existing CIBs i.e. syntax based on older versions of "validate-with"/schemas that are still shipped by pacemaker.

> 
> Or, make a schema-version-checker to drop these functionalities only when using pacemaker-4.0+?

Taking the example of "nvpair name with id-ref", we usually have the considerations as mentioned above. Of course it'd be even better if it's possible to achieve our considerations without requiring version check I think.

liangxin1300:
> No need to consider downgrade since it's not supported

Hi Yan,
If we don't consider downgrade, please also take a look at this PR https://github.com/ClusterLabs/crmsh/pull/1677
That PR provides a completer for all schema versions, for the `configure schema` sub-command, from pacemaker-1.0 to the latest version
My question, how about we don't complete the older version, just complete the newer version, comparing the current schema version we are using? If crmsh is using the latest one, then complete nothing

crmsh/xmlutil.py
liangxin1300: Then I keep that code unchanged, otherwise it wouldn't parse the old syntax
But I drop the syntax mention of '@\<id\>:name' in crm.8.adoc, in `help AttributeReferences`

PR #1668, opened 2025-01-17
Dev: report: Collect pre-migration check info in crm report result

liangxin1300:


crmsh/report/collect.py
```diff
@@ -546,3 +546,12 @@ def collect_qdevice_info(context: core.Context) -> None:
...
     crmutils.str2file(out_string, _file)
     logger.debug(f"Dump quorum/qdevice/qnetd information into {utils.real_path(_file)}")
+
+
+def collect_pre_migration_info(context: core.Context) -> None:
+    cmd = "crm cluster health sles16"
+    out_string = f"#### Run command: {cmd} ####\n"
+    out_string += utils.get_cmd_output(cmd)
+    _file = os.path.join(context.work_dir, constants.PRE_MIGRATION_F)
+    crmutils.str2file(out_string, _file)
```
nicholasyang2022, line 556: Why not just redirect the both stdout and stderr to the file?

liangxin1300: get_cmd_output does include both

nicholasyang2022: But it adds overheads by reading the output into memory, perform unicode decoding and encoding, and writing back to the file.

liangxin1300: Changed

nicholasyang2022 (approved):

PR #1673, opened 2025-01-24
Dev: cibverify: Print output of crm_verify directly

liangxin1300:
To avoid the log level difference between pacemaker and crmsh


nicholasyang2022 (approved):
LGTM

PR #1679, opened 2025-01-27
Dev: ra: Prevent to add unknown operation

liangxin1300:
- When adding unknown operation, sanity check should give an error instead of a warning
- And sanity check will return a big number to indicate this is a fatal error, so that the commit process will be aborted


crmsh/ui_configure.py
```diff
@@ -744,6 +744,8 @@ def _verify(self, set_obj_semantic, set_obj_all):
         rc1 = set_obj_all.verify()
         if config.core.check_frequency != "never":
             rc2 = set_obj_semantic.semantic_check(set_obj_all)
+            if rc2 >= constants.SANITY_FATAL_RC:
+                cib_factory.reset()
```
nicholasyang2022, line 748: I think `reset()` should not be called here.

1. `_verify()` is also called from `do_verify()`. This subcommand should not have a side effect.
2. Even when called from `_commit()`, it is enough to abort the commit, so that users will have a chance to modify their config instead of losing all the changes.

liangxin1300: Changed as raising `utils.TerminateSubCommand`
- non-interactive case:
```
alp-1:~ # crm configure primitive stateful-1 ocf:pacemaker:Stateful op monitor_Slave interval=10s op monitor_Master interval=5s
WARNING: (unpack_config)        warning: Blind faith: not fencing unseen nodes
ERROR: Action 'monitor_Slave' not found in Resource Agent meta-data
ERROR: Action 'monitor_Master' not found in Resource Agent meta-data
alp-1:~ # crm configure show stateful-1
ERROR: object stateful-1 does not exist
```
- interactive case:
```
alp-1:~ # crm
crm(live/alp-1)# configure
crm(live/alp-1)configure# primitive stateful-1 ocf:pacemaker:Stateful op monitor_Slave interval=10s op monitor_Master interval=5s
crm(live/alp-1)configure# primitive d2 Dummy

# terminate the commit, but the temp changes are still there
crm(live/alp-1)configure# commit
WARNING: (unpack_config)        warning: Blind faith: not fencing unseen nodes
ERROR: Action 'monitor_Slave' not found in Resource Agent meta-data
ERROR: Action 'monitor_Master' not found in Resource Agent meta-data
crm(live/alp-1)configure# show stateful-1
        op monitor_Slave interval=10s \
        op monitor_Master interval=5s
crm(live/alp-1)configure# show d2
primitive d2 Dummy
```

nicholasyang2022: Raising TerminateSubcommand looks like a hack. Why not just return False in `_verify()`? Does this unknown operation verification has any difference with other verifications?

nicholasyang2022: So, it will be better to use different return values in `_verify()` to represent fatal and non-fatal errors, and add some comments to note why this is a fatal error.

nicholasyang2022 (commented):
1. It is invalid to use bitwise-or operation with a series defined as 0, 1, 2, 3, ....
2. Please use enum instead instead of standalone integers.

nicholasyang2022 (approved):

PR #1688, opened 2025-02-12
Dev: testcases: Adjust original test cases to match pacemaker-3

liangxin1300:
Major changes:
- Remove stonith:null and stonith:ssh in original test cases
- Set `stonith-enabled=false` to avoid config error

PR #1689, opened 2025-02-14
Fix: ci: test container image build failure in tumbleweed

nicholasyang2022:
* No need to install libopenssl manually any more as bsc#1195792 is fixed.
* libglue-devel is gone.
* python3* pacakges is now named python311*.


liangxin1300 (approved):

liangxin1300:
libglue-devel provides stonith:null and stonith:ssh
Currently, the CI in the main branch failed, it will work after #1688 merged

PR #1709, opened 2025-03-12
Dev: ui_resource: Refactor do_trace function

liangxin1300:
Ensure the trace directory is specified either through the argument or by using the value of the heartbeat_dir option. Eliminate any inconsistencies that might arise from heartbeat_dir (from etc/crm/crm.conf) or from HA_VARLIB (from /usr/lib/ocf/heartbeat/ocf-directories).

### DIFF
--- a/crmsh/ui_resource.py
+++ b/crmsh/ui_resource.py
@@ -708,19 +708,15 @@ class RscMgmt(command.UI):
         rsc = self._get_trace_rsc(rsc_id)
         if not rsc:
             return False
-
-        argl = list(args)
-        force = "force" in utils.fetch_opts(argl, ["force"]) or config.core.force
-        if len(argl) > 3:
+        if len(args) > 3:
             context.fatal_error(usage)
-        op=None
-        interval=None
-        dir=None
-        for arg in argl:
-            if arg[0] == '/':
-                if dir is not None:
+
+        op, interval, trace_dir = None, None, None
+        for arg in args:
+            if arg.startswith("/"):
+                if trace_dir is not None:
                     context.fatal_error(usage)
-                dir = arg
+                trace_dir = arg
             elif arg.isnumeric():
                 if interval is not None:
                     context.fatal_error(usage)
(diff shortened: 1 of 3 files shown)


zzhou1 (approved):

crmsh/ui_resource.py
```diff
@@ -704,6 +704,19 @@ def _trace_op_interval(self, context, rsc_id, rsc, op, interval, dir):
     @command.completers(compl.primitives, _raoperations)
     def do_trace(self, context, rsc_id, *args):
         'usage: trace <rsc> [<op>] [<interval>] [<log-dir>]'
+        def get_ha_varlib_value():
+            """
+            Find the value of HA_VARLIB, which should be defined in ocf-directories
+            """
+            _, out, _ = ShellUtils().get_stdout_stderr("rpm -ql resource-agents")
```
nicholasyang2022, line 711: This will not work when resource-agents are installed from source. We should not depends on package manager here.

nicholasyang2022 (changes requested):

crmsh/ui_resource.py
nicholasyang2022: I think we should enumerate possible libdirs, just like setting up `PATH` in `main.envsetup`.

crmsh/utils.py
```diff
@@ -3209,4 +3209,15 @@ def is_subdict(sub_dict, main_dict):
     Check if sub_dict is a sub-dictionary of main_dict
     """
     return all(main_dict.get(k) == v for k, v in sub_dict.items())
+
+
+def get_variable_value_from_script(script_path, var_name):
+    """
+    source a script and get the value of a variable
+    """
+    cmd = f"source {script_path} && echo ${var_name}"
```
nicholasyang2022, line 3218: **DANGEROUS:**

1. Sourcing a script is **vulnerable to shell injection**.
2. `script_path` and `var_name` is not quoted.

Although it is safe when sourcing trusted code, this should not be defined as a common utility function.

liangxin1300: How about:

```suggestion
    cmd = ['/bin/bash', '-c', f'source {script_path} && echo ${var_name}']
    rc, out, _ = ShellUtils().get_stdout_stderr(cmd, shell=False)
```

zzhou1: I agree Nicholas for the security risk, and not to source the script. Instead, it is safe for the use case to directly use python `re` to parse `{HA_VARLIB:=/var/lib/heartbeat}`.

If "rpm -ql resource-agents" is not available, then crmsh falls back to the hard code of `/usr/lib/ocf/lib/heartbeat/ocf-directories`.

doc/crm.8.adoc
```diff
@@ -2036,7 +2036,8 @@ stop <rsc> [<rsc> ...]
 ==== `trace`
 
 Start tracing RA for the given operation. When `[<log-dir>]`
-is not specified the trace files are stored in `$HA_VARLIB/trace_ra`.
+is not specified, the trace files are stored in `$HA_VARLIB/trace_ra`.
+If $HA_VARLIB is not defined, it defaults to `/var/lib/heartbeat`.
```
nicholasyang2022, line 2040: 1. `HA_VARLIB` is not used in the latest patch.

    ```suggestion
    is not specified, the trace files are stored in `/var/lib/heartbeat/trace_ra`.
    ```

2. config template `etc/crm.conf` needs update:

    ```patch
    -; heartbeat_dir = <detected>
    +; heartbeat_dir = /var/lib/heartbeat
    ```

nicholasyang2022 (approved):

PR #1718, opened 2025-03-19
Fix: idmgmt: Replace hashtag('#') with point('.') in id (bsc#1239782)

liangxin1300:


nicholasyang2022:
Why? It looks like we are trying to bypass cib schema validation.

crmsh/idmgmt.py
```diff
@@ -65,6 +65,7 @@ def new(node, pfx):
             node_id = pfx
     if is_used(node_id):
         node_id = _gen_free_id(node_id)
+    node_id = re.sub(r'#', '.', node_id)
```
nicholasyang2022, line 68: 1. add comments to explain why we need this substitution.
2. `str.replace()` is enough. No need to use regex.

```suggestion
    # The ID type in XML allows only `^[_a-zA-Z][\w\-.]*$`
    # `crm_attribute` command replaces `#` with `.` in this case
    node_id = node_id.replace('#', '.')
```

nicholasyang2022 (approved):

PR #1740, opened 2025-04-11
Dev: sbd: Remove pcmk_delay_max while cacaulating stonith timeout value

### LINKED ISSUE #1739: bootstrap: adapt calculation for pacemaker improvement on stonith-timeout
The calculation  of stonith-timeout in the crmsh bootstrap [1] could remove pcmk_delay_max, since pacemaker now excludes any fencing delay to eat up stonith_timeout [2].

[2] https://github.com/ClusterLabs/pacemaker/pull/3082 
[1] https://github.com/ClusterLabs/crmsh/pull/890

liangxin1300:
fix #1739
- Remove pcmk_delay_max from the formula to calculate stonith-timeout
- Remove pcmk_delay_max from the formula to calculate SBD_DELAY_START

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -305,7 +305,6 @@ class SBDTimeout(object):
         if dev_list:  # disk-based
             self.disk_based = True
             self.msgwait = SBDTimeout.get_sbd_msgwait(dev_list[0])
-            self.pcmk_delay_max = utils.get_pcmk_delay_max(self.two_node_without_qdevice)
         else:  # disk-less
             self.disk_based = False
             self.sbd_watchdog_timeout = SBDTimeout.get_sbd_watchdog_timeout()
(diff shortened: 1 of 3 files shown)


zzhou1 (approved):

crmsh/sbd.py
```diff
@@ -343,12 +342,12 @@ def get_sbd_delay_start_expected(self):
         '''
         Get the value for SBD_DELAY_START, formulas are:
 
-        SBD_DELAY_START = (token + consensus + pcmk_delay_max + msgwait)  # for disk-based sbd
+        SBD_DELAY_START = (token + consensus + msgwait)  # for disk-based sbd
```
gao-yan, line 345: There's a misunderstanding here.

In bsc#1242981, we talked about removing `pcmk_delay_max` from the formula for `stonith-timeout`, but not for `SBD_DELAY_START`.

The technical detail is, with https://github.com/ClusterLabs/pacemaker/pull/3082, pacemaker no longer lets fencing delays consume the *configured* stonith-timeout to prevent any delay from eating up the timeout due to a misconfiguration. Instead, pacemaker internally calculates a *total practical* timeout to be used. That's the reason why `pcmk_delay_max` no longer needs to be considered as a factor when configuring `stonith-timeout`.

OTOH for sbd's `SBD_DELAY_START`, unlike pacemaker itself, sbd technically doesn't have the knowledge about the delay setting in pacemaker especially upon startup. So the formula for `SBD_DELAY_START` still has to take the factor `pcmk_delay_max`.

PR #1692, opened 2025-02-19
Dev: Add error log when cluster services fail to start

liangxin1300:
## Problem 1
crmsh will show success logs when success to start:
```
# crm cluster start --all
INFO: Starting pacemaker.service on alp-1, alp-2
INFO: The cluster stack started on alp-1
INFO: The cluster stack started on alp-2
```
But will show nothing when fail to start:
```
# crm cluster start --all
INFO: Starting pacemaker.service on alp-1, alp-2
```
## Solution
```
# crm cluster start --all
INFO: Starting pacemaker.service on alp-1, alp-2
ERROR: The cluster stack failed to start on alp-1
INFO: The cluster stack started on alp-2
```

## Promblem 2
When the fatal error leads to start pacemaker failing, the bootstrap process will wait for a long time:
```
# crm cluster init -y
...
INFO: Starting pacemaker.service on alp-1
INFO: BEGIN Waiting for cluster
.............................................................                                                                                                                                      
ERROR: FAIL Waiting for cluster
ERROR: cluster.init: Time out waiting for cluster.
```
## Solution
check the return code and quit quickly
```
# crm cluster init -y
...
INFO: Starting pacemaker.service on alp-1
ERROR: cluster.init: Failed to start cluster services
```

PR #1746, opened 2025-04-17
Fix: ui_cluster: Return when cluster service on all nodes are already startd (bsc#1241358)

liangxin1300:
When cluster is already started, the return value of `crm cluster start` is 1, which was expected as 0

```
# crm cluster start
INFO: The cluster stack already started on node1
# echo $?
1
```
This regression was introduced from #1692

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -180,7 +180,7 @@ class Cluster(command.UI):
             logger.info("Please try 'crm cluster start' on each node")
             return False
         if not node_list:
-            return False
+            return
 
         if start_qdevice:
             service_manager.start_service("corosync-qdevice", node_list=node_list)
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -249,7 +249,7 @@ class Cluster(command.UI):
         except utils.NoSSHError as msg:
             logger.error('%s', msg)
             logger.info("Please try 'crm cluster stop' on each node")
-            return
+            return False
         if not node_list:
             return
         logger.debug(f"stop node list: {node_list}")
--- a/test/features/cluster_api.feature
+++ b/test/features/cluster_api.feature
@@ -130,3 +130,13 @@ Feature: Functional test to cover SAP clusterAPI
     And     Directory "hanode2" in "/tmp/report.tar.bz2"
     And     File "pacemaker.log" in "/tmp/report.tar.bz2"
     And     File "corosync.conf" in "/tmp/report.tar.bz2"
+
+  @clean
+  Scenario: crm cluster start should return 0 for successful repeated execution (bsc#1241358)
+    When    Run "crm cluster start" on "hanode1"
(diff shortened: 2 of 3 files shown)


test/features/cluster_api.feature
```diff
@@ -130,3 +130,13 @@ Feature: Functional test to cover SAP clusterAPI
     And     Directory "hanode2" in "/tmp/report.tar.bz2"
     And     File "pacemaker.log" in "/tmp/report.tar.bz2"
     And     File "corosync.conf" in "/tmp/report.tar.bz2"
+
+  @clean
+  Scenario: The return value of crm cluster start changed (bsc#1241358)
```
zzhou1, line 135:
```suggestion
  Scenario: crm cluster start should return 0 for successful repeated execution (bsc#1241358)
```

crmsh/ui_cluster.py
```diff
@@ -178,9 +178,9 @@ def do_start(self, context, *args):
         except utils.NoSSHError as msg:
             logger.error('%s', msg)
             logger.info("Please try 'crm cluster start' on each node")
-            return False
+            return
```
nicholasyang2022, line 181: Why is NoSSHEror also changed?

liangxin1300: It also changed to return False in #1692, which was different on `do_stop` side

Or, this time we define NoSSHError's return code as False, on both start and stop sides?
@zzhou1 What do you think?

zzhou1: It makes sense to me to stay as "1"

```

adm@tw-1:~> sudo crm cluster start --all; echo $?
INFO: The cluster stack already started on tw-1
ERROR: cluster.start: Failed to run command systemctl is-active 'pacemaker.service' on adm@tw-2: adm@tw-2: Permission denied (publickey,password,keyboard-interactive). Environment variable SSH_AUTH_SOCK does not exist. Please check whether ssh-agent is available and consider using "sudo --preserve-env=SSH_AUTH_SOCK".
1

adm@tw-1:~> sudo crm cluster stop --all; echo $?
ERROR: cluster.stop: Failed to run command systemctl is-active 'corosync.service' on adm@tw-2: adm@tw-2: Permission denied (publickey,password,keyboard-interactive). Environment variable SSH_AUTH_SOCK does not exist. Please check whether ssh-agent is available and consider using "sudo --preserve-env=SSH_AUTH_SOCK".
1

```

zzhou1: And, all the following should return "0"

```
adm@tw-1:~> sudo crm cluster stop; echo $?
INFO: The cluster stack stopped on tw-1
0
adm@tw-1:~> sudo crm cluster start; echo $?
INFO: Starting pacemaker.service on tw-1
INFO: The cluster stack started on tw-1
0
adm@tw-1:~> sudo crm cluster start; echo $?
INFO: The cluster stack already started on tw-1
0       # <-- only fix here
```

nicholasyang2022 (approved):

PR #1748, opened 2025-04-18
[crmsh-4.6] Dev: workflows: Update CI image on github action

liangxin1300:
- Dev: workflows: Update CI image on github action
   As ubuntu-20.04 is not provided by github action anymore, we need to update the CI image to ubuntu-24.04.
- Dev: workflows: Run unit test with python container


nicholasyang2022:
For consistency, let's use python container in unit tests for all variants.

zzhou1 (approved):

PR #1751, opened 2025-04-21
Dev: Makefile: Use modern tools python build and pip instead of setup.py

liangxin1300:
to build and install the package.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -57,27 +57,19 @@ install-data-hook:
 
 # Python module installation
 all-local:
-	(cd $(srcdir); $(PYTHON) setup.py build \
-		--build-base $(shell readlink -f $(builddir))/build \
-		--verbose)
+	cd $(srcdir); python3 -m build --outdir $(shell readlink -f $(builddir))/dist --wheel --sdist --verbose
 
 python_prefix = --prefix=$(prefix)
 
 install-exec-local:
 	$(INSTALL) -d -m 770 $(DESTDIR)/${localstatedir}/log/crmsh
-	-mkdir -p $(DESTDIR)$(pkgpythondir)
-	$(PYTHON) $(srcdir)/setup.py install \
-		--root $(DESTDIR)/ \
-		$(python_prefix) \
-		--record $(DESTDIR)$(pkgpythondir)/install_files.txt \
-		--verbose
+	python3 -m pip install --prefix=$(prefix) $(shell readlink -f $(builddir))/dist/*.whl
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -18,27 +18,30 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.
 #
-MAINTAINERCLEANFILES    = Makefile.in aclocal.m4 configure
+# Set a default value for DESTDIR if not provided
+DESTDIR ?= /
```
nicholasyang2022, line 22: https://www.gnu.org/prep/standards/html_node/DESTDIR.html:

> You should not set the value of DESTDIR in your Makefile at all

test/run-functional-tests
```diff
@@ -190,7 +190,7 @@ deploy_ha_node() {
         else
             podman_security=""
         fi
-	make_cmd="cd $PROJECT_INSIDE;./autogen.sh && ./configure --prefix /usr && make install && make install-crmconfDATA prefix= && cp /usr/bin/crm /usr/sbin"
+	make_cmd="cd $PROJECT_INSIDE;./autogen.sh && ./configure --prefix /usr && make all-local && make install && make install-crmconfDATA prefix= && cp /usr/bin/crm /usr/sbin"
```
nicholasyang2022, line 193: If you need to run `make all-lcoal` before `make install`, your target dependencies in Makefile is broken.

Makefile.am
```diff
@@ -55,29 +58,19 @@ install-data-hook:
...
 
-# Python module installation
+# Python module installation using modern tools
 all-local:
-	(cd $(srcdir); $(PYTHON) setup.py build \
-		--build-base $(shell readlink -f $(builddir))/build \
-		--verbose)
-
-python_prefix = --prefix=$(prefix)
+	(cd $(srcdir); python3 -m build --outdir $(shell readlink -f $(builddir))/dist --wheel --sdist --verbose)
```
nicholasyang2022, line 63: As you write change this into a single line, you will not need to use a sub shell any longer.

nicholasyang2022:
I approve to use `python -m build`. But `python -m installer` is low-level and undocumented. We should use pip instead. And the rpm spec has already been using `python -m build` and `pip`.

liangxin1300:
> I approve to use `python -m build`. But `python -m installer` is low-level and undocumented. We should use pip instead. And the rpm spec has already been using `python -m build` and `pip`.

Changed. If it works, please approve #1752 firstly

nicholasyang2022 (approved):

PR #1771, opened 2025-04-29
 Dev: ui_cluster: Skip stopping cluster if dlm_controld is running in maintenance mode

liangxin1300:
- Dev: ui_cluster: Skip stop cluster if cluster is in maintenance mode and dlm is running
```
# crm cluster restart
INFO: The cluster is in maintenance mode
ERROR: Stopping pacemaker/corosync will trigger unexpected node fencing when 'dlm_controld' is running in maintenance mode.
# crm cluster stop
INFO: The cluster is in maintenance mode
ERROR: Stopping pacemaker/corosync will trigger unexpected node fencing when 'dlm_controld' is running in maintenance mode.
```
Dev: ui_cluster: Refactor the `do_restart` function
    
    - Ensure the function returns `False` immediately if `do_stop` returns
      `False`, avoiding the execution of `do_start`.
    - Update the function to return the result of `do_start`, as it may
      return `False`.

### DIFF
--- a/crmsh/ui_cluster.py
+++ b/crmsh/ui_cluster.py
@@ -275,14 +275,17 @@ class Cluster(command.UI):
         for node in node_list:
             logger.info("The cluster stack stopped on {}".format(node))
 
+        return True
+
     @command.skill_level('administrator')
     def do_restart(self, context, *args):
         '''
         Restarts the cluster stack on all nodes or specific node(s)
         '''
-        parse_option_for_nodes(context, *args)
-        self.do_stop(context, *args)
-        self.do_start(context, *args)
+        stop_rc = self.do_stop(context, *args)
+        if stop_rc is False:
+            return False
+        return self.do_start(context, *args)
 
     @command.skill_level('administrator')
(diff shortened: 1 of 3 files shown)


crmsh/ui_cluster.py
```diff
@@ -254,6 +254,11 @@ def do_stop(self, context, *args):
             return
         logger.debug(f"stop node list: {node_list}")
 
+        if utils.is_cluster_in_maintenance_mode() and utils.is_dlm_running():
+            logger.info("The cluster is in maintenance mode")
+            logger.error("Stop pacemaker/corosync will trigger unexpected node fencing when 'dlm_controld' is running in maintenance mode.")
```
zzhou1, line 259:
```suggestion
            logger.error("Stopping pacemaker/corosync will trigger unexpected node fencing when 'dlm_controld' is running in maintenance mode.")
```

zzhou1 (commented):
Minor cosmetic suggestion

zzhou1 (approved):

PR #1775, opened 2025-05-04
Dev: ui_configure: Add ':' suffix to order kind completer

liangxin1300:
Avoid crmsh treating kind of order(Mandatory, Optional, and Serialize) as a resource name.


zzhou1 (approved):
Nice improvement

PR #1777, opened 2025-05-05
Dev: cibconfig: Prevent adding Pacemaker remote resources to groups, orders, or colocations

liangxin1300:


liangxin1300:
- [ ] Should add new behave test case for pacemaker remote

crmsh/cibconfig.py
```diff
@@ -2000,7 +2022,9 @@ def check_sanity(self) -> VerifyResult:
         if self.node is None:
             logger.error("%s: no xml (strange)", self.obj_id)
             return utils.get_check_rc()
-        return _check_if_constraint_ref_is_child(self)
+        rc1 = _check_if_constraint_ref_is_child(self)
+        rc2 = check_if_primitive_in_constraint_is_remote(self)
```
nicholasyang2022: To keep consistent, should we make this function name to start with `_`?

PR #1790, opened 2025-05-15
Fix: Makefile.am: add python wheel target and polish uninstall actions

zzhou1:
- add target to build Python wheel
- fix an error when install tests
- fix uninstall-local target with Python3 pip uninstall


liangxin1300:
Please rebase the master

PR #1792, opened 2025-05-16
Fix: Makefile.am: add python wheel target and .spec Recommends bash-completion

zzhou1:
Makefile.am:
- add target to build Python wheel
- fix an error when install tests
- fix uninstall-local target with Python3 pip uninstall
- install bash-completion

crmsh.spec.in:
- Recommends bash-completion


liangxin1300:
@zzhou1 Please address the conflicts

PR #1801, opened 2025-05-23
[crmsh-4.5] Fix: report: Don't collect cib.txt separately (bsc#1243498)

liangxin1300:
to avoid potential race conditions


liangxin1300:
> What is the race condition?

If cib.xml has not been collected yet, then cib.txt wouldn't be collected

zzhou1 (approved):

PR #1807, opened 2025-05-27
`corosync set` command improvement

liangxin1300:
## Problem
- If an invalid value is set via `corosync edit` or `corosync set`, crmsh does not complain
```
# crm corosync set totem.version 3
# corosync -t
May 27 20:18:05 error   [MAIN  ] parse error in config: This totem parser can only parse version 2 configurations.
May 27 20:18:05 error   [MAIN  ] Corosync Cluster Engine exiting with status 8 at main.c:1456.
```
- The completer of `corosync set` has invalid completion output
```
crm(live/alp-1)corosync# set 
#comment_0                                   
#comment_1
```
- No push note after `corosync set`, which is the same after `corosync edit`
```
crm(live/alp-1)corosync# edit
WARNING: "/etc/corosync/corosync.conf" has changed, should be synced with other nodes
INFO: Use "crm corosync diff" to show the difference
INFO: Use "crm corosync push" to sync
```
## Changed
- Dev: ui_corosync: Call `corosync -t` to do verification
```
# crm corosync set totem.version 3
ERROR: Invalid /etc/corosync/corosync.conf: Failed to run 'corosync -t': May 27 21:01:14 error   [MAIN  ] parse error in config: This totem parser can only parse version 2 configurations.
May 27 21:01:14 error   [MAIN  ] Corosync Cluster Engine exiting with status 8 at main.c:1456.
```
- ui_corosync: Add push reminder after called `corosync set`
```
# crm corosync set totem.token 6000
WARNING: "/etc/corosync/corosync.conf" has changed, should be synced with other nodes
INFO: Use "crm corosync diff" to show the difference
INFO: Use "crm corosync push" to sync
```
- Dev: corosync_config_format: Skip comment key
- Dev: doc: Update help text for `corosync set` command
- Dev: ui_corosync: Write changes made by `corosync.set` to temporary file first

### DIFF
--- a/test/features/corosync_ui.feature
+++ b/test/features/corosync_ui.feature
@@ -48,3 +48,19 @@ Feature: crm corosync ui test cases
     Then    Expected "Duplicated" in stderr
     When    Try "crm corosync link add hanode1=192.0.2.101 hanode2=192.0.2.102 options knet_link_priority=10" on "hanode1"
     Then    Expected "not a configured interface address" in stderr
+
+  Scenario: corosync set
+    Given   Nodes ["hanode1", "hanode2"] are cleaned up
+    And     Cluster service is "stopped" on "hanode1"
+    And     Cluster service is "stopped" on "hanode2"
+    When    Run "crm cluster init -y" on "hanode1"
+    Then    Cluster service is "started" on "hanode1"
+    When    Run "crm cluster join -c hanode1 -y" on "hanode2"
+    Then    Cluster service is "started" on "hanode2"
+    And     Online nodes are "hanode1 hanode2"
+    When    Try "crm corosync set totem.version 3" on "hanode1"
+    Then    Expected "parse error in config" in stderr
+    When    Run "crm corosync get totem.version" on "hanode1"
+    Then    Expected "2" in stdout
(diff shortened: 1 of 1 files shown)


nicholasyang2022:
Why don't you write changes to tmpfile and write to the original file after the tmpfile passes validation?

crmsh/ui_corosync.py
```diff
@@ -412,7 +418,16 @@ def do_get(self, context, path):
     @command.completers(completers.call(corosync.get_all_paths))
     def do_set(self, context, path, value, index: int = 0):
         """Set a corosync configuration value"""
-        corosync.set_value(path, value, index)
+        corosync_conf_file = corosync.conf()
+        with utils.create_tempfile() as temp_file:
+            shutil.copyfile(corosync_conf_file, temp_file)
+            corosync.ConfParser.set_value(path, value, index=index, config_file=temp_file)
+            if corosync.is_valid_corosync_conf(temp_file):
+                shutil.copyfile(temp_file, corosync_conf_file)
```
nicholasyang2022: To achieve atomic replacement, you need to put the temp file in the same directory and use `os.rename` instead of copy.

nicholasyang2022 (approved):
LGTM

PR #1811, opened 2025-06-04
Fix: bootstrap: add sleeps to avoid triggering sshd PerSourcePenalties (bsc#1243141)

nicholasyang2022:


crmsh/bootstrap.py
```diff
@@ -969,8 +962,11 @@ def ssh_copy_id_no_raise(local_user, remote_user, remote_node, shell: sh.LocalSh
     if utils.check_ssh_passwd_need(local_user, remote_user, remote_node, shell):
         configure_ssh_key(local_user)
         public_keys = ssh_key.fetch_public_key_file_list(None, local_user)
+        sleep(5)    # bsc#1243141: sshd PerSourcePenalties
```
liangxin1300, line 965: I'm not clear why sleep 5 seconds can avoid PerSourcePenalties?
Please also write the details into git commit log

nicholasyang2022: I don't know either. To understand it you need to read openssh source code.

liangxin1300 (approved):

crmsh/bootstrap.py
```diff
@@ -964,7 +964,9 @@ def ssh_copy_id_no_raise(local_user, remote_user, remote_node, shell: sh.LocalSh
         public_keys = ssh_key.fetch_public_key_file_list(None, local_user)
         sleep(5)    # bsc#1243141: sshd PerSourcePenalties
```
zzhou1, line 965: NOTE:  5 seconds is a good practical choice. 4 seconds could still possible leads to the penalty during test.

for i in {1..50}; do ssh -o BatchMode=yes adm@server true; sleep 5; done

PR #1812, opened 2025-06-04
Dev: bootstrap: Improve configuration for admin IP

liangxin1300:
## Problem
- If an invalid admin IP is configured (for example, one on a different subnet), the process will wait for a long time
- When a timeout occurs, it will simply exit and will not proceed to configure the qdevice.
```
# crm cluster init -A 40.40.40.123 -y --qnetd-hostname server1
...
INFO: BEGIN Configuring virtual IP (40.40.40.123)
.........................................................    # here we wait for a long time, actually IPaddr2 RA failed quickly                                                                                                   ERROR: FAIL Configuring virtual IP (40.40.40.123)
ERROR: cluster.init: Time out waiting for resource.          # cannot configure qdevice here, just exit
```
## Changes
- Wait for shorter time for admin IP to be started
- Don't exit if admin IP start fails, just log an error

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -474,7 +474,7 @@ def crm_configure_load(action, configuration):
     sh.cluster_shell().get_stdout_or_raise_error(f"crm -F configure load {action} {configuration_tmpfile}")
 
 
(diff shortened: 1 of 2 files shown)


zzhou1 (approved):

zzhou1 (approved):

nicholasyang2022 (approved):
LGTM

PR #1823, opened 2025-06-13
Fix: bootstrap: setup_passwordless_with_other_nodes does not update the authorized_keys on localhost (bsc#1244515)

nicholasyang2022:
1. localhost is missed from the node list when calling `_merge_ssh_authorized_keys`.
2. `_merge_ssh_authorized_keys` should not use parallex, or the keys are merged into the authorized_keys file of root, instead of ssh user.


nicholasyang2022:
No, 4.6 is not affected.

PR #1825, opened 2025-06-16
Dev: bootstrap: Remove pacemaker remote node

liangxin1300:
By using `crm cluster remove <node>`
```
# crm cluster remove pcmk-remote-node1 -y
INFO: Removing node pcmk-remote-node1 from cluster
INFO: Removing pacemaker remote node pcmk-remote-node1 from cluster
INFO: Removing node pcmk-remote-node1 from CIB
INFO: Done (log saved to /var/log/crmsh/crmsh.log on hanode1)
```


nicholasyang2022 (approved):
LGTM

PR #1828, opened 2025-06-17
Fix: bootstrap: Reload corosync after sync corosync.conf (bsc#1244437)

liangxin1300:
Also make minor refactors to the code.


nicholasyang2022 (approved):
LGTM

PR #1832, opened 2025-06-21
Dev: ui_cluster: Enhance membership validation for cluster run and copy commands

liangxin1300:


nicholasyang2022:
> In the `utils.validate_and_get_reachable_nodes` function, if an offline node is detected, run `crm node show` on a that offline node to check whether local node still appears as a cluster member from the offline node's perspective.

Why do you care about the view of the cluster from an offline node? An offline node does not have updated information about the cluster.

nicholasyang2022:
> For the case mentioned in #1832

How exactly does fetching the node list from an offline node help this case? Why don't you fetch from an online node?

> And we can handle other offline cases:
> * cluster services down on other nodes:

It is OK to call systemctl stop on already stopped services. No need to filter.

> * nodes down

It can be filtered by a simple ping test.

PR #1834, opened 2025-06-22
Dev: xmlutil: Address circular import issue

liangxin1300:
This commit is to resolve a circular import issue between xmlutil and utils modules


nicholasyang2022:
What does this fix? It is still a circular import.

liangxin1300:
Already included in #1814

Issue #1817, opened 2025-06-08
pacemake remote node can't standby


liangxin1300:
```
 # crm node standby remote-node
ERROR: node.standby: Node 'remote-node' is not a member of the cluster
```
Actaully, this is allowed:
```
# cibadmin -Q -o nodes
<nodes>
  <node id="1" uname="node-1"/>
  <node id="2" uname="node-2"/>
  <node id="remote-node" uname="remote-node" type="remote">
    <instance_attributes id="nodes-remote-node">
      <nvpair id="nodes-remote-node-standby" name="standby" value="on"/>
    </instance_attributes>
  </node>
</nodes>
```

PR #1855, opened 2025-07-03
Dev: ui_node: Enable standby and maintenance on pacemaker remote node

liangxin1300:
## Problems
- Can't standby pacemaker remote node #1817
- Can't maintenance pacemaker remote node https://github.com/ClusterLabs/crmsh/issues/1817#issuecomment-3026676284
- The way to get remote node's name
```
/cib/status/node_state[@remote_node="true"]/@uname
```
Might have problem:
```
>>> xmlutil.listnodes()
['rm', 'alp-2']
# if the pacemaker remote RA's name changed from `rm` to `rm_node`, then got two remote nodes
>>> xmlutil.listnodes()
['rm', 'rm_node', 'alp-2']
```
## Changes:
-  Drop xmlutil.listnodes function; And refactor xmlutil.CrmMonXmlParser.get_node_list function, get pacemaker remote node from `crm_mon --output-as=xml` looks more accurately
- Enable standby/online pacemaker remote node
- Enable maintenance/ready pacemaker remote node

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2468,7 +2468,7 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
     """
     Check if all cluster nodes are reachable
     """
-    online_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=True, only_member=True, peer=peer_node)
+    online_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=True, node_type="member", peer=peer_node)
     offline_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=False, peer=peer_node)
     dead_nodes = []
     for node in offline_nodes:
@@ -3263,7 +3263,7 @@ def validate_and_get_reachable_nodes(
         fatal("Cannot get the member list of the cluster")
     pcmk_remote_list = []
     if include_remote:
-        pcmk_remote_list = xmlutil.CrmMonXmlParser.get_node_list(online=True, only_remote=True)
+        pcmk_remote_list = xmlutil.CrmMonXmlParser.get_node_list(online=True, node_type="remote")
     for node in nodes_from_args:
         if node not in cluster_member_list and node not in pcmk_remote_list:
             fatal(f"Node '{node}' is not a member of the cluster")
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -2468,8 +2468,9 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
     """
     Check if all cluster nodes are reachable
     """
-    online_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=True, node_type="member", peer=peer_node)
-    offline_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=False, peer=peer_node)
+    crm_mon_inst = xmlutil.CrmMonXmlParser(peer_node)
+    online_nodes = crm_mon_inst.get_node_list(online=True, node_type="member")
+    offline_nodes = crm_mon_inst.get_node_list(online=False)
     dead_nodes = []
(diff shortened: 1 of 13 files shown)


crmsh/xmlutil.py
```diff
@@ -1531,20 +1519,37 @@ def is_node_online(self, node):
...
+            standby: bool = None,
+            only_member: bool = False,
+            only_remote: bool = False,
+            peer: str = None
+        ) -> list[str]:
         """
         Get a list of nodes based on the given attribute
+        Return all nodes if no attributes are given
         """
+        instance = cls(peer)
```
nicholasyang2022: Why do you change this to a classmethod and create new instance in the method? This does not make sense.

crmsh/xmlutil.py
```diff
@@ -1531,20 +1519,37 @@ def is_node_online(self, node):
...
-        standby_value = "true" if standby else "false"
-        conditions.append(f'@standby="{standby_value}"')
-        if exclude_remote:
+
+        if online is not None:
+            online_value = 'true' if online else 'false'
+            conditions.append(f'@online="{online_value}"')
+        if standby is not None:
+            standby_value = 'true' if standby else 'false'
+            conditions.append(f'@standby="{standby_value}"')
```
nicholasyang2022: This is to complex. `online` vs `standby`, and `only_member` vs `only_remote` is mutually exclusive. Please use enum or str instead of bool.

crmsh/utils.py
```diff
@@ -3249,7 +3250,8 @@ def __bool__(self):
 
 def validate_and_get_reachable_nodes(
         nodes: typing.List[str] = [],
-        all_nodes: bool = False
+        all_nodes: bool = False,
```
nicholasyang2022, line 3252: Why not test `all_nodes` at the very beginning in this function? If `all_node` is true, you don't need to do the other complex logic.

crmsh/xmlutil.py
liangxin1300: The get_node_list method is called by some completers. In interactive scenarios, using a classmethod ensures that the returned data is updated each time it is invoked.

crmsh/xmlutil.py
liangxin1300: You can see the outout of `crm_mon --output-as xml`, online/standby/maintenance are not exclusive parameters, I changed only_member and only_remote as one parameter

crmsh/utils.py
liangxin1300: Improved the logic

crmsh/xmlutil.py
nicholasyang2022: You could just create instance when you need to update the data. Keeping it as an instance method provides flexibility to use cached data.

crmsh/utils.py
```diff
@@ -2466,9 +2468,8 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
     """
     Check if all cluster nodes are reachable
     """
-    crm_mon_inst = xmlutil.CrmMonXmlParser(peer_node)
-    online_nodes = crm_mon_inst.get_node_list()
-    offline_nodes = crm_mon_inst.get_node_list(online=False)
+    online_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=True, node_type="member", peer=peer_node)
+    offline_nodes = xmlutil.CrmMonXmlParser.get_node_list(online=False, peer=peer_node)
```
nicholasyang2022: This would provides inconsistent data, e.g. the same node appears in both online_nodes and offline_node, or disappears from both, as the data are fetched from 2 different calls to `crm_mon`.

liangxin1300: Changed back to be an instance method

nicholasyang2022 (approved):

PR #1865, opened 2025-07-17
Fix: sbd: Avoid negative value for the property 'stonith-watchdog-timeout' (bsc#1246622)

liangxin1300:
## Problem
Previously, when configuring diskless SBD with crmsh bootstrap, the option `stonith-watchdog-timeout` was set to -1 to prompt Pacemaker to automatically calculate the appropriate value.

However, Pacemaker now raises a warning when a negative value is used for timeout or duration options such as `stonith-watchdog-timeout`:

```
pacemaker-fenced[...]:  warning: '-1' is not a valid time duration: Negative
pacemaker-controld[...]:  warning: '-1' is not a valid time duration: Negative
```
## Changes
Set the default value of `stonith-watchdog-timeout` to 2*SBD_WATCHDOG_TIMEOUT. The value of SBD_WATCHDOG_TIMEOUT is stored in /etc/sysconfig/sbd and can be kept consistent across all nodes using crmsh.

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -182,7 +182,6 @@ class SBDTimeout(object):
     '''
     Consolidate sbd related timeout methods and constants
     '''
-    STONITH_WATCHDOG_TIMEOUT_DEFAULT = -1
     SBD_WATCHDOG_TIMEOUT_DEFAULT = 5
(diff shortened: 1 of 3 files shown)


gao-yan (approved):
LGTM. Thanks, @liangxin1300 !

PR #1874, opened 2025-07-23
Fix: bootstrap: continue qnetd setup when ssh keypair is not found for some cluster nodes (#1850)

### LINKED ISSUE #1850: bootstrap: the qdevice stage should work with ssh-agent without any local ssh key
Indeed, we can bootstrap the cluster with qdevice all together by using ssh-agent without create any local ssh key on any node. Example as below

```
adm@tw-1:~> sudo -E crm cluster init -y -N adm@tw-2 --qnetd-hostname adm@tw-4
```

However, it doesn't work for the  "qdevice" stage along. Here is the reproducer:

```
adm@tw-1:~> sudo -E crm cluster init -y -N adm@tw-2
adm@tw-1:~> sudo -E crm cluster init qdevice --qnetd-hostname adm@tw-4
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Configure Qdevice/Qnetd:              
ERROR: cluster.init: No public key file found for adm@tw-2
```

The workaround is to create the local ssh key

```
adm@tw-2:~> ssh-keygen -t rsa -N "" -f ~/.ssh/id_rsa
```

nicholasyang2022:
If the cluster was using ssh-agent, keypair would not be generated. And it is OK to skip this step as keys will be provided by ssh-agent.

Close #1850.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1609,9 +1609,16 @@ def _setup_passwordless_ssh_for_qnetd(cluster_node_list: typing.List[str]):
         if node == utils.this_node():
             continue
         local_user, remote_user, node = _select_user_pair_for_ssh_for_secondary_components(node)
-        remote_key_content = ssh_key.fetch_public_key_content_list(node, remote_user)[0]
-        in_memory_key = ssh_key.InMemoryPublicKey(remote_key_content)
-        ssh_key.AuthorizedKeyManager(cluster_shell).add(qnetd_addr, qnetd_user, in_memory_key)
+        try:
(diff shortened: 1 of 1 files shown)


liangxin1300 (approved):
LGTM, thanks!

liangxin1300:
Is this also needed to backport to 4.6?

nicholasyang2022:
> Is this also needed to backport to 4.6?

No. This is related to "enable ssh-agent by default".

PR #1877, opened 2025-07-29
Fix: sbd: should show warning instead of error when unable to restart cluster automatically after changing configs (bsc#1246956)

nicholasyang2022:
When dlm is running, we cannot restart the cluster in maintenance mode as this will trigger node fencing. This is not a fatal error because:

1. The configuration has already been changed successfully.
2. Users may restart the cluster manually.

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -557,12 +557,17 @@ class SBDManager:
     def restart_cluster_if_possible(with_maintenance_mode=False):
         if not ServiceManager().service_is_active(constants.PCMK_SERVICE):
             return
-        if xmlutil.CrmMonXmlParser().is_any_resource_running() and not with_maintenance_mode:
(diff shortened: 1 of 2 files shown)


nicholasyang2022:
Outputs after this change: 

```
suse@ha-2-1:~> sudo crm -F sbd configure watchdog-timeout=51
INFO: No 'msgwait-timeout=' specified in the command, use 2*watchdog timeout: 102
INFO: Set cluster to maintenance mode
WARNING: "maintenance-mode" in crm_config is set to true, it was false
INFO: Configuring disk-based SBD
INFO: Initializing SBD device /dev/loop0
WARNING: Resource is running, need to restart cluster service manually on each node
INFO: Set cluster from maintenance mode to normal
INFO: Delete cluster property "maintenance-mode" in crm_config
```

liangxin1300 (approved):

Issue #1878, opened 2025-07-30
Completion shows help topic but can't run


liangxin1300:
There are `topics` help in crmsh:
```
Help overview for crmsh


Available topics:

        Overview                Help overview for crmsh

        Topics                  Available topics

        Description             Program description
        CommandLine             Command line options
        Introduction            Introduction
        Interface               User interface
        Completion              Tab completion
        Shorthand               Shorthand syntax
        Features                Features
        Shadows                 Shadow CIB usage
        TimeoutFormulas         Timeout formulas
        Checks                  Configuration semantic checks
        Templates               Configuration templates
        Testing                 Resource testing
        Security                Access Control Lists (ACL)
        Resourcesets            Syntax: Resource sets
        AttributeListReferences Syntax: Attribute list references
        AttributeReferences     Syntax: Attribute references
        RuleExpressions         Syntax: Rule expressions
        Lifetime                Lifetime parameter format
        Reference               Command reference
```
At any level of subcommand, these topics can be completed:
```
# crm node help 
AttributeListReferences   Features                  Resourcesets              Testing                   delete                    quit                      up 
AttributeReferences       Interface                 RuleExpressions           TimeoutFormulas           fence                     ready                     utilization 
Checks                    Introduction              Security                  Topics                    help                      server                    
CommandLine               Lifetime                  Shadows                   attribute                 ls                        show                      
Completion                Overview                  Shorthand                 cd                        maintenance               standby                   
Description               Reference                 Templates                 clearstate                online                    status-attr
```
But can't run:
```
# crm node help Features
ERROR: node.help: No help found for "crm node Features". Run "crm help overview" to list all help entries.
```

liangxin1300:
Only failed for master branch

liangxin1300:
#1878 Fix this issue by not showing help topic completion at the sub-level

PR #1879, opened 2025-07-30
Dev: command: Show help topic completion only at root level

liangxin1300:
Fix issue [Completion shows help topic but can't run](https://github.com/ClusterLabs/crmsh/issues/1878)

### DIFF
--- a/crmsh/command.py
+++ b/crmsh/command.py
@@ -231,8 +231,10 @@ def _cd_completer(args, context):
 
 
 def _help_completer(args, context):
-    'TODO: make better completion'
-    return help_module.list_help_topics() + context.current_level().get_completions()
+    current_level_completions = context.current_level().get_completions()
+    if context.current_level().name == 'root':
+        return help_module.list_help_topics() + current_level_completions
+    return current_level_completions
 
 
 class UI(object):


nicholasyang2022:
This does not look right because `crm node help Overview` shows the overview of _crmsh_ instead of _node_ subcommand. I think we should fix the completion instead.

liangxin1300:
> This does not look right because `crm node help Overview` shows the overview of _crmsh_ instead of _node_ subcommand. I think we should fix the completion instead.

Changed from completion side

crmsh/command.py
```diff
@@ -231,8 +231,11 @@ def prev_completions():
 
 
 def _help_completer(args, context):
-    'TODO: make better completion'
-    return help_module.list_help_topics() + context.current_level().get_completions()
+    current_level_completions = context.current_level().get_completions()
+    help_topics_completions = help_module.list_help_topics()
+    if context.current_level().name == 'root':
+        return help_topics_completions + current_level_completions
```
nicholasyang2022: No need to load help if not at 'root' level.

```suggestion
    if context.current_level().name == 'root':
        return help_module.list_help_topics() + current_level_completions
```

PR #1880, opened 2025-07-30
Fix: Add TimeoutFormulas help topic

liangxin1300:
To show SBD and Pacemaker related timeout formulas in the help topic.


doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
 crm(test-2)configure# commit
 ...............
 
+[[topics.TimeoutFormulas,Timeout formulas]]
+=== Timeout formulas
+
+If you are using crmsh to set up a cluster with SBD as the stonith device,
+crmsh includes a set of formulas to calculate the appropriate timeouts
```
nicholasyang2022: How are these formulas used? Are they applied automatically or need users to invoke?

liangxin1300: Only automatically set if user is using crm cluster init to setup SBD

doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
 crm(test-2)configure# commit
 ...............
 
+[[topics.TimeoutFormulas,Timeout formulas]]
+=== Timeout formulas
+
+If you are using crmsh to set up a cluster with SBD as the stonith device,
```
gao-yan: We'd probably better be clear:
"If you use `crm cluster init` command to ..."

Besides, does the formula `stonith-timeout = 60 + corosync.token + corosync.consensus` in below apply without SBD setup?

doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
...
+crmsh includes a set of formulas to calculate the appropriate timeouts
+for both SBD and Pacemaker.
+
+.SBD related timeouts
+`watchdog` timeout: 15s
```
gao-yan: I assume `15s` is the internal default or by the "default" from profiles.xml? And it can be a different value with a different profile, right? 

We'd better be clear about it.

doc/crm.8.adoc
gao-yan: It should be more clear with something like:
"crmsh configures several fencing related settings based on the calculations with a set of formulas"

doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
...
+`msgwait` timeout:
+...............
+    msgwait = 2 * watchdog
+...............
+`SBD_WATCHDOG_TIMEOUT`:
+...............
+    # For diskless SBD
+    SBD_WATCHDOG_TIMEOUT = 15s
```
gao-yan: Likewise, if it can be a different value from a profile rather than hardcoded, we'd better be clear about it.

doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
...
+...............
+`SBD_DELAY_START`:
+...............
+    # For disk-based SBD
+    SBD_DELAY_START =  pcmk_delay_max + msgwait + corosync.token + corosync.consensus
+
+    # For diskless SBD
+    SBD_DELAY_START = 2 * SBD_WATCHDOG_TIMEOUT + corosync.token + corosync.consensus
+...............
+`start timeout` of sbd.service systemd unit:
```
gao-yan: We'd probably better be clear about it: "`TimeoutSec` of sbd.service systemd unit"?

doc/crm.8.adoc
```diff
@@ -296,6 +296,70 @@ INFO: test-2 shadow CIB created
...
+...............
+    if SBD_DELAY_START > 90s:
+	start_timeout = 1.2 * SBD_DELAY_START
+    else:
+	start_timeout = 90s
+...............
+.Pacemaker related timeouts
+`stonith-watchdog-timeout`: (for diskless SBD)
+...............
+    stonith-watchdog-timeout = 2 * SBD_WATCHDOG_TIMEOUT
```
gao-yan, line 351: This is occurring to me:
https://github.com/ClusterLabs/crmsh/commit/c9846e5d880957d3f4a4fb03deb2d8121e4b40fb#r163126460

doc/crm.8.adoc
liangxin1300: > We'd probably better be clear: "If you use `crm cluster init` command to ..."
> 
> Besides, does the formula `stonith-timeout = 60 + corosync.token + corosync.consensus` in below apply without SBD setup?

Yes, only when stonith-enabled=true

doc/crm.8.adoc
liangxin1300: > I assume `15s` is the internal default or by the "default" from profiles.xml? And it can be a different value with a different profile, right?

Yes, this value is from profiles.xml
> 
> We'd better be clear about it.

doc/crm.8.adoc
gao-yan: > > Besides, does the formula `stonith-timeout = 60 + corosync.token + corosync.consensus` in below apply without SBD setup?
> 
> Yes, only when stonith-enabled=true

But this initial description sounds as if this is only about the case "with SBD as the stonith device".

liangxin1300: Then, how about changing
```
# For general scenarios
```
As
```
# For general scenarios with other stonith device
```

gao-yan: Yes, it's good to make this clearer as well.

I meant for the initial description, we probably could put it with something like:
"If you use `crm cluster init` command to set up a cluster, especially with SBD as the STONITH device, crmsh configures ..."

doc/crm.8.adoc
```diff
@@ -296,6 +296,74 @@ INFO: test-2 shadow CIB created
...
+    # For diskless SBD
+    SBD_DELAY_START = 2 * SBD_WATCHDOG_TIMEOUT + corosync.token + corosync.consensus
+...............
+`TimeoutSec` of sbd.service systemd unit:
+...............
+    default_value = 1min30s
+    if SBD_DELAY_START > default_value:
+	start_timeout = 1.2 * SBD_DELAY_START
+    else:
+	start_timeout = default_value
```
gao-yan: Then I think we could refer to `TimeoutSec` rather than `start_timeout`  in here.

gao-yan (approved):
LGTM, thanks!

PR #1883, opened 2025-08-01
Fix: sbd: Ensure stonith-watchdog-timeout is >= 2 * SBD_WATCHDOG_TIMEOUT (bsc#1247415)

liangxin1300:
## Problem
Call `crm sbd configure watchdog-timeout=<new_value>` should also keep the property `stonith-watchdog-timeout` >= 2* SBD_WATCHDOG_TIMEOUT

Example:
```
# crm sbd configure show|grep -E "SBD_WATCHDOG_TIMEOUT|stonith-watchdog-timeout"
SBD_WATCHDOG_TIMEOUT=15
stonith-watchdog-timeout=30

# crm sbd configure watchdog-timeout=20
INFO: Configuring diskless SBD
WARNING: Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for 2-node cluster, should be combined with QDevice.
INFO: Update SBD_WATCHDOG_TIMEOUT in /etc/sysconfig/sbd: 20
INFO: Already synced /etc/sysconfig/sbd to all nodes
INFO: Restarting cluster service
INFO: BEGIN Waiting for cluster
...........                                                                                                                                                                                        INFO: END Waiting for cluster
INFO: Update SBD_DELAY_START in /etc/sysconfig/sbd: 51
INFO: Already synced /etc/sysconfig/sbd to all nodes

# crm sbd configure show|grep -E "SBD_WATCHDOG_TIMEOUT|stonith-watchdog-timeout"
SBD_WATCHDOG_TIMEOUT=20
stonith-watchdog-timeout=30
```
## Changed
- Dev: sbd: Ensure stonith-watchdog-timeout is >= 2 * SBD_WATCHDOG_TIMEOUT (bsc#1247415)
    
    - Refactor: Rename get_stonith_watchdog_timeout() to
      get_stonith_watchdog_timeout_expected() for clarity.
    - Logic: Now always enforces stonith-watchdog-timeout to be at least
      twice SBD_WATCHDOG_TIMEOUT.
- Dev: sbd: Restart cluster after configured sbd and adjusted properties
    
    Because some sbd configurations might already updated via
    `self.update_configuration`, then we need also update related
    properties; Otherwise, restart cluster first might meet unexpected
    behavior.
```
# crm sbd configure watchdog-timeout=20
INFO: Configuring diskless SBD
WARNING: Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for 2-node cluster, should be combined with QDevice.
INFO: Update SBD_WATCHDOG_TIMEOUT in /etc/sysconfig/sbd: 20
INFO: Already synced /etc/sysconfig/sbd to all nodes
INFO: Restarting cluster service
INFO: BEGIN Waiting for cluster
............                                                                         INFO: END Waiting for cluster
WARNING: "stonith-watchdog-timeout" in crm_config is set to 40, it was 30
INFO: Update SBD_DELAY_START in /etc/sysconfig/sbd: 51
INFO: Already synced /etc/sysconfig/sbd to all nodes

# crm sbd configure show |grep -E "SBD_WATCHDOG_TIMEOUT|stonith-watchdog-timeout"
SBD_WATCHDOG_TIMEOUT=20
stonith-watchdog-timeout=40
```
- Dev: sbd: Add with_sbd parameter for the case when sbd is not active
    
  When adding sbd via sbd stage in a running cluster, sbd.service is not
  started yet, so we need to add a with_sbd parameter to adjust
  stonith-timeout for sbd related timeouts.

### DIFF
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -711,7 +711,7 @@ class SBDManager:
 
             if self.cluster_is_running:
                 self.configure_sbd()
-                bootstrap.adjust_properties()
+                bootstrap.adjust_properties(with_sbd=True)
                 SBDManager.restart_cluster_if_possible(with_maintenance_mode=enabled)
 
     def join_sbd(self, remote_user, peer_host):
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -284,16 +284,20 @@ class SBDTimeout(object):
         return int(res)
 
     @staticmethod
-    def get_stonith_watchdog_timeout():
+    def get_stonith_watchdog_timeout_expected():
         '''
-        For non-bootstrap case, get stonith-watchdog-timeout value from cluster property
-        The default value is 2 * SBD_WATCHDOG_TIMEOUT
+        Returns the value of the stonith-watchdog-timeout cluster property.
+
+        If the Pacemaker service is inactive, returns the default value (2 * SBD_WATCHDOG_TIMEOUT).
+        If the property is set and its value is equal to or greater than the default, returns the property value.
+        Otherwise, returns the default value.
         '''
         default = 2 * SBDTimeout.get_sbd_watchdog_timeout()
         if not ServiceManager().service_is_active(constants.PCMK_SERVICE):
             return default
         value = utils.get_property("stonith-watchdog-timeout", get_default=False)
-        return int(value.strip('s')) if value else default
+        return_value = value if utils.crm_msec(value) >= utils.crm_msec(default) else default
+        return int(utils.crm_msec(return_value)/1000)  # convert msec to sec
 
     def _load_configurations(self):
         '''
@@ -309,7 +313,7 @@ class SBDTimeout(object):
         else:  # disk-less
             self.disk_based = False
             self.sbd_watchdog_timeout = SBDTimeout.get_sbd_watchdog_timeout()
-            self.stonith_watchdog_timeout = SBDTimeout.get_stonith_watchdog_timeout()
+            self.stonith_watchdog_timeout = SBDTimeout.get_stonith_watchdog_timeout_expected()
         self.sbd_delay_start_value_expected = self.get_sbd_delay_start_expected() if utils.detect_virt() else "no"
         self.sbd_delay_start_value_from_config = SBDUtils.get_sbd_value_from_config("SBD_DELAY_START")
 
@@ -574,7 +578,7 @@ class SBDManager:
         Configure fence_sbd resource and related properties
         '''
         if self.diskless_sbd:
-            swt_value = self.timeout_dict.get("stonith-watchdog", SBDTimeout.get_stonith_watchdog_timeout())
+            swt_value = self.timeout_dict.get("stonith-watchdog", SBDTimeout.get_stonith_watchdog_timeout_expected())
             utils.set_property("stonith-watchdog-timeout", swt_value)
         else:
             if utils.get_property("stonith-watchdog-timeout", get_default=False):
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -710,9 +710,9 @@ class SBDManager:
             SBDManager.enable_sbd_service()
 
             if self.cluster_is_running:
-                SBDManager.restart_cluster_if_possible(with_maintenance_mode=enabled)
                 self.configure_sbd()
                 bootstrap.adjust_properties()
+                SBDManager.restart_cluster_if_possible(with_maintenance_mode=enabled)
 
     def join_sbd(self, remote_user, peer_host):
         '''
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -2828,11 +2828,11 @@ def adjust_pcmk_delay_max(is_2node_wo_qdevice):
             logger.info("Delete parameter 'pcmk_delay_max' for resource '{}'".format(res))
 
 
-def adjust_stonith_timeout():
+def adjust_stonith_timeout(with_sbd: bool = False):
     """
     Adjust stonith-timeout for sbd and other scenarios
     """
-    if ServiceManager().service_is_active(constants.SBD_SERVICE):
+    if ServiceManager().service_is_active(constants.SBD_SERVICE) or with_sbd:
         SBDTimeout.adjust_sbd_timeout_related_cluster_configuration()
     else:
         value = get_stonith_timeout_generally_expected()
@@ -2840,7 +2840,7 @@ def adjust_stonith_timeout():
             utils.set_property("stonith-timeout", value, conditional=True)
 
 
-def adjust_properties():
+def adjust_properties(with_sbd: bool = False):
     """
     Adjust properties for the cluster:
     - pcmk_delay_max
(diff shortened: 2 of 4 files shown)


crmsh/sbd.py
```diff
@@ -293,7 +293,11 @@ def get_stonith_watchdog_timeout():
         if not ServiceManager().service_is_active(constants.PCMK_SERVICE):
             return default
         value = utils.get_property("stonith-watchdog-timeout", get_default=False)
-        return int(value.strip('s')) if value else default
+        if value:
+            property_value = int(value.strip('s'))
+            if property_value >= default:
+                return property_value
+        return default
```
nicholasyang2022: Could you please change the docstring and function name? It does no longer get the timeout value simply from the cluster property.

crmsh/sbd.py
```diff
@@ -286,14 +286,21 @@ def get_sbd_watchdog_timeout():
...
+        Otherwise, returns the default value.
         '''
         default = 2 * SBDTimeout.get_sbd_watchdog_timeout()
         if not ServiceManager().service_is_active(constants.PCMK_SERVICE):
             return default
         value = utils.get_property("stonith-watchdog-timeout", get_default=False)
-        return int(value.strip('s')) if value else default
+        if value:
+            property_value = int(value.strip('s'))
+            if property_value >= default:
```
gao-yan: Technically it's tricky to strip an "s" and compare timing values this way, in case a value takes a different unit such as "ms"...

It should be safer with something like `crm_msec(value) >= crm_msec(default)` such as how `set_property()` does it if `conditional=True`:

https://github.com/ClusterLabs/crmsh/blob/f29b30db8c2ec9c5ac393dceeb59799ba2479750/crmsh/utils.py#L2871

, or `crm_time_cmp(value, default)`.

crmsh/sbd.py
gao-yan: Indeed. Maybe call it something like `get_stonith_watchdog_timeout_expected()` like the other `*_expected()` functions?

crmsh/sbd.py
```diff
@@ -423,6 +438,7 @@ def adjust_sbd_timeout_related_cluster_configuration(cls):
         cls_inst._load_configurations()
         cls_inst.adjust_sbd_delay_start()
         cls_inst.adjust_stonith_timeout()
+        cls_inst.adjust_stonith_watchdog_timeout()
         cls_inst.adjust_systemd_start_timeout()
```
gao-yan: Isn't `adjust_sbd_timeout_related_cluster_configuration()` only called for `crm cluster init`? I don't get how it's called for `crm sbd configure`. And why is there another setting of it in `configure_sbd()`:
https://github.com/ClusterLabs/crmsh/blob/f29b30db8c2ec9c5ac393dceeb59799ba2479750/crmsh/sbd.py#L578
?

nicholasyang2022: It is called from many places.

```
adjust_sbd_timeout_related_cluster_configuration(cls)
    adjust_stonith_timeout() (crmsh.bootstrap)
        adjust_properties() (crmsh.bootstrap)
            init_qdevice() (crmsh.bootstrap)
                bootstrap_init(context) (crmsh.bootstrap)
            join_cluster(seed_host, remote_user) (crmsh.bootstrap)
                bootstrap_join(context) (crmsh.bootstrap)
            remove_node_from_cluster(node, dead_node=False) (crmsh.bootstrap)
                bootstrap_remove(context) (2 usages) (crmsh.bootstrap)
            remove_qdevice() (crmsh.bootstrap)
                bootstrap_remove(context) (crmsh.bootstrap)
            SBDManager.init_and_deploy_sbd(self) (SBDManager in crmsh.sbd)
                init_sbd() (crmsh.bootstrap)
                    bootstrap_init(context) (crmsh.bootstrap)
                SBD._configure_diskbase(self, parameter_dict) (SBD in crmsh.ui_sbd)
                    SBD.do_configure(self, context, *args) (SBD in crmsh.ui_sbd)
                SBD._configure_diskless(self, parameter_dict) (SBD in crmsh.ui_sbd)
                    SBD.do_configure(self, context, *args) (SBD in crmsh.ui_sbd)
                SBD._device_add(self, devices_to_add) (SBD in crmsh.ui_sbd)
                    SBD.do_device(self, context, *args) (SBD in crmsh.ui_sbd)
```

And `adjuest_stonith_watchdog_timeout` does not handle disk based sbd: https://github.com/ClusterLabs/crmsh/blob/a3cde868522df5e64d17a1af356b2cba8420aa98/crmsh/sbd.py#L411-L416

@liangxin1300 This is confusing. Can you handle both disk-based and diskless cases in a single function?

crmsh/sbd.py
liangxin1300: Changed

crmsh/sbd.py
liangxin1300: I removed `adjust_stonith_watchdog_timeout` this time. Since its functionality is redundant with `configure_sbd`, which can set `stonith-watchdog-timeout` for diskless-sbd

liangxin1300:
I added a new commit
```
Dev: sbd: Restart cluster after configured sbd and adjusted properties

Because some sbd configurations might already updated via
self.update_configuration, then we need also update related
properties; Otherwise, restart cluster first might meet unexpected
behavior.
```

Since I found
```
 # crm sbd configure watchdog-timeout=80
INFO: Configuring diskless SBD
WARNING: Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for 2-node cluster, should be combined with QDevice.
INFO: Update SBD_WATCHDOG_TIMEOUT in /etc/sysconfig/sbd: 80
INFO: Already synced /etc/sysconfig/sbd to all nodes
INFO: Restarting cluster service
INFO: BEGIN Waiting for cluster
..............................
```
After updating SBD_WATCHDOG_TIMEOUT with a large number, restart cluster will wait for a long time, then all nodes seem to get fenced and rebooted

crmsh/sbd.py
zzhou1: > And adjuest_stonith_watchdog_timeout does not handle disk based sbd:

It is on purpose just because stonith-watchdog-timeout is only valid for diskless sbd.

zzhou1:
> crm sbd configure update

It triggers me some thoughts for 2 cents.

I can understand the context and purpose. However, my impression of UI/UX `crm sbd configure update` is no natural enough. I'm thinking the better design is at the system level to detect and fix the related cluster configuration. 

In general, we have a lot of adjust_xxx() code already. I feel that we can add the additional _check_xxx() code. Ideally, for the example Xin mentioned above, I'm thinking when the system admin modified corosync.conf, I would think `crm corosync xxx`, `crm sbd xxx`, and `crm cluster health` could run _check_xxx() then suggestion the corresponding --fix UI.

Moreover, I try to avoid add the yet another UI to crmsh. With that, I'm thinking something like,

crm cluster health [sbd|corosync|cib|xxx] [--fix]
 
I hesitate/struggle to propose the equivalent UI at the sub command level, for example as below. It makes their UI more crowded. 

crm sbd configure health [--fix]
crm corosync health [--fix]

Once, I proposed `crm sbd status` to do some runtime _check_xxx() :)

liangxin1300:
@zzhou1 Can we merge this PR first, and have more discussion for the above topic?

nicholasyang2022 (approved):

PR #1884, opened 2025-08-04
Dev: xmlutil: Add is_non_stonith_resource_running() and use for cluster restart checks

liangxin1300:
- Introduce is_non_stonith_resource_running(), which returns True if only the
  stonith RA is running or if no RA is running.
- Replace previous usage of is_any_resource_running() with
  is_non_stonith_resource_running() for determining safe conditions to restart the cluster.

### DIFF
--- a/test/unittests/test_qdevice.py
+++ b/test/unittests/test_qdevice.py
@@ -39,7 +39,7 @@ def test_evaluate_qdevice_quorum_effect_reload(mock_get_dict, mock_quorate):
 def test_evaluate_qdevice_quorum_effect_later(mock_get_dict, mock_quorate, mock_parser):
     mock_get_dict.return_value = {'Expected': '2', 'Total': '2'}
     mock_quorate.return_value = False
-    mock_parser().is_any_resource_running.return_value = True
+    mock_parser().is_non_stonith_resource_running.return_value = True
     res = qdevice.evaluate_qdevice_quorum_effect(qdevice.QDEVICE_REMOVE)
     assert res == qdevice.QdevicePolicy.QDEVICE_RESTART_LATER
     mock_get_dict.assert_called_once_with()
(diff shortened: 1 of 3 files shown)


crmsh/qdevice.py
```diff
@@ -52,15 +52,15 @@ def evaluate_qdevice_quorum_effect(mode, diskless_sbd=False, is_stage=False):
     elif mode == QDEVICE_ADD and not is_stage:
         # Add qdevice from init process, safe to restart
         return QdevicePolicy.QDEVICE_RESTART
-    elif xmlutil.CrmMonXmlParser().is_any_resource_running():
+    elif xmlutil.CrmMonXmlParser().only_stonith_ra_running():
+        # will lose quorum, only stonith RA running or no RA running
```
gao-yan: "only stonith resources running or no resources running".

crmsh/qdevice.py
```diff
@@ -52,15 +52,15 @@ def evaluate_qdevice_quorum_effect(mode, diskless_sbd=False, is_stage=False):
...
+        # safe to restart cluster service
+        return QdevicePolicy.QDEVICE_RESTART
+    else:
         # will lose quorum, and with RA running
```
gao-yan: "with non-stonith resources running"

crmsh/xmlutil.py
```diff
@@ -1553,12 +1553,16 @@ def is_resource_configured(self, ra_type):
         xpath = f'//resource[@resource_agent="{ra_type}"]'
         return bool(self.xml_elem.xpath(xpath))
 
-    def is_any_resource_running(self):
+    def only_stonith_ra_running(self) -> bool:
```
gao-yan: It'd be better to say `resources` than `ra` like `only_stonith_reources_running()`.

crmsh/xmlutil.py
```diff
@@ -1553,12 +1553,16 @@ def is_resource_configured(self, ra_type):
...
         """
-        Check if any RA is running
+        Check if only stonith RA is running
+        Return True when all are stonith RA or no RA is running
```
gao-yan: Likewise, "resources" rather than `RA`.

gao-yan (commented):
LGTM. Only some nitpickings on the terminology.

Frankly, `only_stonith_resources_running()` is a bit counterintuitive to me. I'd go with a `is_non_stonith_resource_running()` function.

liangxin1300:
> LGTM. Only some nitpickings on the terminology.
> 
> Frankly, `only_stonith_resources_running()` is a bit counterintuitive to me. I'd go with a `is_non_stonith_resource_running()` function.

Indeed, changed
Thanks @gao-yan !

PR #1887, opened 2025-08-12
Dev: sbd: Use SBDUtils.get_sbd_device_from_config for reliable diskless SBD detection

liangxin1300:
The configure_sbd function may be invoked during the bootstrap cluster stage without the -S option. Switching to SBDUtils.get_sbd_device_from_config provides a more reliable way to determine if SBD is diskless.


nicholasyang2022:
`SBDUtils.is_using_diskless_sbd` returns false when sbd service is not running. Is it possible to be called when the cluster services are stopped?

liangxin1300:
> `SBDUtils.is_using_diskless_sbd` returns false when sbd service is not running. Is it possible to be called when the cluster services are stopped?

Use `get_sbd_device_from_config` should be better, when adding sbd via `sbd` stage, the sbd.service is not active yet

PR #1889, opened 2025-08-13
Dev: log: Print new line when progress dot is printed

liangxin1300:


nicholasyang2022:
How does this happen? The `ProgressBar` should append spaces to take the whole line:

https://github.com/ClusterLabs/crmsh/blob/70bae0711be1d45d625c606c0d62023011b69412/crmsh/log.py#L514

nicholasyang2022:
> The problem is that the progress dots and the subsequent "FAIL" or "END" log messages appear on the same line. After bootstrapping is finished, if you zoom in or out in the terminal, the "FAIL" or "END" log messages may shift to a different line.

So you need to overwrite those spaces before printing a LF, or it will be shown as an extra empty line.

Issue #1909, opened 2025-09-08
crmsh failed to copy directory by using `crm cluster copy`


liangxin1300:
```
# crm cluster copy /opt/dir1
ERROR: Failed to copy /opt/dir1 to root@alp-2: sftp> put -p '/opt/dir1' '/opt/dir1'
local "/opt/dir1/" is not a regular file
```

PR #1824, opened 2025-06-13
Dev: bootstrap: Skip csync2 by default

liangxin1300:
What's this PR changed:
- Imply and deprecate "-x" option by default. This flag skips boostrapping csync2
- Drop csync2 stage from join side; Remove csync2 from stage dependency list
- When removing cluster node by using `crm cluster remove`, don't remove csync2 configuration files, don't stop csync2.socket
- crmsh only supports configuring and bringing csync2 up, by using csync2 stage on init side:
```
# crm cluster init csync2 -y
INFO: Configuring csync2
INFO: Sync file /etc/csync2/csync2.cfg to node2
INFO: Sync file /etc/csync2/key_hagroup to node2
INFO: Starting csync2.socket service on node1
INFO: Starting csync2.socket service on node2
INFO: BEGIN csync2 syncing files
INFO: END csync2 syncing files
INFO: Done (log saved to /var/log/crmsh/crmsh.log on node1)
```
- Imply "-d" flag when calling cpio, ensuring directories are created
- Rename function utils.cluster_copy_file to utils.cluster_copy_path
- Refactor utils.cluster_copy_path, support `crm cluster copy` to copy directory #1909
- spec: Change csync2 from requies to recommends

PR #1911, opened 2025-09-09
[crmsh-4.6] support crm cluster copy to copy directory

liangxin1300:
Partly backport from #1824:

- Rename function utils.cluster_copy_file to utils.cluster_copy_path
- Refactor utils.cluster_copy_path, support crm cluster copy to copy directory #1909

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -1990,17 +1990,40 @@ def cluster_copy_file(local_path, nodes=None, output=True):
     """
     if not nodes:
         nodes = list_cluster_nodes_except_me()
-    rc = True
     if not nodes:
-        return rc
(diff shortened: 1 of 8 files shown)


crmsh/utils.py
```diff
@@ -1990,17 +1990,40 @@ def cluster_copy_file(local_path, nodes=None, output=True):
...
+    p = Path(local_path)
+    if not p.exists():
+        logger.error("%s does not exist", local_path)
+        return False
+
+    if p.is_absolute():
+        source_path = local_path
+        parent_path = p.parent
+    else:
+        absolute_path = p.resolve()
```
nicholasyang2022, line 2005: When `p` is a symlink `/foo/link -> /bar`, should we copy the symlink `/foo/link` or the link target `/bar`?

Issue #1918, opened 2025-09-18
No warning when adding ms via configure load


liangxin1300:
Write a text to configure master/slave resource and configure it as master resource
```
primitive stateful-1 ocf:pacemaker:Stateful op monitor role=Master interval=10s op monitor role=Slave interval=5s
ms master1 stateful-1
```
No warning when using `configure load update`
```
crm configure load update text
```

PR #1920, opened 2025-09-19
[crmsh-4.6] Improvements for auto convert role and ms command

liangxin1300:
## Problem

1. When keeping OCF_1_1_SUPPORT to `no`, see #1918 
2. When setting OCF_1_1_SUPPORT to `yes`, `ms` command should be automatically converted to promotable clone
3. When setting OCF_1_1_SUPPORT to `yes` in /etc/crm/crm.conf, crmsh-4.6 has the same issue which was mentioned in https://github.com/ClusterLabs/crmsh/pull/1919#issue-3430202283

## Changed
- To address problems 1 and 2, introduce utils.handle_deprecated_ms_command function, convert `ms` command to promotable clone if meeting conditions, otherwise, raise warning.
- To address problem 3, backport from #1919

### DIFF
--- a/crmsh/utils.py
+++ b/crmsh/utils.py
@@ -3301,4 +3301,20 @@ def validate_and_get_reachable_nodes(
             member_list.remove(node)
 
     return member_list + remote_list
+
+
+def handle_deprecated_ms_command(cli_str: str) -> str:
+    if not re.search(r'^ms\s+', cli_str) or not is_ocf_1_1_cib_schema_detected():
+        return cli_str
+
+    meta_str = " " if "meta" in cli_str else " meta "
+    clone_str = re.sub(r'^ms\s+', 'clone ', cli_str)
+    new_cli_str = f"{clone_str}{meta_str}promotable=true"
+
+    if config.core.OCF_1_1_SUPPORT:
+        logger.info("Convert deprecated \"ms\" command to promotable clone: %s", new_cli_str)
+        return new_cli_str
+    else:
+        logger.warning("The \"ms\" command is deprecated, please use: %s", new_cli_str)
+        return cli_str
 # vim:ts=4:sw=4:et:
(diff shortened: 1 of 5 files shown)


liangxin1300:
When OCF_1_1_SUPPORT to `no`, raise a warning when editing
```
# crm configure show related:stateful-1
        op monitor role=Master interval=10s \
        op monitor role=Slave interval=5s
ms master1 stateful-1

# crm configure edit
WARNING: The "ms" command is deprecated, please use: clone master1 stateful-1 meta promotable=true
```
Raise a warning when loading
```
# crm configure load update <(echo "ms master1 stateful-1")
WARNING: The "ms" command is deprecated, please use: clone master1 stateful-1 meta promotable=true
```

crmsh/utils.py
```diff
@@ -3301,4 +3301,19 @@ def validate_and_get_reachable_nodes(
...
+
+def handle_deprecated_ms_command(cli_str: str) -> str:
+    if not re.search(r'^ms\s+', cli_str) or not is_ocf_1_1_cib_schema_detected():
+        return cli_str
+
+    meta_str = " " if "meta" in cli_str else " meta "
+    clone_str = re.sub(r'^ms\s+', 'clone ', cli_str)
+    new_cli_str = f"{clone_str}{meta_str}promotable=true"
+
+    if config.core.OCF_1_1_SUPPORT:
```
nicholasyang2022, line 3314: To get best performance, boolean flags should be check first.

liangxin1300: I don't get your point.

In this function, we should detect if the given string is `ms` statement and the schema version supports ocf 1.1, then check if the option OCF_1_1_SUPPORT enabled

nicholasyang2022: ~If `OCF_1_1_SUPPORT` is disabled, you could just return the original string without creating `new_cli_str`.~

nicholasyang2022: Oh, it is needed in the warning message when `OCF_1_1_SUPPORT` is disabled.

Can you use `else` statement to make this more clear?

PR #1929, opened 2025-09-24
Fix: cibverify: If no errors found, treat crm_verify result as success (bsc#1250349) 

liangxin1300:
Port from #1926

### DIFF
--- a/test/bugs-test.txt
+++ b/test/bugs-test.txt
@@ -1,4 +1,4 @@
-property stonith-enabled=false
+property fencing-enabled=false
 node node1
 op_defaults timeout=60s
 group g1 gr1 gr2
--- a/test/cibtests/001.exp.xml
+++ b/test/cibtests/001.exp.xml
@@ -2,7 +2,7 @@
   <configuration>
     <crm_config>
       <cluster_property_set id="cib-bootstrap-options">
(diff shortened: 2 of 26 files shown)


test/features/cluster_api.feature
```diff
@@ -162,7 +162,7 @@ Feature: Functional test to cover SAP clusterAPI
 
     When    Run "crm -F configure primitive d-require-fence Dummy meta requires=fencing" on "hanode1"
     When    Try "crm configure verify"
-    Then    Expected return code is "1"
+    Then    Expected return code is "0"
     Then    Expected "Configuration invalid (with warnings)" in stdout
```
zzhou1: Caution that after pacemaker#3933 this will be changed to "Configuration may need attention"

liangxin1300: Then we update the test case after pacemaker upgraded

zzhou1: How about this,

```suggestion
    Then    Expect stdout contains snippets ["Configuration invalid (with warnings)", "Configuration may need attention"]
```

liangxin1300: We can hold on to this PR, until pacemaker#3933 submits to Factory

zzhou1 (approved):

PR #1934, opened 2025-10-05
Dev: sbd: Decrease sbd related timeouts when expected value decreases

liangxin1300:


crmsh/sbd.py
```diff
@@ -331,7 +314,8 @@ def get_stonith_timeout_expected(self):
         if self.disk_based:
             value_from_sbd = int(1.2*self.msgwait)
         else:
-            value_from_sbd = int(1.2*max(self.stonith_watchdog_timeout, 2*self.sbd_watchdog_timeout))
+            stonith_watchdog_timeout = int(utils.get_property("stonith-watchdog-timeout") or 2*self.sbd_watchdog_timeout)
```
nicholasyang2022:
```suggestion
            stonith_watchdog_timeout = int(utils.get_property("stonith-watchdog-timeout") or -1)
```

crmsh/sbd.py
```diff
@@ -187,6 +187,7 @@ class SBDTimeout(object):
     SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE = 35
     QDEVICE_SYNC_TIMEOUT_MARGIN = 5
     SHOW_SBD_START_TIMEOUT_CMD = "systemctl show -p TimeoutStartUSec sbd.service --value"
+    SBD_START_TIMEOUT_DEFAULT = 90
```
nicholasyang2022: Instead of hard coding the default, it would be better to read `DefaultTimeoutStartUSec` from `systemctl show` command or systemd's dbus API.

crmsh/sbd.py
liangxin1300: In #1865 we set the default value as 2* SBD_WATCHDOG_TIMEOUT

nicholasyang2022: Then I think you should reuse that function to get `stonith_watchdog_timeout`.

PR #1940, opened 2025-10-13
Dev: Improve to avoid circular imports

liangxin1300:
## What's changed:
- Dev: Import only module instead of functions from module
   - Avoid potential circular imports
   - Avoid namespace conflicts
- Dev: bootstrap: Lazily initialize tempfile to store original corosync.conf in Context to avoid circular import
- Dev: Get instance of CibFactory by using cibconfig.cib_factory_instance()
- Dev: bootstrap: Avoid circular import by refactoring FILES_TO_SYNC to use get_files_to_sync

### DIFF
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -1072,10 +1072,10 @@ done
     def test_is_online_peer_offline(self, mock_parser, mock_get_hostname, mock_this_node,
             mock_copy, mock_corosync_conf, mock_sync, mock_stop_service, mock_error, mock_cluster_shell):
         bootstrap._context = mock.Mock(cluster_node='node1')
+        bootstrap._context.get_corosync_conf_orig.return_value = "/tmp/crmsh_tempfile"
         mock_parser_inst = mock.Mock()
         mock_parser.return_value = mock_parser_inst
         mock_parser_inst.is_node_online.side_effect = [True, False]
-        bootstrap.COROSYNC_CONF_ORIG = "/tmp/crmsh_tmpfile"
         mock_this_node.return_value = "node2"
         mock_get_hostname.return_value = "node1"
         mock_corosync_conf.side_effect = [ "/etc/corosync/corosync.conf",
@@ -1089,7 +1089,7 @@ done
             mock.call(),
             mock.call()
             ])
-        mock_copy.assert_called_once_with(bootstrap.COROSYNC_CONF_ORIG, "/etc/corosync/corosync.conf")
+        mock_copy.assert_called_once_with(bootstrap._context.get_corosync_conf_orig.return_value, "/etc/corosync/corosync.conf")
         mock_sync.assert_called_once_with("/etc/corosync/corosync.conf", "node1")
         mock_stop_service.assert_called_once_with("corosync")
         mock_error.assert_called_once_with("Cannot see peer node \"node1\", please check the communication IP")
@@ -1483,8 +1483,9 @@ done
 
     @mock.patch('crmsh.utils.cluster_run_cmd')
     @mock.patch('os.path.isfile')
-    def test_sync_files_to_disk(self, mock_isfile, mock_cluster_cmd):
-        bootstrap.FILES_TO_SYNC = ("file1", "file2")
+    @mock.patch('crmsh.bootstrap.get_files_to_sync')
+    def test_sync_files_to_disk(self, mock_get_files, mock_isfile, mock_cluster_cmd):
+        mock_get_files.return_value = ["file1", "file2"]
         mock_isfile.side_effect = [True, True]
         bootstrap.sync_files_to_disk()
         mock_isfile.assert_has_calls([mock.call("file1"), mock.call("file2")])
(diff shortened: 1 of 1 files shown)


crmsh/bootstrap.py
```diff
@@ -68,10 +68,6 @@
 BOOTH_DIR = "/etc/booth"
 BOOTH_CFG = "/etc/booth/booth.conf"
 BOOTH_AUTH = "/etc/booth/authkey"
-FILES_TO_SYNC = (BOOTH_DIR, corosync.conf(), COROSYNC_AUTH, CSYNC2_CFG, CSYNC2_KEY, "/etc/ctdb/nodes",
-        "/etc/drbd.conf", "/etc/drbd.d", "/etc/ha.d/ldirectord.cf", "/etc/lvm/lvm.conf", "/etc/multipath.conf",
-        "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK, sbd.SBDManager.SYSCONFIG_SBD, PCMK_REMOTE_AUTH, watchdog.Watchdog.WATCHDOG_CFG,
-        PROFILES_FILE, CRM_CFG, sbd.SBDManager.SBD_SYSTEMD_DELAY_START_DIR)
```
nicholasyang2022: I can't found any reference to `FILES_TO_SYNC` from other modules.

crmsh/bootstrap.py
```diff
@@ -137,6 +133,10 @@ def __init__(self):
         self.profiles_dict = {}
         self.default_nic = None
         self.default_ip_list = []
+        self.files_to_sync = (BOOTH_DIR, corosync.conf(), COROSYNC_AUTH, CSYNC2_CFG, CSYNC2_KEY,
+            "/etc/ctdb/nodes", "/etc/drbd.conf", "/etc/drbd.d", "/etc/ha.d/ldirectord.cf", "/etc/lvm/lvm.conf",
+            "/etc/multipath.conf", "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK, sbd.SBDManager.SYSCONFIG_SBD,
+            PCMK_REMOTE_AUTH, watchdog.Watchdog.WATCHDOG_CFG, PROFILES_FILE, CRM_CFG, sbd.SBDManager.SBD_SYSTEMD_DELAY_START_DIR)
```
nicholasyang2022: Refactor it to a member variable weaken the constant semantics, making the code harder to understand.

crmsh/bootstrap.py
```diff
@@ -535,7 +540,7 @@ def is_online():
     user, cluster_node = _parse_user_at_host(_context.cluster_node, None)
     cluster_node = get_node_canonical_hostname(cluster_node)
     if not xmlutil.CrmMonXmlParser().is_node_online(cluster_node):
-        shutil.copy(COROSYNC_CONF_ORIG, corosync.conf())
+        shutil.copy(_context.get_corosync_conf_orig(), corosync.conf())
```
nicholasyang2022, line 542: This does not look good. It is an undesired side effect to replace the configuration file when checking if the node is online.

crmsh/bootstrap.py
liangxin1300: sbd will import bootstrap:
```
 # crm 
  File "/sbin/crm", line 35, in <module>
    from crmsh import main
  File "/usr/lib/python3.13/site-packages/crmsh/main.py", line 14, in <module>
    from . import utils
  File "/usr/lib/python3.13/site-packages/crmsh/utils.py", line 47, in <module>
    from . import sbd
  File "/usr/lib/python3.13/site-packages/crmsh/sbd.py", line 6, in <module>
    from . import bootstrap
  File "/usr/lib/python3.13/site-packages/crmsh/bootstrap.py", line 72, in <module>
    "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK, sbd.SBDManager.SYSCONFIG_SBD, PCMK_REMOTE_AUTH, watchdog.Watchdog.WATCHDOG_CFG,
                                                          ^^^^^^^^^^^^^^
AttributeError: partially initialized module 'crmsh.sbd' from '/usr/lib/python3.13/site-packages/crmsh/sbd.py' has no attribute 'SBDManager' (most likely due to a circular import)
```

crmsh/bootstrap.py
liangxin1300: I don't get the point. What is the side effect?

nicholasyang2022: > What is the side effect?

`/etc/corosync/corosync.conf` get overwritten by a temporary file.

crmsh/bootstrap.py
nicholasyang2022: So you did not solve the circular dependencies. You just silent the error by lazy evaluation.

crmsh/bootstrap.py
liangxin1300: This is by design. If the code goes here, that means this join node can't see the init node, maybe due to the crossed network(see [1]), at this time, we should restore the original corosync.conf and sync to other nodes to make sure other nodes have the original corosync.conf, not the wrong one.


[1]: https://github.com/ClusterLabs/crmsh/blob/master/test/features/bootstrap_bugs.feature#L64

crmsh/bootstrap.py
liangxin1300: Then how about this way to fix circular Import without losing constant semantics?
```
STATIC_FILES_TO_SYNC = (
    BOOTH_DIR, COROSYNC_AUTH, CSYNC2_CFG, CSYNC2_KEY, "/etc/ctdb/nodes",
    "/etc/drbd.conf", "/etc/drbd.d", "/etc/ha.d/ldirectord.cf", "/etc/lvm/lvm.conf",
    "/etc/multipath.conf", "/etc/samba/smb.conf", SYSCONFIG_NFS, SYSCONFIG_PCMK,
    PCMK_REMOTE_AUTH, PROFILES_FILE, CRM_CFG
)

def get_files_to_sync():
    from . import corosync, sbd, watchdog
    return (
        (corosync.conf(), sbd.SBDManager.SYSCONFIG_SBD, watchdog.Watchdog.WATCHDOG_CFG, sbd.SBDManager.SBD_SYSTEMD_DELAY_START_DIR)
        + STATIC_FILES_TO_SYNC
    )
```

crmsh/bootstrap.py
liangxin1300: This approach is only a workaround. To completely eliminate circular import issues, it may be necessary to refactor the code and extract shared functionality into separate modules, which could require more substantial effort.

crmsh/bootstrap.py
nicholasyang2022: LGTM. It is good as long as we don't add it to `Context`, as `Context` is used for passing variables around.

nicholasyang2022 (approved):

PR #1941, opened 2025-10-16
Fix: sh: Avoid unnecessary sudo calls in ClusterShell SSH commands (bsc#1229416)

nicholasyang2022:
The `subprocess_run_without_input` method in `ClusterShell` was unconditionally including `sudo -H -u ${user}` in its SSH command construction, even when the target user is the same as the remote user.

This commit modifies the logic to conditionally include `sudo -H -u ${user}` only when the target `user` is different from the `remote_user`. This prevents unnecessary `sudo` calls, which is important in environments where `sudo` might not be installed or configured, and improves the robustness of SSH command execution.

### DIFF
--- a/crmsh/prun/prun.py
+++ b/crmsh/prun/prun.py
@@ -13,7 +13,7 @@ from crmsh.sh import Utils
 
 _DEFAULT_CONCURRENCY = 32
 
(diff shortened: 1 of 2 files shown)


liangxin1300 (commented):
As this PR is a bug fix, please add behave functional test case to avoid possible regression.

Thanks!

liangxin1300 (approved):

PR #1944, opened 2025-10-16
Dev: ui_sbd: Add and check the minimum SBD timeout value

liangxin1300:
Raise ValueError if the timeout value is less than the minimum allowed value.


crmsh/ui_sbd.py
```diff
@@ -89,8 +89,19 @@ class SBD(command.UI):
     - sbd purge
     '''
     name = "sbd"
-    TIMEOUT_TYPES = ("watchdog", "allocate", "loop", "msgwait", "crashdump-watchdog")
-    DISKLESS_TIMEOUT_TYPES = ("watchdog", "crashdump-watchdog")
+    TIMEOUT_TYPES = {
```
nicholasyang2022: It would be better if the name can describe what the values is. For example, `TIMEOUT_TYPE_MINIMUMS`

Issue #1953, opened 2025-10-24
The /etc/systemd/system/sbd.service.d directory remains after the node was removed


liangxin1300:
The file `sbd_delay_start_disabled.conf`(see #1092) in `/run/systemd/system/sbd.service.d` remains, too.

PR #1962, opened 2025-10-30
Dev: sbd: Remove sbd configuration directories while removing cluster node

liangxin1300:
## Problem
As noted in #1953, after removing a cluster node, the directories `/etc/systemd/system/sbd.service.d` and `/run/systemd/system/sbd.service.d` are left behind.

## Changes
- Ensure both directories are removed when a cluster node is removed.
- Refactor the cleanup logic into a new function, `sbd.cleanup_sbd_configurations`, which is now invoked during both `cluster remove <node>` and `sbd purge` subcommands

### DIFF
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -1964,16 +1964,17 @@ class TestValidation(unittest.TestCase):
         mock_run.assert_called_once_with("node2", "crm cluster remove -y -c node1")
         mock_error.assert_called_once_with("Failed to remove this node from node2: err")
 
-    @mock.patch('crmsh.utils.package_is_installed')
+    @mock.patch('crmsh.sbd.cleanup_sbd_configurations')
+    @mock.patch('os.path.exists')
     @mock.patch('crmsh.sh.ClusterShell.get_stdout_or_raise_error')
-    def test_rm_configuration_files(self, mock_run, mock_installed):
+    def test_rm_configuration_files(self, mock_run, mock_exists, mock_rm_sbd):
         bootstrap._context = mock.Mock(rm_list=["file1", "file2"])
-        mock_installed.return_value = True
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -2137,10 +2137,8 @@ def rm_configuration_files(remote=None):
     """
     shell = sh.cluster_shell()
     shell.get_stdout_or_raise_error("rm -f {}".format(' '.join(_context.rm_list)), remote)
```
nicholasyang2022, line 2139: `sbd.SBDManager.SYSCONFIG_SBD` is removed here.

crmsh/sbd.py
```diff
@@ -813,6 +812,19 @@ def cleanup_existing_sbd_resource():
         utils.ext_cmd("crm configure delete {}".format(' '.join(sbd_id_list)))
 
 
+def cleanup_sbd_configurations(remote=None):
+    shell = sh.cluster_shell()
+    sysconfig_sbd_bak = f"{SBDManager.SYSCONFIG_SBD}.bak"
+    logger.info("Rename %s to %s on node %s",
+                SBDManager.SYSCONFIG_SBD, sysconfig_sbd_bak, remote or utils.this_node())
+    cmd = f"test -f {SBDManager.SYSCONFIG_SBD} && mv {SBDManager.SYSCONFIG_SBD} {sysconfig_sbd_bak} || exit 0"
```
nicholasyang2022, line 820: It is too late too bakcup `SBDManager.SYSCONFIG_SBD`.

crmsh/bootstrap.py
liangxin1300: No, in this PR, I removed it from rm_list

nicholasyang2022 (approved):

PR #1984, opened 2025-11-25
Fix: bootstrap: On join node, retrieve qdevice certification files before starting qdevice (bsc#1254243)

liangxin1300:
## Problem
When setup qdevice via init and join commands:
```
- crm cluster init --qnetd-hostname <qnetd> -y
- crm cluster join -c <init_node> -y
```

The name of the qdevice-related certification directory under `/etc/corosync/qdevice/net` is different between nodes:
```
# for example on the init node hanode1:
hanode1:/etc/corosync/qdevice/net # ls
nssdb  qnetd-node

# on join node hanode2:
hanode2:/etc/corosync/qdevice/net # ls
hanode1  nssdb
```
That's because the join node called `crmsh.parallax.parallax_slurp` to fetch the certification files from init node, then left the directory with init node name, which might cause problems when the node re-joins the cluster

## Solution
On the join node, retrieve the qdevice certification files from the peer node's qdevice path (/etc/corosync/qdevice/net) before starting qdevice
    
And remove the function start_qdevice_on_join_node, no need to do certificate on join node for qdevice, because the whole certification process only binds with the cluster name, no related with the specific node info.

### DIFF
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -1421,29 +1421,6 @@ done
         mock_remove_db.assert_called_once_with()
         mock_cluster_shell_inst.get_stdout_or_raise_error.assert_called_once_with("corosync-cfgtool -R")
 
-    @mock.patch('crmsh.utils.this_node')
-    @mock.patch('crmsh.service_manager.ServiceManager.start_service')
-    @mock.patch('crmsh.qdevice.QDevice')
-    @mock.patch('crmsh.corosync.get_value')
-    @mock.patch('crmsh.corosync.is_qdevice_tls_on')
-    @mock.patch('crmsh.log.LoggerUtils.status_long')
-    def test_start_qdevice_on_join_node(self, mock_status_long, mock_qdevice_tls, mock_get_value, mock_qdevice, mock_start_service, mock_this_node):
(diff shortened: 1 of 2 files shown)


crmsh/bootstrap.py
```diff
@@ -2801,14 +2791,18 @@ def adjust_properties(with_sbd: bool = False):
...
 
-def retrieve_all_config_files(cluster_node):
-    """
-    Retrieve config files from cluster_node if exists
-    """
-    with logger_utils.status_long("Retrieve all config files"):
-        cmd = 'cpio -o << EOF\n{}\nEOF\n'.format(
-            '\n'.join((f for f in get_files_to_sync() if f != CSYNC2_KEY and f != CSYNC2_CFG))
-        )
+def retrieve_data(from_node, data_list=None, data_type=None):
```
nicholasyang2022: 1. It would be more clear to name it `retrieve_files` instead of `retrieve_data`, as the word "data" is too general.
2. And also `file_list`.
3. `data_type` is only used for outputting messages and has no actual functions. Passing it makes the code harder to understand. It would be better to pass `msg` directly, or use a more clear parameter name to reflect its usage.

liangxin1300: Already changed in PR
- #2005

PR #1987, opened 2025-11-27
Dev: utils: Check if the peer node needs password to access

liangxin1300:
In the function `utils.check_all_nodes_reachable`, after checking if the node is reachable via SSH, we should also check if it requires a password for specific users. Otherwise, raise UnreachableNodeError to prevent further operations that might cause issues.

### DIFF
--- a/test/unittests/test_qdevice.py
+++ b/test/unittests/test_qdevice.py
@@ -203,7 +203,7 @@ class TestQDevice(unittest.TestCase):
         self.assertEqual(res, "/etc/corosync/qdevice/net/node1.com/qdevice-net-node.p12")
 
     @mock.patch('crmsh.utils.InterfacesInfo.ip_in_local')
-    @mock.patch('crmsh.utils.ssh_reachable_check')
+    @mock.patch('crmsh.utils.ssh_port_reachable_check')
     @mock.patch('socket.getaddrinfo')
     def test_check_qnetd_addr_local(self, mock_getaddrinfo, mock_reachable, mock_in_local):
         mock_getaddrinfo.return_value = [(None, ("10.10.10.123",)),]
--- a/test/unittests/test_utils.py
+++ b/test/unittests/test_utils.py
@@ -982,7 +982,7 @@ def test_is_block_device(mock_stat, mock_isblk):
     mock_isblk.assert_called_once_with(12345)
 
 
-@mock.patch('crmsh.utils.ssh_reachable_check')
+@mock.patch('crmsh.utils.ssh_port_reachable_check')
 @mock.patch('crmsh.xmlutil.CrmMonXmlParser')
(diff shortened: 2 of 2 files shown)


liangxin1300:
Note: this PR is after the PR:
- #1860

crmsh/utils.py
```diff
@@ -2495,8 +2499,35 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
...
+            unreachable_nodes.append(node)
+            continue
+
+        if check_passwd:
+            local_user, remote_user = crmsh.user_of_host.UserOfHost.instance().user_pair_for_ssh(node)
+            if check_ssh_passwd_need(local_user, remote_user, node):
+                password_needed_nodes.append(node)
+
+    if unreachable_nodes:
+        msg = f"""There are unreachable nodes: {', '.join(unreachable_nodes)}.
```
zzhou1:
```suggestion
        msg = f"""There are nodes whose ssh ports are unreachable: {', '.join(unreachable_nodes)}.
```

crmsh/utils.py
```diff
@@ -2495,8 +2499,35 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
...
+Please check the network connectivity before {action_to_do}.
+        """
+        raise UnreachableNodeError(msg, unreachable_nodes)
+
+    if password_needed_nodes:
+        msg = f"""There are nodes need password for SSH access: {', '.join(password_needed_nodes)}.
```
zzhou1:
```suggestion
        msg = f"""There are nodes which requires a password for SSH access: {', '.join(password_needed_nodes)}.
```

crmsh/utils.py
```diff
@@ -2428,25 +2428,23 @@ def package_is_installed(pkg, remote_addr=None):
     return rc == 0
 
 
-def node_reachable_check(node, ping_count=1, port=22, timeout=3):
+def ssh_reachable_check(node):
```
zzhou1: When reading "ssh_reachable" in the code, some users, myself included, might interpret it as "able to log in via SSH." With that, I would personally suggest to change the function name to be more specifically for readability.

Similarly, s/validate_and_get_reachable_nodes/validate_and_get_ssh_port_reachable_nodes/

```suggestion
def ssh_port_reachable_check(node):
```

zzhou1 (approved):
Approved with some cosmetic suggestions if you would like to take.

crmsh/utils.py
```diff
@@ -2495,8 +2499,35 @@ def check_all_nodes_reachable(action_to_do: str, peer_node: str = None):
         """
         raise DeadNodeError(msg, dead_nodes)
 
+    unreachable_nodes = []
+    password_needed_nodes = []
```
zzhou1:
```suggestion
    nodes__ssh_port_unreachable = []
    nodes__need_password = []
```

PR #1996, opened 2025-12-17
Fix: log: Add milliseconds time format to crmsh.log (bsc#1255021)

liangxin1300:
- Add milliseconds time format to crmsh.log
- Add milliseconds time format to corosync.log

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -7,6 +7,7 @@ import shutil
 import logging
 import logging.config
 import typing
+import time
 from contextlib import contextmanager
 
 from . import options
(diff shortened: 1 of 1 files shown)


nicholasyang2022:
> OK, then I will keep the datefmt unchanged, just add milliseconds

No, ISO 8601 timestamp forbids milliseconds. I think the original request is to keep consistent format with pacemaker.log. So it won't help to add milliseconds.

This is a suitable use case of `/etc/crm.conf`. We can add some configurable options for log format.

liangxin1300:
> No, ISO 8601 timestamp forbids milliseconds.

Add milliseconds
```
2025-12-18T14:18:22+0800.518 sle16-1 crmsh.bootstrap: INFO: Done (log saved to /var/log/crmsh/crmsh.log on sle16-1)
```
without milliseconds
```
2025-12-18T14:20:53+0800 sle16-1 crmsh.ui_context: INFO: crm(live/sle16-1)# status
```

> This is a suitable use case of /etc/crm.conf. We can add some configurable options for log format.

If we add the `log_date_format` option, what kind of options do we support? `syslog` and `rfc5424`(default) ?

zzhou1 (approved):

Issue #1936, opened 2025-10-08
Dupicated behave_agent.py


liangxin1300:
There are two behave_agent.py in [1] and [2], should drop one  to avoid conflict

[1]: https://github.com/ClusterLabs/crmsh/blob/master/test_container/behave_agent.py
[2]: https://github.com/ClusterLabs/crmsh/blob/master/test/features/steps/behave_agent.py

nicholasyang2022:
I think we should separate them into `behave_agent_server.py` in `test_container/` and `behave_agent_client.py` in `steps/`.

liangxin1300:
> I think we should separate them into `behave_agent_server.py` in `test_container/` and `behave_agent_client.py` in `steps/`.

Could you please make change?

Thanks!

PR #2004, opened 2025-12-25
Dev: behave_agent: Prevent client hang behind firewall during long cluster joins

liangxin1300:
The behave_agent client could hang when a `crm cluster join` command took a long time (~100s) to finish. In that case the TCP connection to port 1122 sat idle long enough for firewalld/conntrack to drop the flow, so the server’s response (RC/OUT/ERR/EOF) never reached the client and the call blocked indefinitely in `recv()`.

Fix this by enabling aggressive TCP keepalive on the client socket:

- SO_KEEPALIVE = 1
- TCP_KEEPIDLE  = 30s  (start keepalive probes after 30s idle)
- TCP_KEEPINTVL = 10s  (send probes every 10s)
- TCP_KEEPCNT   = 3    (3 failed probes before the OS considers the connection dead)

These keepalive probes keep the connection active in the firewall’s state table, so long‑running `crm cluster join` operations (~100s) now complete successfully without the client hanging.


liangxin1300:
@nicholasyang2022 Please also see 
- #1936

Which one should we remove?

Issue #2011, opened 2026-01-13
Don't set stonith-enabled=true after adding sbd via stage


liangxin1300:
Introduced by #2003

PR #2013, opened 2026-01-15
Dev: sbd: Check and fix stonith_enabled property

liangxin1300:
Fix issue:
- #2011


crmsh/sbd.py
```diff
@@ -811,6 +819,18 @@ def _fix_stonith_timeout(self):
         logger.info("Adjusting stonith-timeout to %d", expected_value)
         utils.set_property("stonith-timeout", expected_value)
 
+    def _check_stonith_enabled(self) -> CheckResult:
+        value = utils.get_property("stonith-enabled", get_default=False)
+        if utils.is_boolean_false(value):
+            if not self.quiet:
+                logger.error("It's recommended that stonith-enabled is set to true, now is false")
```
zzhou1: "Error" means it is mandatory. Wording wise, it is "required", not "recommended". Or, something like,

"The cluster option 'stonith-enabled' must be 'true'. Currently, it is 'false'."

liangxin1300: Yes, then multiple similar places in sbd.py should be changed
I will make change in another commit

Issue #2010, opened 2026-01-12
The `sbd` stage operation on a down cluster should be rejected immediately


zzhou1:
As the title, using the `sbd` stage to migrate the "disk-based sbd" cluster to "diskless sbd" leads to an incomplete sbd configuration without an error reported, eg. missing stonith-watchdog-timeout. 

Here is the reproducer:

```
adm@160-1:~> sudo -E crm cluster init -ys /dev/disk/by-partlabel/sbd-160 -N adm@160-2
adm@160-1:~> sudo -E crm cluster stop --all

adm@160-1:~> sudo -E crm cluster init sbd -yS
INFO: Loading "default" profile from /etc/crm/profiles.yml                                                                                                                                                          
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml                                                                                                                                                     
WARNING: Pacemaker state transition is in progress. Skip restarting cluster in maintenance mode.                                                                                                                    
INFO: Configuring diskless SBD                                                                                                                                                                                      
WARNING: Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for 2-node cluster, should be combined with QDevice.                                                               
INFO: Update SBD_WATCHDOG_TIMEOUT in /etc/sysconfig/sbd: 15                                                                                                                                                         
INFO: Update SBD_WATCHDOG_DEV in /etc/sysconfig/sbd: /dev/watchdog0                                                                                                                                                 
INFO: Done (log saved to /var/log/crmsh/crmsh.log on 160-1) 
```

Issue #1999, opened 2025-12-19
Different configurations for sbd for singal node when using different commands


liangxin1300:
## Setup disk-based with cluster init
```
# crm cluster init -s /dev/sda5 -y

# crm sbd configure show
INFO: crm sbd configure show disk_metadata
==Dumping header on disk /dev/sda5
Header version     : 2.1
UUID               : a8ab7915-5973-488f-83f5-1c7ee92e9073
Number of slots    : 255
Sector size        : 512
Timeout (watchdog) : 15
Timeout (allocate) : 2
Timeout (loop)     : 1
Timeout (msgwait)  : 30
==Header on disk /dev/sda5 is dumped


INFO: crm sbd configure show sysconfig
SBD_PACEMAKER=yes
SBD_STARTMODE=always
SBD_DELAY_START=no
SBD_WATCHDOG_DEV=/dev/watchdog
SBD_TIMEOUT_ACTION=flush,reboot
SBD_MOVE_TO_ROOT_CGROUP=auto
SBD_SYNC_RESOURCE_STARTUP=yes
SBD_OPTS=
SBD_DEVICE=/dev/sda5

INFO: crm sbd configure show property
have-watchdog=true
stonith-enabled=true

INFO: crm configure show related:fence_sbd
primitive stonith-sbd stonith:fence_sbd

INFO: systemctl show -p TimeoutStartUSec sbd.service --value
TimeoutStartUSec=90
```
## Setup disk-based with sbd stage
```
# crm cluster init -y
# crm cluster init sbd -s /dev/sda5 -y
# crm sbd configure show
INFO: crm sbd configure show disk_metadata
==Dumping header on disk /dev/sda5
Header version     : 2.1
UUID               : 53bc14eb-85d1-4930-97da-6503b625370b
Number of slots    : 255
Sector size        : 512
Timeout (watchdog) : 15
Timeout (allocate) : 2
Timeout (loop)     : 1
Timeout (msgwait)  : 30
==Header on disk /dev/sda5 is dumped


INFO: crm sbd configure show sysconfig
SBD_PACEMAKER=yes
SBD_STARTMODE=always
SBD_DELAY_START=41
SBD_WATCHDOG_DEV=/dev/watchdog
SBD_TIMEOUT_ACTION=flush,reboot
SBD_MOVE_TO_ROOT_CGROUP=auto
SBD_SYNC_RESOURCE_STARTUP=yes
SBD_OPTS=
[...]

PR #2021, opened 2026-01-28
Minor improvement for sbd

liangxin1300:
- Dev: sbd: Add quiet attribute to class SBDTimeout to suppress warnings when not needed (when calling classmethod SBDTimeout.get_sbd_watchdog_timeout_expected)

- Dev: sbd: Adjust related properties after cluster started on init process
  Fix the issue:
  - #1999
 
- Dev: bootstrap: The sbd stage now requires a running cluster
  Fix the issue:
  - #2010
 
 - Dev: sbd: Replace magic numbers with IntEnum for check prerequisites
   Replace hard-coded integer indices with enum.IntEnum to track SBD check items.

 - Dev: sbd: rename SBDTimeoutChecker to SBDConfigChecker
   The checker now validates general SBD metadata, service status, and consistency, rather than being limited to timeout-related properties.

- Dev: sbd: Move diskless SBD warning to adjust_properties
  Since the warning logic is dependent on quorum calculation results, it is more accurate to evaluate it within `adjust_properties` after the cluster membership and expected votes have been updated.


crmsh/sbd.py
```diff
@@ -243,7 +246,8 @@ def _set_sbd_msgwait(self):
         if "sbd.msgwait" in self.context.profiles_dict:
             sbd_msgwait = int(self.context.profiles_dict["sbd.msgwait"])
             if sbd_msgwait < sbd_msgwait_default:
-                logger.warning("sbd msgwait is set to %d, it was %d", sbd_msgwait_default, sbd_msgwait)
+                if not self.quiet:
+                    logger.warning("sbd msgwait is set to %d, it was %d", sbd_msgwait_default, sbd_msgwait)
```
nicholasyang2022: It would be better to add a `self.log` method to avoid checking `self.quiet` everywhere.

crmsh/bootstrap.py
```diff
@@ -2790,6 +2792,7 @@ def adjust_properties():
     adjust_stonith_timeout()
     adjust_priority_in_rsc_defaults(is_2node_wo_qdevice)
     adjust_priority_fencing_delay(is_2node_wo_qdevice)
```
zzhou1, line 2794: Some findings to verify https://github.com/ClusterLabs/crmsh/issues/2010

\1. cosmetic finding

> WARNING: "priority-fencing-delay" in crm_config is set to 0, it was 60

Nitpicking, to delete "priority-fencing-delay" to simplify CIB eventually since the default value is 0. The above output could be something like,

WARNING: "priority-fencing-delay" in crm_config is deleted, whose default is 0. It was 60

\2. yet another one, a little strange, the order of the output after `sbd` stage is different than the cluster directly bootstrapped as diskless. It makes more sense to have the stable order from the impression.

INFO: crm sbd configure show property
have-watchdog=true
stonith-enabled=true
stonith-timeout=71
stonith-watchdog-timeout=30

Here is the order from the directly bootstrapped sbd-diskless cluster  `sudo -E crm cluster init -yS -N adm@160-2`

INFO: crm sbd configure show property
have-watchdog=true
stonith-enabled=true
stonith-watchdog-timeout=30
stonith-timeout=71

zzhou1 (approved):
Approved with small cosmetic findings.

crmsh/bootstrap.py
liangxin1300: > Some findings to verify #2010
> 
> \1. cosmetic finding
> 
> > WARNING: "priority-fencing-delay" in crm_config is set to 0, it was 60
> 
> Nitpicking, to delete "priority-fencing-delay" to simplify CIB eventually since the default value is 0. The above output could be something like,
> 

I prefer to explicitly set it to 0 rather than deleting it.

> WARNING: "priority-fencing-delay" in crm_config is deleted, whose default is 0. It was 60
> 
> \2. yet another one, a little strange, the order of the output after `sbd` stage is different than the cluster directly bootstrapped as diskless. It makes more sense to have the stable order from the impression.
> 

That's because you might configure disk-based SBD from diskless, on that way, `stonith-watchdog-timeout` property will be deleted, and will be added when you configure diskless SBD from disk-based

> INFO: crm sbd configure show property have-watchdog=true stonith-enabled=true stonith-timeout=71 stonith-watchdog-timeout=30
> 
> Here is the order from the directly bootstrapped sbd-diskless cluster `sudo -E crm cluster init -yS -N adm@160-2`
> 
> INFO: crm sbd configure show property have-watchdog=true stonith-enabled=true stonith-watchdog-timeout=30 stonith-timeout=71

PR #2028, opened 2026-02-09
minieval.py : remove ast.Num and ast.Str no longer available in Python 3.14*

### LINKED ISSUE #2027: python3.14 changes ast elements breaking unittests
According to [1]

Remove the following classes, which have been deprecated aliases of [Constant](https://docs.python.org/3/library/ast.html#ast.Constant) since Python 3.8 and have emitted deprecation warnings since Python 3.12:

Bytes
Ellipsis
NameConstant
Num
Str

minieval.py is still using ast.Num and is broken for Python 3.14  

[1] https://docs.python.org/3/whatsnew/3.14.html#id9

hector-cao:
Fixes #2027

### DIFF
--- a/crmsh/minieval.py
+++ b/crmsh/minieval.py
@@ -183,8 +183,7 @@ class SimpleEval(object):  # pylint: disable=too-few-public-methods
         self.names = names
 
         self.nodes = {
-            ast.Num: self._eval_num,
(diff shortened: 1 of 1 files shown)


liangxin1300 (approved):
Thank you!

PR #2030, opened 2026-02-11
Dev: Replace terminology 'stonith' with 'fencing' (jsc#PED-14631)

liangxin1300:
Since pacemaker upstream has replaced the term 'stonith' with 'fencing', we should also update the terminology in crmsh to maintain consistency with pacemaker and avoid confusion for users.

Changes:

- Replace deprecated terminology across the entire codebase (e.g., use “fencing” instead of “stonith”)
- When querying, setting, or editing cluster properties, detect deprecated property names and emit appropriate warnings
- Preserve backward compatibility: crm cluster health sbd --fix must only fix invalid values and not rename deprecated property names
- Add a new functional test file deprecated_term.feature to cover most of the above changes
- Hide deprecated property names from completion output for crm configure get_property / crm configure property
- Keep the “stonith” term unchanged in RA classes

### DIFF
--- a/test/unittests/test_bootstrap.py
+++ b/test/unittests/test_bootstrap.py
@@ -1503,22 +1503,22 @@ done
 
     @mock.patch('crmsh.sbd.SBDConfigChecker')
     @mock.patch('crmsh.service_manager.ServiceManager.service_is_active')
-    def test_adjust_stonith_timeout_sbd(self, mock_is_active, mock_sbd_checker):
+    def test_adjust_fencing_timeout_sbd(self, mock_is_active, mock_sbd_checker):
         mock_sbd_checker_inst = mock.Mock()
         mock_sbd_checker.return_value = mock_sbd_checker_inst
         mock_sbd_checker_inst.check_and_fix = mock.Mock()
         mock_is_active.return_value = True
-        bootstrap.adjust_stonith_timeout()
+        bootstrap.adjust_fencing_timeout()
         mock_sbd_checker.assert_called_once_with(quiet=True, fix=True)
 
     @mock.patch('crmsh.utils.set_property')
-    @mock.patch('crmsh.bootstrap.get_stonith_timeout_generally_expected')
+    @mock.patch('crmsh.bootstrap.get_fencing_timeout_generally_expected')
     @mock.patch('crmsh.service_manager.ServiceManager.service_is_active')
-    def test_adjust_stonith_timeout(self, mock_is_active, mock_get_timeout, mock_set):
+    def test_adjust_fencing_timeout(self, mock_is_active, mock_get_timeout, mock_set):
         mock_is_active.return_value = False
         mock_get_timeout.return_value = 30
-        bootstrap.adjust_stonith_timeout()
-        mock_set.assert_called_once_with("stonith-timeout", 30, conditional=True)
+        bootstrap.adjust_fencing_timeout()
+        mock_set.assert_called_once_with("fencing-timeout", 30, conditional=True)
 
     @mock.patch('crmsh.utils.set_property')
     def test_adjust_priority_in_rsc_defaults_2node(self, mock_set):
@@ -1553,17 +1553,17 @@ done
     @mock.patch('crmsh.sbd.SBDManager.warn_diskless_sbd')
     @mock.patch('crmsh.bootstrap.adjust_priority_fencing_delay')
     @mock.patch('crmsh.bootstrap.adjust_priority_in_rsc_defaults')
-    @mock.patch('crmsh.bootstrap.adjust_stonith_timeout')
+    @mock.patch('crmsh.bootstrap.adjust_fencing_timeout')
     @mock.patch('crmsh.bootstrap.adjust_pcmk_delay_max')
     @mock.patch('crmsh.utils.is_2node_cluster_without_qdevice')
     @mock.patch('crmsh.service_manager.ServiceManager.service_is_active')
-    def test_adjust_properties(self, mock_is_active, mock_2node_qdevice, mock_adj_pcmk, mock_adj_stonith, mock_adj_priority, mock_adj_fence, mock_warn_sbd):
+    def test_adjust_properties(self, mock_is_active, mock_2node_qdevice, mock_adj_pcmk, mock_adj_fencing, mock_adj_priority, mock_adj_fence, mock_warn_sbd):
         mock_is_active.return_value = True
         mock_2node_qdevice.return_value = True
         bootstrap.adjust_properties()
         mock_is_active.assert_called_once_with("pacemaker.service")
         mock_adj_pcmk.assert_called_once_with(True)
-        mock_adj_stonith.assert_called_once_with()
+        mock_adj_fencing.assert_called_once_with()
         mock_adj_priority.assert_called_once_with(True)
         mock_adj_fence.assert_called_once_with(True)
         mock_warn_sbd.assert_called_once_with()
--- a/test/unittests/test_bugs.py
+++ b/test/unittests/test_bugs.py
@@ -172,8 +172,8 @@ def test_comments():
 
 def test_eq1():
     xml1 = """<cluster_property_set id="cib-bootstrap-options">
-    <nvpair id="cib-bootstrap-options-stonith-enabled" name="stonith-enabled" value="true"></nvpair>
-    <nvpair id="cib-bootstrap-options-stonith-timeout" name="stonith-timeout" value="180"></nvpair>
+    <nvpair id="cib-bootstrap-options-fencing-enabled" name="fencing-enabled" value="true"></nvpair>
+    <nvpair id="cib-bootstrap-options-fencing-timeout" name="fencing-timeout" value="180"></nvpair>
     <nvpair id="cib-bootstrap-options-symmetric-cluster" name="symmetric-cluster" value="false"></nvpair>
     <nvpair id="cib-bootstrap-options-no-quorum-policy" name="no-quorum-policy" value="freeze"></nvpair>
     <nvpair id="cib-bootstrap-options-batch-limit" name="batch-limit" value="20"></nvpair>
@@ -184,8 +184,8 @@ def test_eq1():
   </cluster_property_set>
   """
     xml2 = """<cluster_property_set id="cib-bootstrap-options">
-    <nvpair id="cib-bootstrap-options-stonith-enabled" name="stonith-enabled" value="true"></nvpair>
-    <nvpair id="cib-bootstrap-options-stonith-timeout" name="stonith-timeout" value="180"></nvpair>
+    <nvpair id="cib-bootstrap-options-fencing-enabled" name="fencing-enabled" value="true"></nvpair>
+    <nvpair id="cib-bootstrap-options-fencing-timeout" name="fencing-timeout" value="180"></nvpair>
     <nvpair id="cib-bootstrap-options-symmetric-cluster" name="symmetric-cluster" value="false"></nvpair>
     <nvpair id="cib-bootstrap-options-no-quorum-policy" name="no-quorum-policy" value="freeze"></nvpair>
     <nvpair id="cib-bootstrap-options-batch-limit" name="batch-limit" value="20"></nvpair>
@@ -285,25 +285,25 @@ def test_copy_nvpairs():
 
     to = etree.fromstring('''
     <node>
-    <nvpair name="stonith-enabled" value="true"/>
+    <nvpair name="fencing-enabled" value="true"/>
     </node>
     ''')
     copy_nvpairs(to, etree.fromstring('''
     <node>
-    <nvpair name="stonith-enabled" value="false"/>
+    <nvpair name="fencing-enabled" value="false"/>
     </node>
     '''))
 
-    assert ['stonith-enabled'] == to.xpath('./nvpair/@name')
+    assert ['fencing-enabled'] == to.xpath('./nvpair/@name')
     assert ['false'] == to.xpath('./nvpair/@value')
 
     copy_nvpairs(to, etree.fromstring('''
     <node>
-    <nvpair name="stonith-enabled" value="true"/>
+    <nvpair name="fencing-enabled" value="true"/>
     </node>
     '''))
 
-    assert ['stonith-enabled'] == to.xpath('./nvpair/@name')
+    assert ['fencing-enabled'] == to.xpath('./nvpair/@name')
     assert ['true'] == to.xpath('./nvpair/@value')
 
 
@@ -745,7 +745,7 @@ property cib-bootstrap-options: \
 	dc-version="1.1.13+git20150917.20c2178-224.2-1.1.13+git20150917.20c2178" \
 	cluster-infrastructure=corosync \
 	cluster-name=hacluster \
-	stonith-enabled=true \
+	fencing-enabled=true \
 	no-quorum-policy=ignore
 rsc_defaults rsc-options: \
 	resource-stickiness=1 \
@@ -773,9 +773,9 @@ def test_bug_110():
     """
     configuring attribute-based fencing-topology
     """
-    factory.create_object(*"primitive stonith-libvirt stonith:fence_sbd".split())
+    factory.create_object(*"primitive fencing-libvirt stonith:fence_sbd".split())
     factory.create_object(*"primitive fence-nova stonith:fence_sbd".split())
-    cmd = "fencing_topology attr:OpenStack-role=compute stonith-libvirt,fence-nova".split()
+    cmd = "fencing_topology attr:OpenStack-role=compute fencing-libvirt,fence-nova".split()
     ok = factory.create_object(*cmd)
     assert ok
     obj = cibconfig.mkset_obj()
--- a/test/unittests/test_cibquery.py
+++ b/test/unittests/test_cibquery.py
@@ -24,8 +24,8 @@ class TestCibQuery(unittest.TestCase):
         <nvpair id="cib-bootstrap-options-dc-version" name="dc-version" value="2.1.5+20221208.a3f44794f-150500.4.9-2.1.5+20221208.a3f44794f"/>
         <nvpair id="cib-bootstrap-options-cluster-infrastructure" name="cluster-infrastructure" value="corosync"/>
         <nvpair id="cib-bootstrap-options-cluster-name" name="cluster-name" value="hacluster"/>
-        <nvpair name="stonith-enabled" value="true" id="cib-bootstrap-options-stonith-enabled"/>
-        <nvpair id="cib-bootstrap-options-stonith-timeout" name="stonith-timeout" value="71"/>
+        <nvpair name="fencing-enabled" value="true" id="cib-bootstrap-options-fencing-enabled"/>
+        <nvpair id="cib-bootstrap-options-fencing-timeout" name="fencing-timeout" value="71"/>
       </cluster_property_set>
     </crm_config>
     <nodes>
@@ -40,11 +40,11 @@ class TestCibQuery(unittest.TestCase):
           <op name="monitor" interval="10" timeout="20" id="admin-ip-monitor-10"/>
         </operations>
       </primitive>
-      <primitive id="stonith-sbd" class="stonith" type="external/sbd">
+      <primitive id="fencing-sbd" class="stonith" type="external/sbd">
         <operations>
-          <op name="monitor" timeout="20" interval="3600" id="stonith-sbd-monitor-3600"/>
-          <op name="start" timeout="20" interval="0s" id="stonith-sbd-start-0s"/>
-          <op name="stop" timeout="15" interval="0s" id="stonith-sbd-stop-0s"/>
+          <op name="monitor" timeout="20" interval="3600" id="fencing-sbd-monitor-3600"/>
+          <op name="start" timeout="20" interval="0s" id="fencing-sbd-start-0s"/>
+          <op name="stop" timeout="15" interval="0s" id="fencing-sbd-stop-0s"/>
         </operations>
       </primitive>
       <clone id="ocfs2-clone">
--- a/test/unittests/test_cluster_fs.py
+++ b/test/unittests/test_cluster_fs.py
@@ -177,12 +177,12 @@ class TestClusterFSManager(unittest.TestCase):
         mock_confirm.assert_called_once_with("/dev/sda1 contains a ext4 file system - overwrite?")
         mock_cluster_shell_inst.get_stdout_or_raise_error.assert_called_once_with("wipefs -a /dev/sda1")
 
-    @mock.patch("crmsh.utils.has_stonith_running")
-    def test_init_verify_no_stonith(self, mock_has_stonith_running):
-        mock_has_stonith_running.return_value = False
+    @mock.patch("crmsh.utils.has_fence_device_registered")
+    def test_init_verify_no_fence_device(self, mock_has_fence_device_registered):
+        mock_has_fence_device_registered.return_value = False
         with self.assertRaises(cluster_fs.Error) as context:
             self.instance_ocfs2_stage_with_device.init_verify()
-        self.assertIn("OCFS2 requires stonith device configured and running", str(context.exception))
+        self.assertIn("OCFS2 requires fence device configured and running", str(context.exception))
 
     def test_gen_ra_scripts_unsupport_type(self):
         with self.assertRaises(cluster_fs.Error) as context:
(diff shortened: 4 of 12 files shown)


liangxin1300:
```
# crm configure show 
...
        fencing-enabled=false
...

# crm configure get_property stonith-enabled
WARNING: Querying the value of "stonith-enabled" but it is not configured, return the value of new name "fencing-enabled"
false

# crm configure get_property fencing-enabled
false
```

crmsh/utils.py
```diff
@@ -3400,4 +3414,55 @@ def able_to_restart_cluster(in_maintenance_mode: bool = False) -> bool:
         logger.warning("Understand risks that running RA has no cluster protection while the cluster is in maintenance mode and restarting")
         logger.info("Aborting the configuration change attempt")
         return False
+
+
+def check_deprecated_term(original_term: str, existing_xml_node = None) -> None:
```
zzhou1: Sounds, my impression would rely on the xml output of crm_attribute, instead of manually maintain the mapping table.

```
 adm@161-1:~> sudo crm_attribute -L cluster --all --output-as=xml

      <parameter name="stonith-enabled" advanced="1" generated="0">
        <deprecated>
          <replaced-with name="fencing-enabled"/>
        </deprecated>

      <parameter name="stonith-action" advanced="0" generated="0">
        <deprecated>
          <replaced-with name="fencing-action"/>
        </deprecated>

```

crmsh/utils.py
```diff
@@ -175,6 +176,20 @@ def get_cluster_option_metadata(show_xml=True) -> str:
     return None
 
 
+def get_deprected_cluster_option_dict() -> dict:
```
nicholasyang2022:
```suggestion
def get_deprecated_cluster_option_dict() -> dict:
```

crmsh/utils.py
```diff
@@ -3400,4 +3429,41 @@ def able_to_restart_cluster(in_maintenance_mode: bool = False) -> bool:
...
+
+    if existing_xml_node is not None:
+        deprecated_term_configured = existing_xml_node.find(f".//*[@name='{deprecated_term}']") is not None
+        new_term_configured = existing_xml_node.find(f".//*[@name='{new_term}']") is not None
+    else:
+        deprecated_term_configured = property_configured(deprecated_term)
+        new_term_configured = property_configured(new_term)
+
+    if deprecated_term_configured and new_term_configured:
+        if use_deprecated_term:
```
nicholasyang2022: Should we also warn when the new term is used?

crmsh/utils.py
zzhou1:
``` suggestion
def get_deprecated_cluster_option_with_replacement_dict() -> dict:
```

The reason is that

```
def get_deprecated_cluster_option_dict() -> dict:
```
could be used for the bigger scope to include several other deprecated options without "replaced_with", for example, "concurrent-fencing, enable-startup-probes, etc".  Ideally, it can be used to proactively provide WARN for UI(s)
```
crm configure get_property $any_deprecated_option
crm configure show
```

crmsh/utils.py
liangxin1300: - When querying
  - fencing-* property, just return value, no need to warn, since the user might not care about the deprecated one
  - stonith-* property, need to warn that this value will be ignored
- When setting/editing
  - No matter whether a new or deprecated name, this warning will always be triggered, because `sanity_check_nvpairs` will iterate over all nvpairs
  
The variable `use_new_term` seems useless this time, might be usable when needed to do auto-convert in the future

crmsh/utils.py
```diff
@@ -3400,4 +3429,41 @@ def able_to_restart_cluster(in_maintenance_mode: bool = False) -> bool:
...
+            logger.warning(
+                "The new property name \"%s\" is configured, the value of deprecated name \"%s\" will be ignored",
```
zzhou1: I prefer the concise and direct sentence, for example,

```suggestion
                "\"%s\" is configured; \"%s\" is deprecated and ignored",
```

crmsh/utils.py
```diff
@@ -3400,4 +3429,41 @@ def able_to_restart_cluster(in_maintenance_mode: bool = False) -> bool:
         logger.warning("Understand risks that running RA has no cluster protection while the cluster is in maintenance mode and restarting")
         logger.info("Aborting the configuration change attempt")
         return False
+
+
+def check_deprecated_term(original_term: str, existing_xml_node = None) -> None:
```
zzhou1: We could expand this function to handle more than just cases where both new and deprecated parameters are configured.

1. The WARN is expected for the migrated CIB where only deprecated properties are configured
    eg. `crm configure get_property stonith-timeout`  # where CIB has no `fencing-timeout`

2. The WARN is also expected for those deprecated properties without replacement
    eg.  `crm configure get_property enable-startup-probes`

crmsh/utils.py
liangxin1300: I considered the deprecated term without replacement term case;
So I changed back the name `get_deprecated_cluster_option_dict`

crmsh/utils.py
```diff
@@ -3400,4 +3431,57 @@ def able_to_restart_cluster(in_maintenance_mode: bool = False) -> bool:
...
+            if new_term:
+                logger.warning(
+                    "\"%s\" is deprecated, please consider using \"%s\"",
+                    deprecated_term, new_term
+                )
+            else:
+                logger.warning(
+                    "\"%s\" is deprecated, please consider removing it",
+                    deprecated_term
+                )
```
zzhou1: As long as the user try to access the deprecated property, the WARN is expected no matter it is in CIB or not. eg. `crm configure get_property stonith-timeout`


```suggestion
    elif use_deprecated_term:
        if new_term:
            logger.warning(
                "\"%s\" is deprecated, please consider using \"%s\"",
                deprecated_term, new_term
            )
        else:
            logger.warning(
                "\"%s\" is deprecated, please consider removing it",
                deprecated_term
            )
```


However, eg. `crm sbd configure show` need be revisited for the migrated installation from the old release, where the WARN is expected only if the deprecated are configured and has no replace_with in CIB. Basically, the above code suggestion will impact sbd.py > def _check_deprecated_property().

crmsh/crash_test/task.py
```diff
@@ -102,16 +102,16 @@ def task_pre_check(self, need_fence=True):
...
         if not ServiceManager().service_is_active("pacemaker.service"):
             raise TaskError("Cluster not running!")
         if need_fence:
             self.get_fence_info()
             if not self.fence_enabled:
-                raise TaskError("Require stonith enabled")
+                raise TaskError("Require fencing enabled")
             if not self.fence_configured:
-                raise TaskError("Require stonith device configured and running")
+                raise TaskError("Require fence device configured and running")
```
gao-yan, line 114: We could just be consistent with the replacement pattern "stonith- > fencing", so better as well call it "fencing device".

crmsh/crash_test/utils.py
```diff
@@ -108,27 +108,27 @@ class FenceInfo(object):
...
         if action_result is None or action_result not in supported_actions:
-            msg_error(f"Cluster property \"stonith-action\" should be {'|'.join(supported_actions)}")
+            msg_error(f"Cluster property \"fencing-action\" should be {'|'.join(supported_actions)}")
             return None
         return action_result
 
     @property
     def fence_timeout(self):
-        timeout_result = crmshutils.get_property("stonith-timeout")
+        timeout_result = crmshutils.get_property("fencing-timeout")
```
gao-yan, line 131: Will crash_test work correctly for an existing cluster setup with deprecated `stonith-*` options? Otherwise should we for instance implement and call an alternative version of "get_property" function for getting the practical value by recognizing both the old and new option names?

crmsh/sbd.py
```diff
@@ -184,7 +184,7 @@ def diskbased_sbd_configured() -> bool:
 
     @staticmethod
     def diskless_sbd_configured() -> bool:
-        value = utils.get_property("stonith-watchdog-timeout")
+        value = utils.get_property("fencing-watchdog-timeout")
```
gao-yan, line 187: Depending on possible contexts, we might want to detect if diskless SBD is configured with the deprecated option name?

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
                 return CheckResult.WARNING
         return CheckResult.SUCCESS
 
-    def _fix_stonith_watchdog_timeout(self):
+    def _fix_fencing_watchdog_timeout(self):
         if self.disk_based:
-            logger.info("Removing stonith-watchdog-timeout property")
-            utils.delete_property("stonith-watchdog-timeout")
+            logger.info("Removing fencing-watchdog-timeout property")
+            utils.delete_property("fencing-watchdog-timeout")
```
gao-yan: Likewise, crmsh should delete any possible leftover setting with "stonith-watchdog-timeout" as well. Similarly, maybe with an alternative "delete_property" function?

crmsh/bootstrap.py
```diff
@@ -2799,23 +2799,23 @@ def adjust_pcmk_delay_max(is_2node_wo_qdevice):
...
+    Adjust fencing-timeout for sbd and other scenarios
     """
     if ServiceManager().service_is_active(constants.SBD_SERVICE):
         sbd.SBDConfigChecker(quiet=True, fix=True).check_and_fix()
     else:
-        value = get_stonith_timeout_generally_expected()
+        value = get_fencing_timeout_generally_expected()
         if value:
-            utils.set_property("stonith-timeout", value, conditional=True)
+            utils.set_property("fencing-timeout", value, conditional=True)
```
gao-yan, line 2811: Wherever if it's crmsh that orchestrates the setup and decides to set a property by itself rather than by an admin, maybe it should meanwhile try deleting any possible setting with the corresponding deprecated option?

Similarly, maybe it could be achieved with an alternative version of "set_property" function.

crmsh/cluster_fs.py
```diff
@@ -189,8 +189,8 @@ def init_verify(self):
         """
         Verify after cluster running on init node
         """
-        if not utils.has_stonith_running():
-            msg = f"{self.type} requires stonith device configured and running." + self.error_hints_for_stage
+        if not utils.has_fence_device_registered():
+            msg = f"{self.type} requires fence device configured and running." + self.error_hints_for_stage
```
gao-yan, line 193: Likewise about "stonith -> fencing".

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
         bootstrap.sync_path(SBDManager.SBD_SYSTEMD_DELAY_START_DIR)
         utils.cluster_run_cmd("systemctl daemon-reload")
 
-    def _check_stonith_watchdog_timeout(self) -> CheckResult:
-        value = utils.get_property("stonith-watchdog-timeout")
+    def _check_fencing_watchdog_timeout(self) -> CheckResult:
+        value = utils.get_property("fencing-watchdog-timeout")
```
gao-yan: The context seems to be checking whether the configuration is existing and whether it's as expected. If not, fix it, right?

We need to consider what to do or what not if it's configured with the deprecated option.

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
-            utils.set_property("stonith-watchdog-timeout", self.stonith_watchdog_timeout)
+            logger.info("Adjusting fencing-watchdog-timeout to %d", self.fencing_watchdog_timeout)
+            utils.set_property("fencing-watchdog-timeout", self.fencing_watchdog_timeout)
 
-    def _check_stonith_timeout(self) -> CheckResult:
-        expected_value = self.get_stonith_timeout_expected()
-        value = utils.get_property("stonith-timeout")
+    def _check_fencing_timeout(self) -> CheckResult:
+        expected_value = self.get_fencing_timeout_expected()
+        value = utils.get_property("fencing-timeout")
```
gao-yan: Likewise.

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
         return CheckResult.SUCCESS
 
-    def _fix_stonith_timeout(self):
-        expected_value = self.get_stonith_timeout_expected()
-        logger.info("Adjusting stonith-timeout to %d", expected_value)
-        utils.set_property("stonith-timeout", expected_value)
+    def _fix_fencing_timeout(self):
+        expected_value = self.get_fencing_timeout_expected()
+        logger.info("Adjusting fencing-timeout to %d", expected_value)
+        utils.set_property("fencing-timeout", expected_value)
```
gao-yan: Likewise.

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
 
-    def _check_stonith_enabled(self) -> CheckResult:
-        value = utils.get_property("stonith-enabled", get_default=False)
+    def _check_fencing_enabled(self) -> CheckResult:
+        value = utils.get_property("fencing-enabled", get_default=False)
```
gao-yan: Likewise.

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
             )
             return CheckResult.ERROR
         return CheckResult.SUCCESS
 
-    def _fix_stonith_enabled(self):
-        logger.info("Setting stonith-enabled to true")
-        utils.set_property("stonith-enabled", "true")
+    def _fix_fencing_enabled(self):
+        logger.info("Setting fencing-enabled to true")
+        utils.set_property("fencing-enabled", "true")
```
gao-yan: Likewise.

crmsh/utils.py
```diff
@@ -3296,12 +3304,19 @@ def fuzzy_match(rx):
...
+        "fencing-timeout",
+        "priority-fencing-delay"
+    ]
+    for p in related_properties:
         if get_property(p, get_default=False):
             delete_property(p)
-    if get_property("stonith-enabled") == "true":
-        set_property("stonith-enabled", "false")
+    if get_property("fencing-enabled") == "true":
+        set_property("fencing-enabled", "false")
```
gao-yan, line 3347: Just noticed this. Why don't we unset/default "stonith-enabled" and "fence-enabled" anyway? We don't want users' clusters to risk missing proper fencing configuration and proceed resource management by ignoring any situations where fencing is needed. After all, any cluster with fencing disabled is not supported.

crmsh/qdevice.py
```diff
@@ -620,7 +620,7 @@ def adjust_sbd_watchdog_timeout_with_qdevice(self):
                 sbd_watchdog_timeout_qdevice = sbd.SBDTimeout.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
                 sbd.SBDManager.update_sbd_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout_qdevice)})
                 if self.is_stage:
-                    utils.set_property("stonith-watchdog-timeout", 2*sbd_watchdog_timeout_qdevice)
+                    utils.set_property("fencing-watchdog-timeout", 2*sbd_watchdog_timeout_qdevice)
```
gao-yan, line 623: Likewise.

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
         if self.disk_based:
-            logger.info("Removing stonith-watchdog-timeout property")
-            utils.delete_property("stonith-watchdog-timeout")
+            logger.info("Removing fencing-watchdog-timeout property")
+            utils.delete_property("fencing-watchdog-timeout")
         else:
-            logger.info("Adjusting stonith-watchdog-timeout to %d", self.stonith_watchdog_timeout)
-            utils.set_property("stonith-watchdog-timeout", self.stonith_watchdog_timeout)
+            logger.info("Adjusting fencing-watchdog-timeout to %d", self.fencing_watchdog_timeout)
+            utils.set_property("fencing-watchdog-timeout", self.fencing_watchdog_timeout)
```
gao-yan: Likewise

crmsh/bootstrap.py
```diff
@@ -2763,18 +2763,18 @@ def bootstrap_arbitrator(context):
...
-    Adjust stonith-timeout for all scenarios, formula is:
+    Adjust fencing-timeout for all scenarios, formula is:
 
-    stonith-timeout = STONITH_TIMEOUT_DEFAULT + token + consensus
+    fencing-timeout = FENCING_TIMEOUT_DEFAULT + token + consensus
     """
-    stonith_enabled = utils.get_property("stonith-enabled")
-    # When stonith disabled, return
-    if utils.is_boolean_false(stonith_enabled):
+    fencing_enabled = utils.get_property("fencing-enabled")
```
gao-yan, line 2772: Likewise.

crmsh/crash_test/task.py
liangxin1300: I use 'fence device' because the output of stonith_admin
```
# stonith_admin -L
fencing-sbd
1 fence device found
```
Keep the same with stonith_admin, it is also no harm

crmsh/crash_test/utils.py
liangxin1300: Changed like:
```
crm(live/sle16-1)cluster# crash_test --fence-node sle16-2
WARNING: "fencing-timeout" is not configured; use configured deprecated "stonith-timeout"
```
Give a warning and return the value of the old name, if the new term is not configured, but the old one is configured

I put the `translate` logic into `utils.get_property` function, only do `translate` job when querying new name, but only deprecated name configured, then return the value of the deprecated name and give a warning

crmsh/sbd.py
liangxin1300: See the above change, 
>Give a warning and return the value of the old name, if the new term is not configured, but the old one is configured

crmsh/sbd.py
liangxin1300: I will make such change when deleting a new name but only the deprecated name configured:
```
INFO: Removing fencing-watchdog-timeout property
WARNING: "fencing-watchdog-timeout" is not configured; use configured deprecated "stonith-watchdog-timeout"
INFO: Removing stonith-watchdog-timeout property
```

crmsh/cluster_fs.py
liangxin1300: Like I said before the output of stonith_admin

crmsh/crash_test/task.py
gao-yan: Well, as a matter of fact, pacemaker itself is not consistent about this :-), which may eventually need to be unified... As well taking stonith_admin as an example:

```
stonith_admin --help-all | grep -i "fencing device"
  -R, --register=DEVICE             Register the named fencing device. Requires: --agent.
  -D, --deregister=DEVICE           De-register the named fencing device.
```

crmsh/bootstrap.py
liangxin1300: I will add the logic while calling set_property:
```
    deprecated_inst = DeprecatedTermTranslator(property_name)
    if deprecated_inst.both_configured():
        delete_property(deprecated_inst.deprecated_term)
```
After setting the property, if both deprecated and new properties are detected configured, delete the deprecated one

crmsh/sbd.py
liangxin1300: Like I said before

I put the translate logic into utils.get_property function, only do translate job when querying new name, but only deprecated name configured, then return the value of the deprecated name and give a warning

If the value of the deprecated name doesn't meet the expectation, then fix function(triggered by --fix option) will call set_property to set the name name, and fininally, the deprecated one will be removed

crmsh/utils.py
liangxin1300: This is only called by [purge_sbd_from_cluster](https://github.com/ClusterLabs/crmsh/blob/master/crmsh/sbd.py#L1343-L1366) will only be triggered when sbd is the last registered fence device
And purge_sbd_from_cluster will be called under maintenance mode, see https://github.com/ClusterLabs/crmsh/blob/master/crmsh/ui_sbd.py#L630-L642

So, that's why here I set fencing-enabled as false

crmsh/sbd.py
gao-yan: What if anyhow both the new and deprecated names are configured? We'll need to delete both. And in order to prevent the deprecated one from getting any chance to take an effect, we need to delete the deprecated one first and then the new one.

crmsh/utils.py
gao-yan: But that'll exactly leave a `fencing-enabled=false` setting, so the cluster will get into a risky and unsupported situation, right?

Be aware, no matter if there's any fencing resource configured or not,  as long as fencing-enabled is false, whenever there's a bad situation where it requires a fencing, pacemaker will just ignore the bad situation and proceed resource failover, hence concurrency violation, data corruption, etc...

So fencing-enabled=false should better never be used.

crmsh/sbd.py
liangxin1300: Changed:
```
# crm cluster health sbd --fix
INFO: Removing fencing-watchdog-timeout property
INFO: Delete cluster property "stonith-watchdog-timeout" in crm_config
INFO: Delete cluster property "fencing-watchdog-timeout" in crm_config
```

crmsh/crash_test/task.py
liangxin1300: Let's keep this unchanged for now, since the function utils.has_fence_device_registered relies on the output of stonith_admin
```
def has_fence_device_registered():
    """
    Check if any fence device registered
    """
    out = sh.cluster_shell().get_stdout_or_raise_error("stonith_admin -L")
    has_fence_device = re.search("[1-9]+ fence device[s]* found", out) is not None
    using_diskless_sbd = sbd.SBDUtils.is_using_diskless_sbd()
    return has_fence_device or using_diskless_sbd
```

crmsh/utils.py
liangxin1300: I already raised an issue #2057  for this. I will fix this in another PR

crmsh/sbd.py
```diff
@@ -821,94 +830,94 @@ def _fix_sbd_systemd_start_timeout(self):
...
+    def _fix_fencing_timeout(self):
+        expected_value = self.get_fencing_timeout_expected()
+        logger.info("Adjusting fencing-timeout to %d", expected_value)
+        utils.set_property("fencing-timeout", expected_value)
 
-    def _check_stonith_enabled(self) -> CheckResult:
-        value = utils.get_property("stonith-enabled", get_default=False)
+    def _check_fencing_enabled(self) -> CheckResult:
+        value = utils.get_property("fencing-enabled", get_default=False)
         if utils.is_boolean_false(value):
```
gao-yan: The `fencing-enabled`/`stonith-enabled` option defaults to true, meaning even if not setting them, it's still "true".

liangxin1300: The parameter `get_default=False` means this function will get the value from CIB by using `crm_attribute`
Then the condition should be changed like:
```suggestion
        if value == "false":
```
or 
if value and utils.is_boolean_false(value)
That's means `fencing-enabled` really been set as `false`

gao-yan: Right. Actually I don't understand why it has to use  `get_default=False` for this `get_property()` call first of all.

crmsh/sbd.py
```diff
@@ -821,94 +837,105 @@ def _fix_sbd_systemd_start_timeout(self):
...
+            logger.info("Removing %s property", current_term)
+            utils.delete_property(current_term)
         else:
-            logger.info("Adjusting stonith-watchdog-timeout to %d", self.stonith_watchdog_timeout)
-            utils.set_property("stonith-watchdog-timeout", self.stonith_watchdog_timeout)
+            logger.info(
+                "Adjusting %s to %d",
+                current_term, self.fencing_watchdog_timeout
+            )
+            utils.set_property(current_term, self.fencing_watchdog_timeout)
```
gao-yan, line 893: Does this mean crmsh may still set the deprecated option to fix the config now?

liangxin1300: Yes, to avoid any possible compatibility issue, `current_term` here might be a deprecated name
e.g., the user only configured a deprecated name with an invalid value, then `crm cluster health sbd --fix` will only fix the value, keep the deprecated name unchanged

gao-yan: Alright, just wanted to know if it was intentional since the logic about this seemed to have changed back and forth. 

I assume people might have different opinions about what/how a"--fix" option should fix, for example whether it should as well fix uses of deprecated options. But I can tell either way gets its point, so it should be fine this way at least for now.

PR #2033, opened 2026-02-25
Dev: spec: create dirs in /var with tmpfiles.d (jsc#PED-14865)

nicholasyang2022:


liangxin1300 (approved):
LGTM

Thanks!

nicholasyang2022:
We should have used "squash merge" to eliminate the log about "applying suggestions".

Issue #2039, opened 2026-03-02
`cluster remove --qdevice` should be rejected on the last standing node with qdevice for the 2-node diskless sbd cluster


zzhou1:
As the title, if the peer node cluster stack is stopped for a 2-node diskless sbd cluster, `cluster remove --qdevice`  should be rejected. Well, this is not the case not for the 2-node disk-based sbd cluster. 

Here is the reproducer,
```
adm@160-1:~> sudo -E crm -F cluster init -yS --qnetd-hostname adm@160-q -N adm@160-2

adm@160-2:~> sudo crm cluster stop

adm@160-1:~> sudo -E crm -F cluster remove --qdevice
```

Basically, even the single diskless sbd node is in two_node=1 stay quorate, and  will run into this
```
sbd[1751]:    cluster:  warning: set_servant_health: Connected to corosync but requires both nodes present
...
sbd[1748]:  warning: inquisitor_child: Latency: No liveness for 15s exceeds watchdog warning timeout of 9s (healthy servants: 0)Eventually get rebooted
```

REF:  https://github.com/ClusterLabs/crmsh/pull/2014 , https://github.com/ClusterLabs/crmsh/pull/741

PR #2041, opened 2026-03-03
Dev: bootstrap: Reject to remove qdevice when it will cause quorum lost

liangxin1300:
`crm cluster remove --qdevice` command will be rejected if it will cause quorum lost for the cluster configured with diskless SBD, since the diskless SBD will self-fence the node without quorum

Fix issue:
- #2039

### DIFF
--- a/test/unittests/test_qdevice.py
+++ b/test/unittests/test_qdevice.py
@@ -22,12 +22,14 @@ def test_evaluate_qdevice_quorum_effect_reload(mock_get_dict, mock_quorate):
     mock_quorate.assert_called_once_with(3, 2)
 
 
+@mock.patch('crmsh.corosync.get_value')
 @mock.patch('crmsh.xmlutil.CrmMonXmlParser')
 @mock.patch('crmsh.utils.calculate_quorate_status')
 @mock.patch('crmsh.utils.get_quorum_votes_dict')
-def test_evaluate_qdevice_quorum_effect_later(mock_get_dict, mock_quorate, mock_parser):
+def test_evaluate_qdevice_quorum_effect_later(mock_get_dict, mock_quorate, mock_parser, mock_get_value):
     mock_get_dict.return_value = {'Expected': '2', 'Total': '2'}
     mock_quorate.return_value = False
+    mock_get_value.return_value = '1'
     mock_parser().is_non_stonith_resource_running.return_value = True
     res = qdevice.evaluate_qdevice_quorum_effect(qdevice.QDEVICE_REMOVE)
     assert res == qdevice.QdevicePolicy.QDEVICE_RESTART_LATER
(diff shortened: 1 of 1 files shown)


crmsh/qdevice.py
```diff
@@ -44,10 +45,16 @@ def evaluate_qdevice_quorum_effect(mode):
...
+        vote = corosync.get_value("quorum.device.votes") or 1
+        actual_votes -= int(vote)
     diskless_sbd = sbd.SBDUtils.is_using_diskless_sbd()
 
-    if utils.calculate_quorate_status(expected_votes, actual_votes) and not diskless_sbd:
+    quorate = utils.calculate_quorate_status(expected_votes, actual_votes)
+    if not quorate and diskless_sbd and mode == QDEVICE_REMOVE:
+        # Reject to remove qdevice, as it will lose quorum
+        # then diskless SBD will self-fence the node
+        return QdevicePolicy.QDEVICE_REMOVE_REJECT
```
zzhou1, line 56:
```suggestion pseudo code and the revised comments
    qdevice_votes = pseudo.code.to.get.corosync.conf.quorum.device.votes

    quorate = utils.calculate_quorate_status(expected_votes-qdevice_votes, actual_votes-qdevice_votes)
    if not quorate and diskless_sbd and mode == QDEVICE_REMOVE:
        # Rejected since it would result in a malfunctioning cluster lacking quorum
        # either the disk-based sbd cluster or the diskless sbd cluster
```

crmsh/bootstrap.py
```diff
@@ -2425,6 +2425,8 @@ def remove_qdevice() -> None:
 
     utils.check_all_nodes_reachable("removing QDevice from the cluster")
     qdevice_reload_policy = qdevice.evaluate_qdevice_quorum_effect(qdevice.QDEVICE_REMOVE)
+    if qdevice_reload_policy == qdevice.QdevicePolicy.QDEVICE_REMOVE_REJECT:
+        utils.fatal("Can't remove QDevice since the quorum will be lost for diskless SBD cluster")
```
zzhou1, line 2429:
```suggestion
        utils.fatal("operation is rejected since it would result in a malfunctioning cluster lacking quorum")
```

gao-yan: It wouldn't necessarily become "malfunctioning" purely based on the logic here :-) It's just that the next cluster startup will require an actual majority or "wait_for_all" to become first quorate.

What I can tell about the case in here is, an user may want to remove qdevice when not all cluster nodes are online, right? But actually it's technically feasible no matter what the quorum status is like, as long as the offline nodes are still reachable through ssh.

The actual problem is, after removal of qdevice, if they/we still want to only bring up the same subset of nodes and expect them to reach the status as before the removal of qdevice right away, it won't be in the same condition any more though. But it's actually about how quorum is first granted on cluster startup, which either requires an actual majority or "wait_for_all" (automatically enabled when "two_node" is enabled). But that's reasonable, even if qdevice is not in place.

So to me, this could be a case where we'd rather tell user "I can remove the on-disk configuration of qdevice for you, but you'll either want to restart cluster later manually by yourself, otherwise I can restart the cluster for you now but bring up all the nodes".

BTW, it's really confusing to still call it QDEVICE_RESTART_LATER when it's leveraging maintenance mode and actually restarting the cluster.  The name QDEVICE_RESTART_LATER should probably be preserved for the situation where it really requires users to manually restart their cluster by themselves.

After all, as said, to determine whether a cluster would become malfunctioning, the logic in "warn_diskless_sbd()" and the warning there tell the key:

"Diskless SBD requires cluster with three or more nodes. If you want to use diskless SBD for 2-node cluster, should be combined with QDevice."

PR #2056, opened 2026-03-19
[crmsh-4.5] Fix: corosync: Reuse nodeid from cmap to avoid nodeid and ip address mix-up (bsc#1259683)

nicholasyang2022:
When a node is removed and then added back, crmsh should reuse the original nodeid if it's still present in corosync's runtime state (cmap). Reusing a different node's previous nodeid can cause corosync to incorrectly map IP addresses, leading to unexpected node removals or fencing when a node is later removed.

- Add `remote` parameter support to `get_nodeinfo_from_cmaptool`, `add_node_ucast`, and related functions, to retrieve nodeinfo from seed host during joining
- Add get_nodeid_by_ip_from_cmap() to lookup nodeid in cmap
- Update get_free_nodeid() to exclude nodeids currently in cmap
- Update add_node_ucast() and add_node() to reuse nodeid from cmap
- Add support for remote cmap lookup in utils.get_nodeinfo_from_cmaptool()
- Rename find_configured_ip to assert_ip_not_configured for clarity
- Add unit tests to verify the new logic and remote lookup functionality

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -2091,7 +2091,7 @@ def join_cluster(seed_host, remote_user):
                 break
         invoke("rm -f /var/lib/heartbeat/crm/* /var/lib/pacemaker/cib/*")
         try:
-            corosync.add_node_ucast(ringXaddr_res)
+            corosync.add_node_ucast(ringXaddr_res, remote=seed_host)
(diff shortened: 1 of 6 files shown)


nicholasyang2022:
Fixed in #2061.

PR #1803, opened 2025-05-23
[crmsh-4.6] Dev: pre-migration: run cib check in local mode (jsc#PED-11808)

nicholasyang2022:

PR #1787, opened 2025-05-15
Merge branch '4.6-pre-migration-staging' into crmsh-4.6

nicholasyang2022:

PR #2075, opened 2026-04-09
[crmsh-4.5] Dev: pre-migration: backport pre-migration checks from crmsh-4.6 (jsc#PED-16076)

nicholasyang2022:
- backport pre-migration checks from crmsh-4.6:
  * #1787
  * #1803
  * 4c02d4b19314403c06526799cb3d2800c8af0d03
  * f07fa1dd7236bc989e8d8273cd9be7ca85727e11
- backport some other commits to make `crm configure show` available in offline mode:
  * 64ac0c89: Fix: cibconfig: do not load cib to check if cib is changed (#1466) — Prevents redundant CIB loading attempts when checking for changes.                                                                                  
  * e720df0e: Fix: cli: loading cib should be a soft depenedency — Modifies the requires() method to allow entering configuration levels without a running cluster.                                                                    
  * efcdf72b: Dev: utils: Load CIB_file env before some readonly commands — Implements load_cib_file_env() to set the CIB_file environment variable, enabling cibadmin to read from the local CIB file when the service is inactive.
- decrease the lower boundary of requried pacemaker version


liangxin1300 (approved):
LGTM

Thanks!

Issue #2092, opened 2026-04-20
--qdevice-port should influence the qnetd server /etc/sysconfig/corosync-qnetd


zzhou1:
Besides the title, the health check like functionality should consider if the port is same between qnetd sysconfig and corosync.conf on the local cluster. 

Extending this a bit, to add a new cluster to the qnetd server should use the existing qnetd port to config corosync.conf for qdevice. And, also to reject if --qdevice-port  is different than the existing one.

liangxin1300:
Yes, should be reject

liangxin1300:
For now, crmsh doesn't change /etc/sysconfig/corosync-qnetd, why should we do that?
When adding qdevice, we can just run `ss -ltnp | grep -i corosync-qnetd` on the qnetd server to get the existing port

zzhou1:
> For now, crmsh doesn't change /etc/sysconfig/corosync-qnetd, why should we do that?

Technically, this should be done to configure and start qnetd for the first time, given that  --qdevice-port is used.

liangxin1300:
Demo:
```
# success case
# crm cluster init qdevice --qnetd-hostname server1 -y
...
INFO: Configure Qdevice/Qnetd:
INFO: Update COROSYNC_QNETD_OPTIONS="-p 5403" to /etc/sysconfig/corosync-qnetd on server1
...

# conflict case
# crm cluster init qdevice --qnetd-hostname server1 --qdevice-port 5402 -y
...
INFO: Configure Qdevice/Qnetd:
ERROR: cluster.init: The port 5402 is different from the port 5403 in /etc/sysconfig/corosync-qnetd on server1
Please use '--qdevice-port 5403' to keep consistent

# corner case, corosync-qnetd is running(maybe is providing vote for other cluster)
# crm cluster init qdevice --qnetd-hostname server1 --qdevice-port 5400 -y
...
INFO: Configure Qdevice/Qnetd:
INFO: Update COROSYNC_QNETD_OPTIONS="-p 5400" to /etc/sysconfig/corosync-qnetd on server1
WARNING: Please restart corosync-qnetd.service on server1 for the new port to take effect
...
# need to warn user to restart corosync-qnetd.service manually

# corner case, corosync-qnetd is running(maybe is providing vote for other cluster) with the option "-m 10"
# crm cluster init qdevice --qnetd-hostname server1 --qdevice-port 5400 -y
...
INFO: Configure Qdevice/Qnetd:
[...]

PR #2094, opened 2026-04-22
Dev: qdevice: Configure "-p <port>" for /etc/sysconfig/corosync-qnetd on qnetd server

liangxin1300:
Changes include:
- Configure "-p <port>" for /etc/sysconfig/corosync-qnetd on qnetd server if needed
- Raise an error if the given port is different from the port that the qnetd server is using
- Add --qnetd-port option and deprecate --qdevice-port option
- Refactor qdevice options validation

Fix the issue:
- #2092

### DIFF
--- a/crmsh/qdevice.py
+++ b/crmsh/qdevice.py
@@ -212,9 +212,9 @@ class QDevice(object):
 
 
     @staticmethod
-    def check_qdevice_port(qdevice_port):
-        if qdevice_port and not utils.valid_port(qdevice_port):
-            raise ValueError("invalid qdevice port range(1024 - 65535)")
+    def check_qnetd_port(qnetd_port):
+        if qnetd_port and not utils.valid_port(qnetd_port):
+            raise ValueError("invalid qnetd port range(1024 - 65535)")
 
     @staticmethod
     def check_qdevice_algo(qdevice_algo):
@@ -262,7 +262,7 @@ class QDevice(object):
             utils.check_all_nodes_reachable("setup Qdevice")
         self.check_corosync_qdevice_available()
         self.check_qnetd_addr(self.qnetd_addr)
-        self.check_qdevice_port(self.port)
+        self.check_qnetd_port(self.port)
         self.check_qdevice_algo(self.algo)
         self.check_qdevice_tie_breaker(self.tie_breaker)
         self.check_qdevice_tls(self.tls)
@@ -526,7 +526,7 @@ class QDevice(object):
             port_in_qnetd = int(res.group(1))
             if self.port is not None and self.port != port_in_qnetd:
                 error_msg = f"The port {self.port} is different from the port {port_in_qnetd} that corosync-qnetd is using"
-                suggestion_msg = f"Please use '--qdevice-port {port_in_qnetd}' to keep consistent"
+                suggestion_msg = f"Please use '--qnetd-port {port_in_qnetd}' to keep consistent"
                 raise ValueError(f"{error_msg}\n{suggestion_msg}")
             else:
                 self.port = port_in_qnetd
@@ -555,7 +555,7 @@ class QDevice(object):
                     port_in_sysconfig = int(res.group(1))
                     if self.port is not None and self.port != port_in_sysconfig:
                         error_msg = f"The port {self.port} is different from the port {port_in_sysconfig} in {self.SYSCONFIG_QNETD}"
-                        suggestion_msg = f"Please use '--qdevice-port {port_in_sysconfig}' to keep consistent"
+                        suggestion_msg = f"Please use '--qnetd-port {port_in_sysconfig}' to keep consistent"
                         raise ValueError(f"{error_msg}\n{suggestion_msg}")
                     else:
                         self.port = port_in_sysconfig
--- a/crmsh/qdevice.py
+++ b/crmsh/qdevice.py
@@ -105,12 +105,16 @@ def get_node_list(is_stage: bool) -> list[str]:
         return [me]
 
 
+QNETD_DEFAULT_PORT = 5403
+
+
 class QDevice(object):
     """Class to manage qdevice configuration and services
 
     Call `certificate_process_on_init` to generate all of CA, server, and client certs.
     """
 
+    SYSCONFIG_QNETD = "/etc/sysconfig/corosync-qnetd"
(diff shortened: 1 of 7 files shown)


zzhou1:
https://github.com/ClusterLabs/crmsh/issues/2092#issuecomment-4294524858

The better UX in my view is go further steps to align the qdevice default to the qnetd port. Report the error if the specified --qdevice-port is different than qnetd port.

Currently, rejection ERROR message is kind of too early. The user has to add one more step to use qnetd port, that is unnecessary.

crmsh/qdevice.py
```diff
@@ -506,10 +512,49 @@ def remove_certification_files_on_qnetd(cls) -> None:
         cmd = "test -f {crq_file} && rm -f {crq_file}".format(crq_file=cls_inst.qdevice_crq_on_qnetd)
         shell.get_stdout_or_raise_error(cmd, qnetd_host)
 
+    def config_qnetd_port_in_sysconfig(self):
+        shell = sh.cluster_shell()
+
+        cmd = f"test -f {self.SYSCONFIG_QNETD}"
+        rc, _, _ = shell.get_rc_stdout_stderr_without_input(self.qnetd_addr, cmd)
+        if rc != 0:
```
zzhou1, line 541: Sounds to me, the better UX is to create the sysconfig corosync-qnetd if it doesn't exist, especially when the user wants to set a different qnetd port.

nicholasyang2022:
The option name `--qdevice-port` is strange. `corosync-qdevice` does not need a listening port.

crmsh/qdevice.py
```diff
@@ -506,10 +512,49 @@ def remove_certification_files_on_qnetd(cls) -> None:
...
+            return
+
+        cmd = f"grep COROSYNC_QNETD_OPTIONS= {self.SYSCONFIG_QNETD}"
```
zzhou1: Ideally, the code can be helped by AI code assistant to handle the use case scenario, for example,  to ignore `COROSYNC_QNETD_OPTIONS=` in  the comment lines.

liangxin1300:
> The option name `--qdevice-port` is strange. `corosync-qdevice` does not need a listening port.

I know, but for back compatibility, we should keep it unchanged

crmsh/qdevice.py
```diff
@@ -506,10 +512,49 @@ def remove_certification_files_on_qnetd(cls) -> None:
         cmd = "test -f {crq_file} && rm -f {crq_file}".format(crq_file=cls_inst.qdevice_crq_on_qnetd)
         shell.get_stdout_or_raise_error(cmd, qnetd_host)
 
+    def config_qnetd_port_in_sysconfig(self):
```
zzhou1: Thinking about when can we populate a specific different qnetd port to sysconfig.

1. The first scenario is a bran new installation, where `qnetd/nssdb` doesn't exist at all.
2. The 2nd scenarios is the user intends to regenerating `qnetd/nssdb` because of the security concern, for example. If need, we could probably leverage the `--force` option as UI to support this scenario.

To extend this to the multi-cluster context, this means to reject changing the different qnetd port on purpose as long as `nssdb` does exist, since it likely to break other running cluster's qdevice configuration.

zzhou1:
> > The option name `--qdevice-port` is strange. `corosync-qdevice` does not need a listening port.
> 
> I know, but for back compatibility, we should keep it unchanged

Maybe this is the moment to WARN `--qdevice-port` is deprecated and replaced by `--qnet-port`. Drop it when we drop "stonith-enabled=" together in the next major release, that will be a couple of years to come.

crmsh/qdevice.py
liangxin1300: I’m not sure I understand—could you explain a bit more?

crmsh/ui_cluster.py
```diff
@@ -459,13 +459,15 @@ def do_init(self, context, *args):
...
         qdevice_group.add_argument("--qnetd-hostname", dest="qnetd_addr_input", metavar="[USER@]HOST",
                                    help="User and host of the QNetd server. The host can be specified in either hostname or IP address.")
-        qdevice_group.add_argument("--qdevice-port", dest="qdevice_port", metavar="PORT", type=int, default=5403,
-                                   help="TCP PORT of QNetd server (default:5403)")
-        qdevice_group.add_argument("--qdevice-algo", dest="qdevice_algo", metavar="ALGORITHM", default="ffsplit", choices=['ffsplit', 'lms'],
+        qdevice_group.add_argument("--qdevice-port", dest="qdevice_port", metavar="PORT", type=int,
+                                   help=f"TCP PORT of QNetd server (default:{qdevice.QNETD_DEFAULT_PORT}, deprecated, use --qnetd-port instead)")
+        qdevice_group.add_argument("--qnetd-port", dest="qnetd_port", metavar="PORT", type=int,
+                                   help=f"TCP PORT of QNetd server (default:{qdevice.QNETD_DEFAULT_PORT})")
```
nicholasyang2022: `--qdevice-port` and `--qnetd-port` need to be mutual exclusive, or be validated to have the same value.

crmsh/qdevice.py
```diff
@@ -105,12 +105,16 @@ def get_node_list(is_stage: bool) -> list[str]:
         return [me]
 
 
+QNETD_DEFAULT_PORT = 5403
```
nicholasyang2022: This should either be a class-scoped constant or be placed at the beginning of the file.

crmsh/ui_cluster.py
liangxin1300: raise "Options --qdevice-port and --qnetd-port can't be used together"

nicholasyang2022 (commented):
The PR looks good overall, but I have a couple of suggestions for the tests:

1. **Mutual Exclusion Test**: The logic to prevent using both `--qdevice-port` and `--qnetd-port` is implemented in `crmsh/bootstrap.py`, but it doesn't seem to be covered by any tests. It would be good to add a test case for this.
2. **Unit Test Updates**: Some unit tests in `test/unittests/test_bootstrap.py` (e.g., around lines 116 and 125) are still using `ctx.qdevice_port`. While this works due to backward compatibility, it would be cleaner to update them to use `ctx.qnetd_port` to reflect the new preferred usage.

nicholasyang2022 (approved):

PR #2098, opened 2026-04-23
Dev: qdevice: Improve local address check for qnetd (bsc#1262094)

nicholasyang2022:
- Refactor check_qnetd_addr to use ipaddress and iproute2 for local address check.
- Use 'ip -j addr show' to get all local IP addresses.
- Support both IPv4 and IPv6 check.
- Update unit tests for check_qnetd_addr.

### DIFF
--- a/crmsh/qdevice.py
+++ b/crmsh/qdevice.py
@@ -1,3 +1,5 @@
+import ipaddress
+import json
 import os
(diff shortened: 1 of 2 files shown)


liangxin1300 (approved):
Thanks!

PR #2110, opened 2026-05-09
Dev: Leverage maintenance mode while removing node from cluster

liangxin1300:


crmsh/bootstrap.py
```diff
@@ -2522,7 +2522,10 @@ def bootstrap_remove(context):
     else:
         configured_nodes = xmlutil.CrmMonXmlParser().get_node_list()
         if cluster_node in configured_nodes:
-            remove_node_from_cluster(cluster_node)
+            with utils.leverage_maintenance_mode() as enabled:
+                if not utils.able_to_restart_cluster(enabled):
+                    return
```
nicholasyang2022, line 2527: Are you going to abort the removal process if it is not possible to restart the cluster?

crmsh/bootstrap.py
```diff
@@ -2522,7 +2522,10 @@ def bootstrap_remove(context):
     else:
         configured_nodes = xmlutil.CrmMonXmlParser().get_node_list()
         if cluster_node in configured_nodes:
-            remove_node_from_cluster(cluster_node)
+            with utils.leverage_maintenance_mode() as enabled:
```
nicholasyang2022, line 2525: Is it safe to remove a node in maintenance mode? What will happen if some resources are running on the node to be removed?

crmsh/bootstrap.py
```diff
@@ -2522,7 +2522,10 @@ def bootstrap_remove(context):
...
+                if not utils.able_to_restart_cluster(enabled):
```
nicholasyang2022, line 2526: I think you should check if restarting the cluster is possible before entering maintenance mode, or you will enter and exit maintenance mode without doing anything.

crmsh/bootstrap.py
liangxin1300: Yeah, that sounds in danger

PR #2113, opened 2026-05-12
Dev: sbd: Give a warning and return True when need to skip check and fix

liangxin1300:


nicholasyang2022 (commented):
It is good to return success without running the checks when SBD it not configured. But should we return success when SBD is configured and the checks are skipped?

liangxin1300:
It's enough to raise utils.TerminateSubCommand, no need to create a new class

crmsh/bootstrap.py
```diff
@@ -258,7 +258,8 @@ def _validate_sbd_option(self):
             if not with_sbd_option and self.yes_to_all:
                 utils.fatal("Stage sbd should specify sbd device by -s or diskless sbd by -S option")
             if ServiceManager().service_is_active(constants.SBD_SERVICE) and not crmsh.options.force:
-                utils.fatal("Can't configure stage sbd: sbd.service already running! Please use crm option '-F' if need to redeploy")
+                logger.warning("Can't configure stage sbd: sbd.service already running! Please use crm option '-F' if need to redeploy")
+                raise utils.TerminateSubCommand(success=True)
```
nicholasyang2022, line 262: This change is not related to the topic of this pull request.

crmsh/sbd.py
```diff
@@ -667,9 +667,10 @@ def _get_current_terms(self):
     def check_and_fix(self) -> CheckResult:
         if not ServiceManager().service_is_active(constants.SBD_SERVICE):
             if self.fix:
-                raise FixAborted("%s is not active, skip fixing SBD timeout issues" % constants.SBD_SERVICE)
+                raise FixAborted(f"{constants.SBD_SERVICE} is not active, skip fixing SBD timeout issues")
             elif not SBDUtils.diskbased_sbd_configured() and not SBDUtils.diskless_sbd_configured():
-                raise FixAborted("Neither disk-based nor disk-less SBD is configured, skip checking SBD timeout issues")
+                logger.warning("Neither disk-based nor diskless SBD is configured, skip checking SBD timeout issues")
+                raise utils.TerminateSubCommand(success=True)
```
nicholasyang2022, line 673: I think we should check whether SBD is configured prior to checking flag `self.fix`. If it is not configured, there is nothing needing fix, and *fix* should return success.

Issue #2117, opened 2026-05-26
SBD: Should detect whether the given device exists on peer node


liangxin1300:
- Case 1: 'sbd device add' success even the given device does not exist on the peer node
```
# /dev/sdb1 not exists on peer
# crm sbd device add /dev/sdb1
INFO: Configured sbd devices: /dev/sda5
INFO: Append devices: /dev/sdb1
INFO: Configuring disk-based SBD
INFO: Initializing SBD device /dev/sdb1
INFO: Update SBD_DEVICE in /etc/sysconfig/sbd: /dev/sda5;/dev/sdb1
INFO: Sync file /etc/sysconfig/sbd to sle16-2
INFO: Restarting cluster service
INFO: BEGIN Waiting for cluster
...........                                                                                                                                                                                                                            
INFO: END Waiting for cluster
WARNING: "stonith-timeout" is deprecated, please consider using "fencing-timeout"
WARNING: "stonith-enabled" is deprecated, please consider using "fencing-enabled"
```
- Case 2: Failed to add sbd via stage when the device does not exist on the peer node
```
# crm cluster init sbd -s /dev/sdb1 -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Configuring disk-based SBD
INFO: Initializing SBD device /dev/sdb1
INFO: Update SBD_DEVICE in /etc/sysconfig/sbd: /dev/sdb1
INFO: Update SBD_WATCHDOG_DEV in /etc/sysconfig/sbd: /dev/watchdog
INFO: Sync file /etc/sysconfig/sbd to sle16-2
INFO: Enable sbd.service on node sle16-1
INFO: Enable sbd.service on node sle16-2
INFO: Restarting cluster service
ERROR: cluster.init: Failed on sle16-2: ERROR: The cluster stack failed to start on sle16-2. Please check failed services: sbd.service
```
Should check and raise error early

liangxin1300:
Case2 does exist on crmsh-4.6, too

PR #2118, opened 2026-05-28
Fix: sbd: Verify sbd devices on all nodes

liangxin1300:
The function `utils.is_block_device()` only checks whether the device is a block device on the local node. When adding an SBD device with `crm sbd device add`, or adding SBD/OCFS2/GFS2 during the stage, errors are not detected early if the specified device is not a block device or does not exist on peer nodes.

 This commit renames the function to `utils.get_non_block_device_nodes` and refactors it to check the device on all nodes in the cluster and return a list of nodes where the device is not a block device or does not exist.
Fix issue:
- #2117

Other changes:
- Dev: bootstrap: Call join_sbd before changing corosync.conf
to avoid potential inconsistency of corosync.conf when join_sbd raises an exception
- Dev: sbd: Call SBDUtils.verify_sbd_device when doing sbd health check
- Dev: sbd: Compare SBD device UUID optionally
- Dev: sbd: Introduce function sbd.SBDUtils.get_sbd_device_metadata_raw
- Dev: ui_sbd: Call SBDUtils.verify_sbd_device in multiple places

### DIFF
--- a/crmsh/ui_sbd.py
+++ b/crmsh/ui_sbd.py
@@ -467,7 +467,7 @@ class SBD(command.UI):
         Implement sbd device add command, add devices to sbd configuration
         '''
         all_device_list = self.device_list_from_config + devices_to_add
-        sbd.SBDUtils.verify_sbd_device(all_device_list)
+        sbd.SBDUtils.verify_sbd_device(all_device_list, self.cluster_nodes)
 
         devices_to_init, _ = sbd.SBDUtils.handle_input_sbd_devices(
             devices_to_add,
--- a/crmsh/ui_sbd.py
+++ b/crmsh/ui_sbd.py
@@ -2,7 +2,6 @@ import logging
 import typing
 import re
 import os
-import shlex
 
 from crmsh import sbd
 from crmsh import watchdog
@@ -216,9 +215,7 @@ class SBD(command.UI):
         if self.device_list_from_config:
             logger.info("crm sbd configure show disk_metadata")
         for dev in self.device_list_from_config:
-            cmd = f"sbd -d {shlex.quote(dev)} dump"
-            _, out, _ = self.cluster_shell.get_rc_stdout_stderr_without_input(None, cmd)
-            print(out)
+            print(sbd.SBDUtils.get_sbd_device_metadata_raw(dev))
 
     def _show_property(self) -> None:
         '''
--- a/crmsh/ui_sbd.py
+++ b/crmsh/ui_sbd.py
@@ -383,6 +383,8 @@ class SBD(command.UI):
         '''
         Configure disk-based SBD based on input parameters and runtime config
         '''
+        sbd.SBDUtils.verify_sbd_device(self.device_list_from_config, self.cluster_nodes)
+
         update_dict = {}
         watchdog_device = parameter_dict.get("watchdog-device")
         if watchdog_device != self.watchdog_device_from_config:
@@ -406,8 +408,7 @@ class SBD(command.UI):
             result_dict = self._set_sbd_opts(crashdump_watchdog_timeout)
             update_dict = {**update_dict, **result_dict}
 
-        device_list = sbd.SBDUtils.get_sbd_device_from_config()
-        devices_consistent = sbd.SBDUtils.check_devices_metadata_consistent(device_list, quiet=True)
+        devices_consistent = sbd.SBDUtils.check_devices_metadata_consistent(self.device_list_from_config, quiet=True)
         if timeout_dict == self.device_meta_dict_runtime and not update_dict and devices_consistent:
             logger.info("No change in SBD configuration")
             return
@@ -486,6 +487,8 @@ class SBD(command.UI):
         '''
         Implement sbd device remove command, remove devices from sbd configuration
         '''
+        sbd.SBDUtils.verify_sbd_device(self.device_list_from_config, self.cluster_nodes)
+
         for dev in devices_to_remove:
             if dev not in self.device_list_from_config:
                 raise self.SyntaxError(f"Device {dev} is not in config")
@@ -612,6 +615,8 @@ class SBD(command.UI):
         self._load_attributes()
         if not self._service_is_active(constants.SBD_SERVICE):
             return False
+        if self.device_list_from_config:
+            sbd.SBDUtils.verify_sbd_device(self.device_list_from_config, self.cluster_nodes)
 
         args = _handle_force_option(args)
         purge_crashdump = False
--- a/crmsh/ui_sbd.py
+++ b/crmsh/ui_sbd.py
@@ -2,6 +2,7 @@ import logging
 import typing
 import re
 import os
+import shlex
 
 from crmsh import sbd
 from crmsh import watchdog
@@ -215,8 +216,9 @@ class SBD(command.UI):
         if self.device_list_from_config:
             logger.info("crm sbd configure show disk_metadata")
         for dev in self.device_list_from_config:
-            print(self.cluster_shell.get_stdout_or_raise_error(f"sbd -d {dev} dump"))
-            print()
+            cmd = f"sbd -d {shlex.quote(dev)} dump"
+            _, out, _ = self.cluster_shell.get_rc_stdout_stderr_without_input(None, cmd)
+            print(out)
 
     def _show_property(self) -> None:
         '''
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -2053,6 +2053,8 @@ def join_cluster(seed_host, remote_user):
     if not os.path.exists(corosync.conf()):
         utils.fatal("{} is not readable. Please ensure that hostnames are resolvable.".format(corosync.conf()))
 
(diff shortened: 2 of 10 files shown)


nicholasyang2022 (changes requested):
### Summary
Thanks for the contribution! This is a great improvement for ensuring cluster consistency during storage configuration. I have a few suggestions to improve the robustness and maintainability of the code.

### Findings

#### 1. Shell Quoting / Security
In `crmsh/utils.py`, the `is_block_device` function constructs a shell command using an f-string:
```python
rc, _, _ = shell.get_rc_stdout_stderr_without_input(node, f"test -b {dev}")
```
If a device path contains spaces (e.g., `/dev/mapper/mpath a`) or shell metacharacters, this check will fail or potentially lead to unintended shell behavior.
**Suggestion:** Use `shlex.quote(dev)` to sanitize the device path.
```python
import shlex
# ...
rc, _, _ = shell.get_rc_stdout_stderr_without_input(node, f"test -b {shlex.quote(dev)}")
```

#### 2. Logging in Utility Functions
Currently, `is_block_device` logs an error directly. This reduces the utility's reusability as it always performs a side effect (logging) regardless of the caller's needs. It also strips context from the error message.
**Suggestion:** Refactor `is_block_device` to be a "pure" check. It could return a list of nodes where the check failed. This allows the caller to log an error that is contextual to the specific operation (e.g., "SBD device validation failed" vs. "OCFS2 device validation failed") and then raise `TerminateSubCommand`.

### Conclusion
Please consider these changes to improve security and code separation.

crmsh/ui_sbd.py
```diff
@@ -215,8 +216,9 @@ def _show_disk_metadata(self) -> None:
         if self.device_list_from_config:
             logger.info("crm sbd configure show disk_metadata")
         for dev in self.device_list_from_config:
-            print(self.cluster_shell.get_stdout_or_raise_error(f"sbd -d {dev} dump"))
-            print()
+            cmd = f"sbd -d {shlex.quote(dev)} dump"
+            _, out, _ = self.cluster_shell.get_rc_stdout_stderr_without_input(None, cmd)
+            print(out)
```
nicholasyang2022: Stderr is silently dropped here.

nicholasyang2022 (changes requested):
# Code Review: Pull Request #2118

## Summary
This pull request addresses a critical gap in SBD and filesystem stage validation (Issue #2117). Refactoring the block device verification (`is_block_device`) to be cluster-wide via `utils.get_non_block_device_nodes` is an excellent security and robustness improvement.

However, during a deep dive of the implementation, two **critical crash-level regressions** were identified under uninitialized/unconfigured cluster states, alongside an **architectural code smell** regarding how behavior is toggled in `verify_sbd_device`.

---

## Findings

### 🔴 Critical (Crash Risks & Regressions)

#### 1. Unhandled `ValueError` in `get_sbd_device_metadata`
*   **Location:** `crmsh/sbd.py` — `get_sbd_device_metadata()`
*   **The Problem:** 
    *   Previously, if the `sbd dump` command failed or returned empty (common when SBD is not yet configured or initialized), the function safely returned an empty dictionary `{}`.
    *   With this PR, `get_sbd_device_metadata` delegates to `get_sbd_device_metadata_raw`, which explicitly raises a `ValueError` on command failure or empty stdout. Additionally, strict key assertions on `"msgwait"` and `"watchdog"` will always throw if they are missing from `sbd_info`.
    *   **Why this is critical:** Highly sensitive call sites (e.g., `_load_attributes` in `crmsh/ui_sbd.py` during shell initialization, and `_load_configurations_from_runtime` during status/health checks) **do not catch** this `ValueError`. Thus, entering the `crm sbd` shell or running `crm cluster health` on an unconfigured or uninitialized cluster will now crash with an unhandled exception stack trace.
*   **Suggested Remediation:** Restore the previous "best-effort dict" contract of `get_sbd_device_metadata`. It should catch exceptions from raw metadata extraction and safely return an empty or partial dictionary `{}` without throwing:
    ```python
    @staticmethod
    def get_sbd_device_metadata(dev, timeout_only=False, remote=None) -> dict:
        '''
        Extract metadata from sbd device header
        '''
        sbd_info = {}
        try:
            out = SBDUtils.get_sbd_device_metadata_raw(dev, remote)
        except ValueError as e:
            logger.debug("Failed to get raw SBD metadata: %s", e)
            return sbd_info

        pattern = r"UUID\s+:\s+(\S+)|Timeout\s+\((\w+)\)\s+:\s+(\d+)"
        matches = re.findall(pattern, out)
        for uuid, timeout_type, timeout_value in matches:
            if uuid and not timeout_only:
                sbd_info["uuid"] = uuid
            elif timeout_type and timeout_value:
                sbd_info[timeout_type] = int(timeout_value)

        return sbd_info
    ```

### folded low confidence problem
<details>
<summary>Unhandled `TypeError` in `_check_sbd_disk_metadata`</summary>

#### 2. Unhandled `TypeError` in `_check_sbd_disk_metadata`
*   **Location:** `crmsh/sbd.py` — `_check_sbd_disk_metadata()`
*   **The Problem:** 
    *   If `get_sbd_device_metadata` is restored to be non-throwing (or returned empty results previously), `self.sbd_watchdog_timeout` or `self.sbd_msgwait` could be set to `None` in `_load_configurations_from_runtime()`.
    *   The check `if self.sbd_watchdog_timeout < self.sbd_watchdog_timeout_expected:` will then raise a `TypeError` in Python 3 (comparing `None < int`), crashing the health check command.
*   **Suggested Remediation:** Add a defensive guard to log a clean error and return `CheckResult.ERROR` rather than allowing a `TypeError` to propagate:
    ```python
            if self.sbd_watchdog_timeout is None or self.sbd_msgwait is None:
                self._log_when_not_quiet(
                    logging.ERROR,
                    "Cannot check SBD disk metadata: msgwait or watchdog timeout is missing"
                )
                return CheckResult.ERROR
    ```
</details>

---

### 🟡 Improvements & Refactoring (Code Smell)

#### 3. Use of Boolean Flag Parameter in `verify_sbd_device`
*   **Location:** `crmsh/sbd.py` — `verify_sbd_device(dev_list, node_list=None, compare_uuid=False)`
*   **The Problem:** Controlling function behavior with a boolean flag (the **"Flag Argument"** pattern) violates the *Single Responsibility Principle* and creates a *"Boolean Trap"* at the call site, making it harder to read and maintain (e.g., `SBDUtils.verify_sbd_device(dev_list, node_list, True)`). It has more than one responsibility.
    * Without `compare_uuid`: The function validates device structure and presence (i.e., "Is this a block device on all nodes?").
    * With `compare_uuid`: It also validates the actual initialized payload data (i.e., "Does this already-initialized device have matching metadata across the cluster?").
*   **Suggested Remediation (Distinct Functions):** Decouple the concerns. Have one function verify block device topology (presence and device type), and another verify payload/metadata consistency (UUIDs). Callers can then compose these cleanly:
    ```python
    @staticmethod
    def verify_sbd_block_devices(dev_list, node_list=None):
        """Verify that devices are valid block devices across the specified nodes."""
        if len(dev_list) > SBDManager.SBD_DEVICE_MAX:
            raise ValueError(f"Maximum number of SBD device is {SBDManager.SBD_DEVICE_MAX}")
        for dev in dev_list:
            failed_nodes = utils.get_non_block_device_nodes(dev, node_list)
            if failed_nodes:
                raise ValueError(f"{dev} is not a block device on {', '.join(failed_nodes)}")
        utils.detect_duplicate_device_path(dev_list)

    @staticmethod
    def verify_sbd_devices_uuid_consistency(dev_list, node_list=None):
        """Verify that all devices share consistent UUIDs across the specified nodes."""
        for dev in dev_list:
            SBDUtils.compare_device_uuid(dev, node_list)
    ```

---

### 🟢 Strengths
*   **Robust Multi-Node Verification:** Refactoring the block device checks from local-only (`is_block_device`) to cluster-wide (`get_non_block_device_nodes` via `test -b`) is highly robust, secure against injection (using `shlex.quote`), and solves the core problem elegantly.
*   **Sequence Correction in Bootstrap:** Calling `join_sbd` before updating `corosync.conf` is a fantastic sequence improvement. If the join raises an exception (e.g., during device verification), it prevents leaving `corosync.conf` in an inconsistent state.
*   **High Test Coverage:** The unit tests and feature files are meticulously updated to mock and assert the new behavior.

---

## Conclusion

### ⚠️ Recommendation: **Request Changes**

The architectural shift to multi-node verification is excellent and highly recommended for merge. However, the PR should **not be merged in its current state** due to the crash regressions. Once the suggested changes to restore the non-throwing, best-effort dict contract are applied and the flag argument is resolved, this PR is an outstanding contribution.

liangxin1300:
>1. Unhandled ValueError in get_sbd_device_metadata

This is the behavior before #2120.
Now I prefer to revert the behavior change introduced in #2120.
Since there are many places calling `get_sbd_device_metadata` that don't detect whether the metadata dict is empty,
I think it's better to raise the exception directly to terminate the command at once because I think the failure of `sbd -d <dev> dump` is critical

nicholasyang2022 (approved):

PR #2120, opened 2026-05-29
Fix: sbd: Compare SBD device UUID when adding SBD via sbd stage

liangxin1300:
The previous code only compared SBD device UUID when joining the cluster, that's not enough.  When adding SBD via sbd stage, after initializing the SBD device, also need to compare SBD device UUID between cluster nodes

Fix issue:
- #2121


nicholasyang2022 (approved):
LGTM

Issue #2076, opened 2026-04-10
Not to mix up crm report debug log with crmsh debug log


liangxin1300:
```
# crm report -v
DEBUG: su_subprocess_run: ['/bin/sh', '-c', 'ssh -A -o StrictHostKeyChecking=no -o BatchMode=yes root@sle16-2 /bin/bash'], env={'SSH_AUTH_SOCK': ''}, {'input': b"grep 'INFO: Trace for .* is written to ' /var/log/crmsh/crmsh.log*|grep -v 'collect'", 'start_new_session': True, 'stdout': -1, 'stderr': -1}
DEBUG: su_subprocess_run: ['/bin/sh', '-c', 'ssh -A -o StrictHostKeyChecking=no -o BatchMode=yes root@sle16-2 /bin/bash'], env={'SSH_AUTH_SOCK': ''}, {'input': b'/usr/sbin/crm report __collector \'{"name": "crm_report", "from_time": 1775745149.268397, "to_time": 1775788349.0, "no_compress": false, "speed_up": false, "extra_log_list": ["/var/log/messages", "/var/log/crmsh/crmsh.log", "/etc/crm/profiles.yml", "/etc/crm/crm.conf", "~/.config/crm/crm.conf"], "rm_exist_dest": false, "single": false, "sensitive_regex_list": [], "regex_list": ["CRIT:", "ERROR:", "error:", "warning:", "crit:"], "me": "sle16-1", "main_node": "sle16-1", "help": false, "node_list": ["sle16-1", "sle16-2"], "ssh_user": null, "ssh_option_list": [], "no_log_list": [], "sanitize": false, "debug": 1, "dest": "crm_report-Fri-10-Apr-2026", "dest_dir": ".", "compress_prog": "gzip", "compress_suffix": ".gz", "dest_path": "./crm_report-Fri-10-Apr-2026.tar.gz", "cib_dir": "/var/lib/pacemaker/cib", "pcmk_exec_dir": "/usr/libexec/pacemaker", "pe_dir": "/var/lib/pacemaker/pengine", "pcmk_lib_dir": "/var/lib/pacemaker", "cores_dir_list": ["/var/lib/pacemaker/cores", "/var/lib/corosync"], "work_dir": "/tmp/crmsh_okzb5zpk/crm_report-Fri-10-Apr-2026", "trace_dir_list": []}\'', 'start_new_session': True, 'stdout': -1, 'stderr': -1}
[...]

PR #2123, opened 2026-06-01
Dev: log: Lower corosync config tokenizer log level to reduce noise and decouple logging from config

nicholasyang2022:
This PR reduces log noise when standard debug logging is enabled and refactors the logging architecture to be more standard and decoupled from global configuration.

#### 1. Corosync Config Tokenizer Log Level
- Lowered the level of debug logs from the tokenizer in `corosync_config_format` to `DEBUG2` (defined as `logging.DEBUG - 1`). 
- These logs are now hidden by default even when standard debug logging is active, significantly reducing noise unless specifically debugging the parser.

#### 2. Logging Architecture & Decoupling
- **Decoupled from Config**: Introduced a listener mechanism in `crmsh.config` to notify observers of configuration changes. `crmsh.log` now operates independently, with `crmsh.main` syncing logger levels via the listener.
- **Filter-based Output Control**: Refactored `LoggerUtils` to use standard Python `logging` filters (e.g., `BlockingFilter`) instead of dynamically adding/removing handlers to control output.
- **Standardization**: Replaced the custom `crmsh.log.setup_logger` with standard `logging.getLogger(__name__)` calls across the entire codebase.

#### 3. Crash Test Logging Refactoring
- Refactored `crash_test` to use the new log filtering mechanism to toggle output instead of temporarily removing handlers, leading to more robust state management during tests.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -13,6 +13,7 @@ import codecs
 import dataclasses
 import io
 import json
+import logging
 import os
 import subprocess
 import sys
@@ -56,7 +57,7 @@ from . import watchdog
 import crmsh.healthcheck
 
 
-logger = log.setup_logger(__name__)
+logger = logging.getLogger(__name__)
 logger_utils = log.LoggerUtils(logger)
 
 
--- a/crmsh/cibconfig.py
+++ b/crmsh/cibconfig.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 import copy
+import logging
 import os
 import sys
 import re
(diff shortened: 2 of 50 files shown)


liangxin1300:
Nice work!
Then
- #2076
can be closed

liangxin1300:
Q: How to trigger debug2? Only when using `crm report -vv`?

nicholasyang2022:
> Q: How to trigger debug2? Only when using `crm report -vv`?

Yes, do we need also to support `crm -dd`?

liangxin1300:
> > Q: How to trigger debug2? Only when using `crm report -vv`?
> 
> Yes, do we need also to support `crm -dd`?

Can be implemented in the future if needed

liangxin1300 (approved):

PR #2129, opened 2026-06-10
Dev: build: Migrate build system to Meson

### LINKED ISSUE #139: Get rid of autotools
I'd like to switch to using only python-setuptools to configure and install, ideally so that crmsh can be submitted to PyPI. The biggest problem is building the man page using asciidoc.

nicholasyang2022:
This pull request migrates the build system from Autotools and setuptools to Meson, simplifying configuration variables and runtime build config fetching.

### Changes:
- **Build System Migration**: Replaced Autotools and setuptools with Meson.
- **Configuration Simplification**: Combined version variables, removed redundant options, and standardized paths.
- **Runtime Build Config**: Replaced version.in with a templated crmsh/build_config.py.in.
- **Configurability**: Made docdir configurable to support distribution-specific layouts.
- **CI Container Optimization**: Integrated local building, caching, and sharing of the test container across functional tests.

Closes #139


meson.build
```diff
@@ -0,0 +1,137 @@
...
+profiles_html = custom_target('profiles_html',
+  input: ['doc/profiles.adoc'],
+  output: ['profiles.html'],
+  command: [asciidoc, '--unsafe', '--backend=xhtml11', '-o', '@OUTPUT@', '@INPUT@'],
+  install: true,
+  install_dir: docdir
+)
+crmsh_crm_report_html = custom_target('crmsh_crm_report_html',
+  input: ['doc/crmsh_crm_report.8.adoc'],
+  output: ['crmsh_crm_report.8.html'],
```
liangxin1300, line 133: Considering there is no real content in crmsh_crm_report.8.adoc, how about we remove this file from now?

If we still keep it, the output should not the html file

liangxin1300 (approved):
LGTM, except the above comment

Thanks!

meson.build
nicholasyang2022: It is built into an html in the master branch. This pull request is a refactoring of the build system and I tried to produce the same build result. We can clean this up in another pull request.

```
~ ❯ bsdtar -tf ~/Downloads/crmsh-5.1.0+20260814.d9688661-1.1.noarch.rpm | rg html
./usr/share/doc/packages/crmsh/crmsh_crm_report.8.html
./usr/share/doc/packages/crmsh/profiles.html
```

PR #2136, opened 2026-06-19
Dev: ui_sbd: Check sbd health before running 'crm sbd' subcommands

liangxin1300:
Call SBD.check_sbd_health() before running:
- crm sbd configure
- crm sbd device
- crm sbd purge

Return False if SBD health check fails

### DIFF
--- a/crmsh/ui_sbd.py
+++ b/crmsh/ui_sbd.py
@@ -526,6 +526,9 @@ class SBD(command.UI):
                 raise self.SyntaxError(f"Invalid argument: {args[0]}")
             if len(args) < 2:
                 raise self.SyntaxError("No device specified")
+            if not SBD.check_sbd_health():
+                logger.error("This command is aborted due to the above SBD issues")
+                return False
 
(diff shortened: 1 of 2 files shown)


nicholasyang2022:
IMO users should be able to change config or purge the SBD when it is unhealthy.

liangxin1300:
> IMO users should be able to change config or purge the SBD when it is unhealthy.

They can use `crm cluster health sbd --fix` to fix at first
@zzhou1 What do you think?

nicholasyang2022:
> > IMO users should be able to change config or purge the SBD when it is unhealthy.
> 
> They can use `crm cluster health sbd --fix` to fix at first

After considering the scenarios carefully, I still have strong concerns about blocking these subcommands, because it would block administrators from disaster recovery. In case of hardware(disk/network) failure, `crm cluster health sbd --fix` will not help, and blocking the subcommands make it impossible to move to another disk.

And it is also counter-productive when users' goal is to purge SBD. When they just want to get rid of SBD, they should not be force to troubleshoot first.

PR #2148, opened 2026-07-03
Dev: utils: Drop utils.pacemaker_daemon function

liangxin1300:
No need to support legacy pacemaker(1.x) daemon name


nicholasyang2022 (changes requested):
`crmsh.utils.pacemaker_fenced` is called in `crmsh.ra.get_fenced_meta`.

nicholasyang2022 (approved):

PR #2150, opened 2026-07-06
Fix: ui_node: Use boolean values for node state attrs

liangxin1300:
Write standby and maintenance node attributes as true/false while still accepting existing on/off values.

Also add guidance in command output and help text for restoring nodes from standby or maintenance mode.

### DIFF
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -365,8 +365,8 @@ class NodeMgmt(command.UI):
                     xmlutil.rmnodes(item_to_del)
                 # If the standby nvpair already exists, set and continue
                 item = cib.xpath(xml_query_path.format(node_id=node_id))
-                if item and item[0].get("value") != "on":
-                    item[0].set("value", "on")
+                if item and item[0].get("value") not in ("true", "on"):
+                    item[0].set("value", "true")
                     continue
                 # Create standby nvpair
                 interface_item = xml_item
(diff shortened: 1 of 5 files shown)


crmsh/ui_node.py
```diff
@@ -433,7 +438,10 @@ def do_ready(self, context, node=None):
             node = utils.this_node()
         if not utils.is_name_sane(node):
             return False
-        return utils.ext_cmd(self.node_maint % (node, "off")) == 0
```
nicholasyang2022, line 436: 1. Changing the implementation of setting node attributes is worth mentioning in commit message.
2.  `self.node_maint` (and also `self.node_standby`) is dead code and should be removed.

crmsh/ui_node.py
```diff
@@ -422,7 +423,11 @@ def do_maintenance(self, context, node=None):
             node = utils.this_node()
         if not utils.is_name_sane(node):
             return False
-        return self._commit_node_attr(context, node, "maintenance", "true")
+        rc = self._commit_node_attr(context, node, "maintenance", "true")
+        if rc:
+            logger.info("Node %s is now in maintenance mode", node)
+            logger.info("To restore the node from maintenance mode, use the `ready` command")
```
nicholasyang2022, line 429: As a user-facing application, changes in output messages is a significant and should be mention in commit message title and pull request title.

nicholasyang2022 (commented):
Multiple changes are combined in one commit and pull request. All the changes need to be documented in messages (probably with separated commits), and the major, user-visible change should be described in the title.

liangxin1300:
Combine this PR with #2163

Issue #2151, opened 2026-07-06
Failed to configure sbd with softdog


liangxin1300:
```
 # crm cluster init sbd -w softdog -s /dev/mapper/shared_iscsi_lun1-part5 -y
INFO: Loading "default" profile from /etc/crm/profiles.yml
INFO: Loading "knet-default" profile from /etc/crm/profiles.yml
INFO: Configuring disk-based SBD
INFO: Initializing SBD device /dev/mapper/shared_iscsi_lun1-part5
INFO: Update SBD_DEVICE in /etc/sysconfig/sbd: /dev/mapper/shared_iscsi_lun1-part5
INFO: Update SBD_WATCHDOG_DEV in /etc/sysconfig/sbd: /dev/watchdog1
INFO: Sync file /etc/sysconfig/sbd to sle16-2
INFO: Enable sbd.service on node sle16-1
INFO: Enable sbd.service on node sle16-2
INFO: Restarting cluster service
ERROR: cluster.init: Failed on sle16-2: ERROR: The cluster stack failed to start on sle16-2. Please check failed services: sbd.service
```

liangxin1300:
And crmsh should give a warning when user wants to configure softdog as watchdog driver

PR #2155, opened 2026-07-12
Fix problems of softdog watchdog driver

liangxin1300:
Problems:
1. Peer node doesn't load the softdog watchdog driver when running the sbd stage
    see:
    - #2151
2. `crm cluster health sbd` should give a warning when using softdog as watchdog driver
3. `crm cluster init -w softdog` fails with `ERROR: cluster.init: Failed to get node list from cluster` when softdog isn't loaded yet, because the driver-loading code tried to discover the cluster's node list from the CIB before the cluster exists

Changes:
1. fix(watchdog): load driver on the correct nodes, and on the right cluster state
    - `cluster_run_cmd()` now takes an explicit node list instead of relying on a mutable default argument
    - Split driver loading into two low-level primitives, `_configure_and_load_driver()` (write `WATCHDOG_CFG` and reload) and `_reload_driver()` (reload only, when the config was already written elsewhere), and let the callers (`init_watchdog()`/`join_watchdog()`) decide which one to use and on which nodes
    - Add a `cluster_is_running` flag to `Watchdog`, threaded through from `SBDManager`'s bootstrap context, to tell apart the "cluster already running" and "cluster not running yet" cases
    - Load the watchdog driver on all cluster nodes when fixing up the local sbd stage with the cluster already running (node list discovered from the running CIB)
    - Load the watchdog driver on the local node only when initializing a cluster that isn't running yet (e.g. `crm cluster init`), since there is no cluster to query nor other nodes to reach at that point
    - Reload the watchdog module on the local node only when joining a cluster: the watchdog config file has already been synced from the cluster, so there is no need to query a remote peer for its driver name
    - Drop `Watchdog._get_driver_through_device_remotely()`: joining nodes no longer need to query a peer for the watchdog driver name
    - Add a behave scenario covering init with softdog followed by a successful join
2. refactor(watchdog): prefer hardware watchdog over softdog by default
    - Prefer a non-softdog device from the query-watchdog result
    - Fall back to softdog only when no other driver is found
    - Drop the unused `_get_first_unused_device()` helper and `ignore_error` flag
    - Update watchdog unit tests to cover the new selection order
3. refactor(watchdog): share query-watchdog parsing and warn on softdog use
    - Add `Watchdog.get_watchdog_info()` as the single query-watchdog parser
    - Reuse it in `ui_sbd` and when looking up the local device
    - Add `Watchdog.warn_if_using_softdog()` and call it from `SBDConfigChecker`
    - Cover the new parser and warning helper with unit tests
    - Add a behave assertion that `crm cluster health sbd` warns about softdog in stderr after an init+join with softdog

### DIFF
--- a/crmsh/watchdog.py
+++ b/crmsh/watchdog.py
@@ -35,16 +35,13 @@ class Watchdog(object):
         return self._watchdog_device_name
 
     @staticmethod
-    def verify_watchdog_device(dev, ignore_error=False):
+    def verify_watchdog_device(dev):
         """
         Use wdctl to verify watchdog device
         """
-        rc, _, err = ShellUtils().get_stdout_stderr("wdctl {}".format(dev))
+        rc, _, err = ShellUtils().get_stdout_stderr(f"wdctl {dev}")
         if rc != 0:
-            if ignore_error:
-                return False
-            else:
-                utils.fatal("Invalid watchdog device {}: {}".format(dev, err))
+            utils.fatal(f"Invalid watchdog device {dev}: {err}")
         return True
 
     @staticmethod
@@ -149,29 +146,16 @@ class Watchdog(object):
         else:
             utils.fatal("Failed to run {} remotely: {}".format(self.QUERY_CMD, err))
 
-    def _get_first_unused_device(self):
-        """
-        Get first unused watchdog device name
-        """
-        for dev in self._watchdog_info_dict:
-            if self.verify_watchdog_device(dev, ignore_error=True):
-                return dev
-        return None
-
     def _set_input(self):
-        """
-        If self._input was not provided by option:
-          1. Try to get it from sbd config file
-          2. Try to get the first valid device from result of sbd query-watchdog
-          3. Set the self._input as softdog
-        """
-        if not self._input:
-            dev = self.get_watchdog_device_from_sbd_config()
-            if dev and self.verify_watchdog_device(dev, ignore_error=True):
+        if self._input:
+            return
+
+        for dev, driver in self._watchdog_info_dict.items():
+            if driver != "softdog":
                 self._input = dev
                 return
-            first_unused = self._get_first_unused_device()
-            self._input = first_unused if first_unused else "softdog"
+
+        self._input = "softdog"
 
     def _valid_device(self, dev):
         """
--- a/crmsh/watchdog.py
+++ b/crmsh/watchdog.py
@@ -1,8 +1,13 @@
+import logging
 import re
+
 from . import utils
 from .sh import ShellUtils
 from . import sh
 from . import sbd
+
+
+logger = logging.getLogger(__name__)
 
 
 class Watchdog(object):
@@ -96,6 +101,19 @@ class Watchdog(object):
                     driver = configured_driver
             watchdog_info[device] = driver
         return watchdog_info
+
+    @staticmethod
+    def warn_if_using_softdog():
+        """
+        Warn if SBD is using softdog as watchdog driver.
+        """
+        rc, out, err = ShellUtils().get_stdout_stderr(Watchdog.QUERY_CMD)
+        if rc != 0 or not out:
+            logger.debug("Failed to run %s: %s", Watchdog.QUERY_CMD, err)
+            return
+
+        if "softdog" in Watchdog.get_watchdog_info(out, sbd_only=True).values():
+            logger.warning("It's not recommended to use softdog as watchdog driver in production environment")
 
     def _set_watchdog_info(self):
         """
--- a/crmsh/watchdog.py
+++ b/crmsh/watchdog.py
@@ -41,12 +41,13 @@ class Watchdog(object):
         return True
 
     @staticmethod
-    def _load_watchdog_driver(driver):
+    def _load_watchdog_driver(driver, join=False):
         """
         Load specific watchdog driver
         """
-        ShellUtils().get_stdout_stderr(f"echo {driver} > {Watchdog.WATCHDOG_CFG}")
-        ShellUtils().get_stdout_stderr("systemctl restart systemd-modules-load")
+        cmd = f"echo {driver} > {Watchdog.WATCHDOG_CFG} && systemctl restart systemd-modules-load"
+        node_list = utils.this_node() if join else None
+        utils.cluster_run_cmd(cmd, node_list)
 
     @staticmethod
     def get_watchdog_device_from_sbd_config():
@@ -136,7 +137,7 @@ class Watchdog(object):
 
     def join_watchdog(self):
         """
-        In join proces, get watchdog device from config
+        In join process, get watchdog device from config
         If that device not exist, get driver name from init node, and load that driver
         """
         self._set_watchdog_info()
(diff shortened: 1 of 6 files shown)


crmsh/watchdog.py
```diff
@@ -28,25 +35,23 @@ def watchdog_device_name(self):
...
-            if ignore_error:
-                return False
-            else:
-                utils.fatal("Invalid watchdog device {}: {}".format(dev, err))
+            utils.fatal(f"Invalid watchdog device {dev}: {err}")
         return True
 
     @staticmethod
-    def _load_watchdog_driver(driver):
+    def _load_watchdog_driver(driver, join=False):
```
nicholasyang2022, line 48: Passing `join` here couples a low level driver loading util with the cluster-level concept (init vs join). A better approach is to pass `node_list` here and let the caller decide.

crmsh/watchdog.py
```diff
@@ -64,16 +69,57 @@ def _driver_is_loaded(driver):
...
+        """
+        if not out:
+            return {}
+
+        watchdog_info = {}
+        for device, identity, driver in re.findall(cls.DEVICE_FIND_REGREX, out):
+            if sbd_only and not re.search(r"Busy: .*sbd", identity):
+                continue
+            if driver == "<unknown>":
+                configured_driver = cls._get_configured_watchdog_driver()
```
nicholasyang2022, line 96: This is incorrect as `get_watchdog_info` accepts arbitrary `out`, possibly from a remote node, but `_get_configured_watchdog_driver()` query the local node for driver information.

Probably we should fix `sbd` to make it to print the correct driver instead of add workarounds in crmsh.

liangxin1300: Try to fix that in https://github.com/ClusterLabs/sbd/pull/162

liangxin1300: The above sbd PR#162 has been merged

crmsh/watchdog.py
```diff
@@ -148,7 +163,7 @@ def join_watchdog(self):
 
         if not self._valid_device(self._input):
             driver = self._get_driver_through_device_remotely(self._input)
-            self._load_watchdog_driver(driver)
+            self._load_watchdog_driver(driver, node_list=utils.this_node())
```
nicholasyang2022, line 166:
```suggestion
            self._load_watchdog_driver(driver, node_list=[utils.this_node()])
```

liangxin1300: Good catch!
Should add a functional test case

crmsh/watchdog.py
liangxin1300: `crm cluster health sbd` wouldn't give a warning until the above sbd PR goes into production

crmsh/watchdog.py
```diff
@@ -28,25 +33,27 @@ def watchdog_device_name(self):
...
-            if ignore_error:
-                return False
-            else:
-                utils.fatal("Invalid watchdog device {}: {}".format(dev, err))
+            utils.fatal(f"Invalid watchdog device {dev}: {err}")
         return True
 
     @staticmethod
-    def _load_watchdog_driver(driver):
+    def _load_watchdog_driver(driver=None, cluster_is_running=True, is_join=False):
```
nicholasyang2022: Passing `cluster_is_running`, `is_join` here couples a low level driver loading util with the cluster-level orchestration. A better approach is to provide different functions to manage `WATCHDOG_CFG` and to load modules, and let the caller decide which function/on which node to call.

nicholasyang2022 (changes requested):
### Summary

This PR addresses several key issues with watchdog configuration, specifically around `softdog` handling:
1. Ensuring peer nodes load the driver when configuring SBD on an existing cluster (#2151).
2. Avoiding premature CIB queries when initializing a cluster that is not running yet.
3. Preferring hardware watchdog devices over `softdog` by default.
4. Consolidating `sbd query-watchdog` parsing into `Watchdog.get_watchdog_info()` and warning when `softdog` is used in production.

All automated GitHub Actions CI checks have passed. However, the changes introduce a critical issue regarding persistent driver configuration as well as opportunities for architectural decoupling and earlier error detection.

---

### Findings

#### 1. Critical

* **Writing `/etc/modules-load.d/watchdog.conf` is skipped if the module is already loaded in the kernel:**
  In `crmsh/watchdog.py`:
  ```python
  # self._input is a driver name, load it if it was unloaded
  node_list = None if self._cluster_is_running else [utils.this_node()]
  if not self._driver_is_loaded(self._input, node_list=node_list):
      self._configure_and_load_driver(self._input, node_list=node_list)
      self._set_watchdog_info()
  ```
  `_configure_and_load_driver` bundles two distinct operations:
  1. Writing `echo {driver} > /etc/modules-load.d/watchdog.conf` (persistence across reboots & csync2 sync target).
  2. Reloading systemd modules (`systemctl restart systemd-modules-load`).

  If a driver (such as `softdog`) is **already loaded** in memory (e.g. loaded beforehand via `modprobe softdog`, booted with it, or chosen by `_set_input()` when only softdog exists):
  * `_driver_is_loaded` evaluates to `True`.
  * `_configure_and_load_driver` is **skipped entirely**.
  * `/etc/modules-load.d/watchdog.conf` is **never created or updated**.

  **Consequences:**
  * **On host reboot:** Systemd does not load `softdog`, and `sbd.service` fails to start on boot.
  * **On join:** The joining node retrieves an empty or non-existent `watchdog.conf` via `retrieve_files()`, so `join_watchdog()`'s `_reload_driver()` does not load `softdog` on the joining node, causing SBD to fail on the peer.

  **Recommendation:**
  Writing the persistent configuration file (`WATCHDOG_CFG`) should **always** take place when configuring a driver, regardless of whether the module happens to be currently present in `lsmod`.

---

#### 2. Improvements

* **Decouple `Watchdog` from cluster orchestration by passing `node_list`:**
  `Watchdog` still retains `cluster_is_running` in `__init__` and `get_watchdog_device(dev_or_driver=None, cluster_is_running=True)`.
  As noted in previous review comments, `Watchdog` is a low-level utility for managing watchdog devices and kernel module files. It should not be coupled to the high-level Pacemaker/CIB cluster state.
  * Callers like `SBDManager` already determine target nodes (e.g. `cluster_nodes = utils.list_cluster_nodes() if self.cluster_is_running else [utils.this_node()]`).
  * Passing `node_list` directly to `Watchdog(node_list=...)` or `get_watchdog_device(dev_or_driver=None, node_list=None)` simplifies the interface and avoids embedding cluster-state discovery inside `Watchdog`.

* **Missing post-reload verification in `join_watchdog()`:**
  In `crmsh/watchdog.py`:
  ```python
  if not self._valid_device(self._input):
      self._reload_driver([utils.this_node()])
  ```
  If `_reload_driver` fails (e.g., module failed to load or failed to create the device node), `join_watchdog()` silently finishes. SBD then fails later when starting services.
  Re-running `self._set_watchdog_info()` and asserting `self._valid_device(self._input)` after `_reload_driver()` would surface failures immediately with a descriptive fatal message.

---

### Conclusion & Recommendation

**Recommendation: Request Changes**

The extraction of `get_watchdog_info()` and preferring hardware watchdogs by default are great improvements. However, skipping `WATCHDOG_CFG` creation when the driver is in memory leads to persistent boot and join failures. This point and the decoupling of cluster state should be addressed before merging.

PR #2158, opened 2026-07-14
Dev: log: Add quiet-aware logger adapter

liangxin1300:
- Add QuietLoggerAdapter with early quiet filtering
- Let quiet-aware modules use normal logger methods
- Avoid SBD-specific logging helper duplication
- Update SBD checks to use the quiet-aware adapter
- Add unit tests for quiet logging behavior

### DIFF
--- a/crmsh/log.py
+++ b/crmsh/log.py
@@ -26,6 +26,20 @@ class DEBUG2Logger(logging.Logger):
             self._log(DEBUG2, msg, args, **kwargs)
 
 
+class QuietLoggerAdapter(logging.LoggerAdapter):
+    def __init__(self, logger, quiet):
+        super().__init__(logger, {})
+        self.quiet = quiet
+
+    def log(self, level, msg, *args, **kwargs):
+        quiet = self.quiet
+        if callable(quiet):
+            quiet = quiet()
+        if quiet:
+            return
+        self.logger.log(level, msg, *args, **kwargs)
+
+
 class NumberedLoggerInterface(DEBUG2Logger):
     """
     Interface to prepend a number to the message, used for regression test. When this class is used directly, no numbers are prepend.
--- a/crmsh/sbd.py
+++ b/crmsh/sbd.py
@@ -255,11 +255,9 @@ class SBDTimeout(object):
         self.quiet = False
         if self.context:
             self.quiet = self.context.quiet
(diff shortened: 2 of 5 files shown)


nicholasyang2022 (changes requested):
We need to simplify this architecture before merging. The current implementation introduces unnecessary complexity by adding a separate `QuietFilter` and modifying `LOGGING_CFG` across multiple handlers.

### Suggested Refactoring: Pass `quiet` directly and filter early in the Adapter

Instead of passing a `quiet` attribute to `extra` and registering a `QuietFilter` on all handlers inside `LOGGING_CFG`, we should customize the constructor of `QuietLoggerAdapter` to accept a direct `quiet` argument and drop logs early within `QuietLoggerAdapter.log(...)`.

#### Proposed implementation in `crmsh/log.py`:
```python
class QuietLoggerAdapter(logging.LoggerAdapter):
    def __init__(self, logger, quiet):
        super().__init__(logger, {})
        self.quiet = quiet

    def log(self, level, msg, *args, **kwargs):
        quiet = self.quiet
        if callable(quiet):
            quiet = quiet()
        if quiet:
            return  # Drop the log early at the adapter level
        super().log(level, msg, *args, **kwargs)
```

#### Instantiation in `crmsh/sbd.py`:
```python
self.logger = log.QuietLoggerAdapter(logger, quiet=lambda: self.quiet)
```

### Why this must be changed:
1. **Eliminate Configuration Pollution:** This removes the need for `QuietFilter` completely and keeps `LOGGING_CFG` untouched. There is no reason to modify the configuration of multiple individual handlers (`console`, `console_report`, `buffered_console`, `file`).
2. **Prevent Log Leakage:** Filtering at the handler level is fragile. If a new handler is added in the future and `QuietFilter` is missed, quiet logs will leak to it. Filtering early at the adapter level guarantees no leak.
3. **No Dictionary Mutation:** We bypass `extra` and `process(...)` entirely, meaning we don't mutate dictionary parameters or pollute the namespaces of shared log records.
4. **Prefer Standard Log Levels for Granular Control:** A custom boolean `quiet` flag is fundamentally all-or-nothing. If fine-grained diagnostic control is ever needed in the future (e.g. enabling warnings while hiding info logs), we should leverage log levels rather than custom handler-level boolean filters.

nicholasyang2022 (approved):
~deleted~

PR #2159, opened 2026-07-16
Dev: sbd: Check and fix deprecated properties

liangxin1300:
## Description

This PR updates crmsh to use Pacemaker replacement cluster-property names by default and adds deprecated-property handling to the SBD health check/fix flow.

It reverts bootstrap-side use of deprecated `stonith-*` names, discovers configured deprecated cluster properties from Pacemaker metadata, warns when deprecated properties are still present, and migrates or removes them during `crm cluster health sbd --fix`.

## Changes

- Use `fencing-*` properties in bootstrap, SBD, qdevice, crash-test, UI, feature tests, and unit tests instead of deprecated `stonith-*` names.
- Hide deprecated RA parameters from configure completion and property listings while keeping metadata available for translation.
- Add an SBD deprecated-property check/fix item to `SBDConfigChecker`.
- Discover configured deprecated properties dynamically from Pacemaker metadata.
- Split RA helpers into predictable active, deprecated, and deprecated-mapping helpers.
- Add raw property helpers for migration I/O without recursive translation.
- Add translator checks that return success/warning state and use quiet-aware warning output.
- For deprecated terms without replacements, delete configured defaults and keep values that are non-default or have no metadata default.
- Add unit coverage for deprecated-property discovery, SBD check/fix behavior, raw migration paths, no-replacement handling, default-value cleanup, and missing-value handling.

## Fix policy for deprecated properties

`utils.DeprecatedTermTranslator.fix()` applies this policy when the deprecated term is configured and the running cluster uses deprecated properties:

- If the replacement term is also configured, delete the deprecated term because Pacemaker ignores it.
- If a replacement term exists, copy the deprecated value to the replacement term with raw property I/O, then delete the deprecated term.
- If no replacement term exists and the configured value equals the Pacemaker metadata default, delete the deprecated term.
- If no replacement term exists and the property has no metadata default, keep the deprecated term.
- If no replacement term exists and the configured value differs from the metadata default, keep the deprecated term.

## Notes

- `get_property()` and `set_property()` remain the translated public wrappers.
- `_get_raw_property()` and `_set_raw_property()` are used where the exact property name must be addressed during migration.

### DIFF
--- a/crmsh/bootstrap.py
+++ b/crmsh/bootstrap.py
@@ -1417,7 +1417,7 @@ def init_cluster():
 
     logger.info("Loading initial cluster configuration")
 
-    crm_configure_load("update", """property cib-bootstrap-options: stonith-enabled=false
+    crm_configure_load("update", """property cib-bootstrap-options: fencing-enabled=false
 op_defaults op-options: timeout=600
 rsc_defaults rsc-options: resource-stickiness=1 migration-threshold=3
 """)
@@ -2708,7 +2708,7 @@ def adjust_fencing_timeout():
     """
     value = get_fencing_timeout_generally_expected()
     if value:
-        utils.set_property("stonith-timeout", value, conditional=True)
+        utils.set_property("fencing-timeout", value, conditional=True)
 
 
 def adjust_properties():
--- a/crmsh/crash_test/utils.py
+++ b/crmsh/crash_test/utils.py
@@ -123,7 +123,7 @@ class FenceInfo(object):
         action_result = crmshutils.get_property("fencing-action")
         supported_actions = ra.get_property_options("fencing-action")
         if action_result is None or action_result not in supported_actions:
-            current_action_term = crmshutils.DeprecatedTermTranslator.get_working_term("stonith-action")
+            current_action_term = crmshutils.DeprecatedTermTranslator.get_working_term("fencing-action")
             msg_error(f"Cluster property \"{current_action_term}\" should be {'|'.join(supported_actions)}")
             return None
         return action_result
--- a/crmsh/qdevice.py
+++ b/crmsh/qdevice.py
@@ -630,7 +630,7 @@ class QDevice(object):
                 sbd_watchdog_timeout_qdevice = sbd.SBDTimeout.SBD_WATCHDOG_TIMEOUT_DEFAULT_WITH_QDEVICE
                 sbd.SBDManager.update_sbd_configuration({"SBD_WATCHDOG_TIMEOUT": str(sbd_watchdog_timeout_qdevice)})
                 if self.is_stage:
-                    utils.set_property("stonith-watchdog-timeout", 2*sbd_watchdog_timeout_qdevice)
+                    utils.set_property("fencing-watchdog-timeout", 2*sbd_watchdog_timeout_qdevice)
 
     @qnetd_lock_for_same_cluster_name
     def certificate_and_config_qdevice(self):
--- a/crmsh/ra.py
+++ b/crmsh/ra.py
@@ -332,9 +332,6 @@ class RAInfo(object):
             if params[name]["deprecated"] == "1"
         }
 
-    def get_all_params_dict(self):
-        """Return non-deprecated parameters with full metadata."""
-        return self.get_active_params()
 
 
     def params(self):
--- a/crmsh/ra.py
+++ b/crmsh/ra.py
@@ -143,7 +143,7 @@ def get_properties_list():
 @functools.cache
 def get_properties_without_deprecated():
     try:
-        return get_properties_meta().get_param_list_without_deprecated()
+        return list(get_properties_meta().get_active_params())
     except:
         return []
 
@@ -299,7 +299,30 @@ class RAInfo(object):
             replaced_with = replaced_with_elem.get("name")
         return deprecated, replaced_with
 
-    def get_deprecated_params_dict(self):
+    def get_active_params(self):
+        """Return non-deprecated parameters with full metadata."""
+        params = self.params()
+        if not params:
+            return {}
+        return {
+            name: params[name]
+            for name in params
+            if params[name]["deprecated"] != "1"
+        }
+
+    def get_deprecated_params(self):
+        """Return deprecated parameters with full metadata."""
+        params = self.params()
+        if not params:
+            return {}
+        return {
+            name: params[name]
+            for name in params
+            if params[name]["deprecated"] == "1"
+        }
+
+    def get_deprecated_mapping(self):
+        """Return deprecated parameter names mapped to replacement names."""
         params = self.params()
         if not params:
             return {}
@@ -309,14 +332,10 @@ class RAInfo(object):
             if params[name]["deprecated"] == "1"
         }
 
-    def get_param_list_without_deprecated(self):
-        params = self.params()
-        if not params:
-            return []
-        return [
-            name for name in params
-            if params[name]["deprecated"] != "1"
-        ]
+    def get_all_params_dict(self):
+        """Return non-deprecated parameters with full metadata."""
+        return self.get_active_params()
+
 
     def params(self):
         '''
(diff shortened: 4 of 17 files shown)


nicholasyang2022 (changes requested):
Thank you for this PR! Moving towards dynamic, metadata-driven discovery for deprecated properties is a fantastic architectural improvement that will make `crmsh` much more resilient to upstream Pacemaker changes.

However, during review, I identified a few critical safety concerns regarding how the migration (`--fix`) is handled, as well as some API design issues that need to be addressed before this can be merged.

#### Critical/Blocking Issues (Safety)

**1. Unsafe Scope of SBD Health Check (`crm cluster health sbd --fix`)**
Currently, `utils.get_all_configured_deprecated_properties()` fetches **all** deprecated properties defined in the Pacemaker metadata.
*   **The Risk:** Running `crm cluster health sbd --fix` will now indiscriminately check, migrate, and potentially delete *any* deprecated cluster property (e.g., legacy quorum policies, placement strategies) across the entire cluster, not just those related to SBD or fencing.
*   **Recommendation:** Please scope the dynamic discovery strictly to properties relevant to the SBD module. For instance, filter the dynamically discovered list to only include properties starting with `stonith-` or `fencing-`, or pass a targeted prefix list to the utility.

**2. Unsafe Deletion of Properties Without Replacements**
In `utils.py`, `DeprecatedTermTranslator.fix()` automatically deletes properties that have no replacement (`res.new_term` is `None`):
```python
        if res.using_deprecated and res.deprecated_configured:
            # ...
            else:
                delete_property(res.deprecated_term)
```
*   **The Risk:** Automatically deleting a configured property just because it is marked deprecated (without a replacement) is highly dangerous. An administrator may rely on that legacy property for specific behavior.
*   **Recommendation:** If a property has no replacement, the `--fix` command should **not** delete it. It is perfectly fine to log a warning during the check phase, but destructive actions without a clear migration path must be avoided to prevent breaking the cluster.

#### API Design & Maintainability

**3. The `direct` Flag in `get_property` / `set_property`**
Introducing `direct=True` to bypass the auto-translation logic solves the infinite recursion bug, but using a boolean flag to turn off core business logic is a code smell. Since most of the codebase expects these functions to be "smart" (auto-translating), we shouldn't clutter the signature.
*   **Recommendation:** Instead of a `direct` flag, extract the core I/O logic into explicit "raw" or "dumb" functions (e.g., `_get_raw_property` and `_set_raw_property`). The migration code in `DeprecatedTermTranslator.fix()` can safely call these raw functions directly. The public `get_property`/`set_property` functions can remain as "smart" wrappers that handle translation and then call the raw functions.

**4. Predictability of `RAInfo.get_all_params_dict()`**
The boolean flags in `def get_all_params_dict(self, deprecated=False, replaced_with=False)` make the return type unpredictable. 
*   **The Issue:** Depending on the flags, it either returns `dict[str, dict]` (full parameter metadata) or `dict[str, str|None]` (mapping a name to its replacement). Furthermore, if `deprecated=False` and `replaced_with=True`, the `replaced_with` flag is silently ignored. This breaks type hinting and readability.
*   **Recommendation:** Remove the boolean flags and split this into explicit, single-purpose properties/methods that return consistent types. For example:
    *   `get_active_params(self) -> dict`
    *   `get_deprecated_params(self) -> dict`
    *   `get_deprecated_mapping(self) -> dict[str, str|None]`

**5. Missing Null-Check on Value Migration**
In `DeprecatedTermTranslator.fix()`:
```python
            elif res.new_term:
                value = get_property(res.deprecated_term, get_default=False, direct=True)
                set_property(res.new_term, value, direct=True)
```
*   **Recommendation:** If `get_property` returns `None` (e.g., deleted concurrently), `set_property` will literally set the value to the string `"None"`. Please wrap this in a quick null-check: `if value is not None:` before setting the new property.

--- 

Overall, the foundation of this PR is very strong. Addressing these scope and API design issues will make this a robust addition to the codebase. Let me know if you have any questions!

liangxin1300:
> Thank you for this PR! Moving towards dynamic, metadata-driven discovery for deprecated properties is a fantastic architectural improvement that will make `crmsh` much more resilient to upstream Pacemaker changes.
> 
> However, during review, I identified a few critical safety concerns regarding how the migration (`--fix`) is handled, as well as some API design issues that need to be addressed before this can be merged.
> 
> #### Critical/Blocking Issues (Safety)
> **1. Unsafe Scope of SBD Health Check (`crm cluster health sbd --fix`)** Currently, `utils.get_all_configured_deprecated_properties()` fetches **all** deprecated properties defined in the Pacemaker metadata.
> 
> * **The Risk:** Running `crm cluster health sbd --fix` will now indiscriminately check, migrate, and potentially delete _any_ deprecated cluster property (e.g., legacy quorum policies, placement strategies) across the entire cluster, not just those related to SBD or fencing.
> * **Recommendation:** Please scope the dynamic discovery strictly to properties relevant to the SBD module. For instance, filter the dynamically discovered list to only include properties starting with `stonith-` or `fencing-`, or pass a targeted prefix list to the utility.
> 
> **2. Unsafe Deletion of Properties Without Replacements** In `utils.py`, `DeprecatedTermTranslator.fix()` automatically deletes properties that have no replacement (`res.new_term` is `None`):
> 
> ```python
>         if res.using_deprecated and res.deprecated_configured:
>             # ...
>             else:
>                 delete_property(res.deprecated_term)
> ```
> 
> * **The Risk:** Automatically deleting a configured property just because it is marked deprecated (without a replacement) is highly dangerous. An administrator may rely on that legacy property for specific behavior.
> * **Recommendation:** If a property has no replacement, the `--fix` command should **not** delete it. It is perfectly fine to log a warning during the check phase, but destructive actions without a clear migration path must be avoided to prevent breaking the cluster.
> 
> #### API Design & Maintainability
> **3. The `direct` Flag in `get_property` / `set_property`** Introducing `direct=True` to bypass the auto-translation logic solves the infinite recursion bug, but using a boolean flag to turn off core business logic is a code smell. Since most of the codebase expects these functions to be "smart" (auto-translating), we shouldn't clutter the signature.
> 
> * **Recommendation:** Instead of a `direct` flag, extract the core I/O logic into explicit "raw" or "dumb" functions (e.g., `_get_raw_property` and `_set_raw_property`). The migration code in `DeprecatedTermTranslator.fix()` can safely call these raw functions directly. The public `get_property`/`set_property` functions can remain as "smart" wrappers that handle translation and then call the raw functions.
> 
> **4. Predictability of `RAInfo.get_all_params_dict()`** The boolean flags in `def get_all_params_dict(self, deprecated=False, replaced_with=False)` make the return type unpredictable.
> 
> * **The Issue:** Depending on the flags, it either returns `dict[str, dict]` (full parameter metadata) or `dict[str, str|None]` (mapping a name to its replacement). Furthermore, if `deprecated=False` and `replaced_with=True`, the `replaced_with` flag is silently ignored. This breaks type hinting and readability.
> * **Recommendation:** Remove the boolean flags and split this into explicit, single-purpose properties/methods that return consistent types. For example:
>   
>   * `get_active_params(self) -> dict`
>   * `get_deprecated_params(self) -> dict`
>   * `get_deprecated_mapping(self) -> dict[str, str|None]`
> 
> **5. Missing Null-Check on Value Migration** In `DeprecatedTermTranslator.fix()`:
> 
> ```python
>             elif res.new_term:
>                 value = get_property(res.deprecated_term, get_default=False, direct=True)
>                 set_property(res.new_term, value, direct=True)
> ```
> 
> * **Recommendation:** If `get_property` returns `None` (e.g., deleted concurrently), `set_property` will literally set the value to the string `"None"`. Please wrap this in a quick null-check: `if value is not None:` before setting the new property.
> 
> Overall, the foundation of this PR is very strong. Addressing these scope and API design issues will make this a robust addition to the codebase. Let me know if you have any questions!

Changed items 3, 4, and 5.
Items 1 and 2 need discussion.

liangxin1300:
- [x] Need to complete some test cases in deprecated_term.feature

zzhou1 (approved):
Approved, with only one nitpicking comment for a possible improvement: if it's simple enough, we could delete deprecated options with default values (e.g., concurrent-fencing=true). I assume that would clean up the warnings, since most users stick with the defaults and don't need the noisy messages.

nicholasyang2022 (approved):

PR #2161, opened 2026-07-23
Dev: migration: Add --force-breaks-pacemaker switch to bypass health check blockers (bsc#1250586)

nicholasyang2022:
Allows users to force the corosync configuration migration even if CIB
prerequisite checks return critical blockers (such as unsupported resource or
fence agents). This enables starting the cluster in an unsupported state on
the new OS so that users can carry out manual actions to fix the unsupported
agents afterwards.

- Adds --force-breaks-pacemaker option to the migration parser.
- Skips _check_impl() and runs migrate_corosync_conf() directly if the switch is set.
- Adds new unit tests to test_migration.py verifying the bypass logic.
- Adds a functional test scenario to migration.feature to test the bypass behavior under blocker conditions.

### DIFF
--- a/crmsh/migration.py
+++ b/crmsh/migration.py
@@ -159,8 +159,15 @@ class CheckResultInteractiveHandler(CheckResultHandler):
 def migrate(args: typing.Sequence[str]):
     parser = argparse.ArgumentParser(args[0])
     parser.add_argument('--local', action='store_true')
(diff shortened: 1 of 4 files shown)


liangxin1300 (approved):
Nice work

Thanks!

liangxin1300:
Need also to mention this new option in crm help

PR #2163, opened 2026-07-26
feat(ui_node): add on|off option to `crm node maintenance` and deprecate `node ready`

liangxin1300:
## Summary

Expand `crm node maintenance` so maintenance can be enabled or cleared
from the same command, matching the cluster and resource maintenance
command forms, and support the same local-node-default / multi-node
conventions already used by `crm node standby`.

## Changes

- Accept `crm node maintenance [<node>...] [on|off]` to enable/disable
  maintenance
- Set node maintenance attribute using `true`/`false` instead of only
  enabling it
- Log successful maintenance attribute updates
- Deprecate `crm node ready` in favor of `crm node maintenance <node>
  off`, with a deprecation warning in command output
- `crm node maintenance off` (no node given) now targets the local
  node, instead of misinterpreting `off` as a node name
- `crm node maintenance <node1> <node2> ... [on|off]` now supports
  multiple nodes in a single invocation
- Document the on|off option, multi-node/local-node-default support,
  and the `ready` deprecation in `crm.8.adoc`
- Update regression test expectations (`cluster_api.feature`)

## Testing

Full unit test suite passes: `pytest -q test/unittests`

### DIFF
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -222,7 +222,6 @@ class NodeMgmt(command.UI):
     name = "node"
 
     node_standby = "crm_attribute -t nodes -N '%s' -n standby -v '%s' %s"
-    node_maint = "crm_attribute -t nodes -N '%s' -n maintenance -v '%s'"
     node_delete = """cibadmin -D -o nodes -X '<node uname="%s"/>'"""
     node_delete_status = """cibadmin -D -o status -X '<node_state uname="%s"/>'"""
     node_cleanup_resources = "crm_resource --cleanup --node '%s'"
@@ -435,7 +434,12 @@ class NodeMgmt(command.UI):
             node = utils.this_node()
         if not utils.is_name_sane(node):
             return False
-        return utils.ext_cmd(self.node_maint % (node, "off")) == 0
+        rc = self._commit_node_attr(context, node, "maintenance", "false")
+        if rc:
+            logger.info("Setting maintenance=false on node %s", node)
+        suggestion = f"crm node maintenance {node} off"
+        logger.warning("The 'ready' command is deprecated and will be removed in a future release. Use '%s' instead.", suggestion)
+        return rc
 
     @command.wait
     @command.completers(compl.nodes)
--- a/crmsh/ui_node.py
+++ b/crmsh/ui_node.py
@@ -411,14 +411,20 @@ class NodeMgmt(command.UI):
             logger.info("online node %s", node)
 
     @command.wait
-    @command.completers(compl.nodes)
-    def do_maintenance(self, context, node=None):
-        'usage: maintenance [<node>]'
+    @command.completers(compl.nodes, compl.choice(['on', 'off']))
+    def do_maintenance(self, context, node=None, on_off='on'):
+        'usage: maintenance [<node>] [on|off]'
         if not node:
             node = utils.this_node()
         if not utils.is_name_sane(node):
             return False
-        return self._commit_node_attr(context, node, "maintenance", "true")
+        if on_off not in ['on', 'off']:
+            context.fatal_error("Expected <node> [on|off]")
+        _value = "true" if on_off == 'on' else "false"
+        rc = self._commit_node_attr(context, node, "maintenance", _value)
+        if rc:
+            logger.info("Setting maintenance=%s on node %s", _value, node)
+        return rc
 
 
     @command.wait
--- a/test/features/cluster_api.feature
+++ b/test/features/cluster_api.feature
@@ -22,6 +22,29 @@ Feature: Functional test to cover SAP clusterAPI
     When    Run "echo 'export PATH=$PATH:/usr/sbin/' > ~hacluster/.bashrc" on "hanode2"
 
   @clean
+  Scenario: crm node maintenance supports on|off
+    When    Run "crm node maintenance hanode2 on" on "hanode1"
+    Then    Node "hanode2" is maintenance
+    When    Run "crm node maintenance hanode2 off" on "hanode1"
+    Then    Node "hanode2" is ready
+
+    # default value is "on"
+    When    Run "crm node maintenance hanode2" on "hanode1"
+    Then    Node "hanode2" is maintenance
+    When    Run "crm node maintenance hanode2 off" on "hanode1"
+    Then    Node "hanode2" is ready
+
+    When    Try "crm node maintenance hanode2 xxx"
+    Then    Except "ERROR: node.maintenance: Expected <node> [on|off]"
+
+    # "crm node ready" is deprecated in favor of "crm node maintenance <node> off"
+    When    Run "crm node maintenance hanode2 on" on "hanode1"
+    Then    Node "hanode2" is maintenance
+    When    Run "crm node ready hanode2" on "hanode1"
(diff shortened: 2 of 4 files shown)


crmsh/ui_node.py
```diff
@@ -365,22 +363,23 @@ def do_standby(self, context, *args):
...
                 res_item = xmlutil.get_set_nodes(interface_item, "instance_attributes", create=True)
-                xmlutil.set_attr(res_item[0], "standby", "on")
+                xmlutil.set_attr(res_item[0], "standby", "true")
 
         rc = utils.diff_and_patch(xmlutil.xml_tostring(orig_cib), xmlutil.xml_tostring(cib))
         if not rc:
             return False
         for node in node_list:
             logger.info("standby node %s", node)
+        logger.info("To restore node(s) from standby, use the `online` command")
```
gao-yan: If we are to unify to have `crm node maintenance true|false` and deprecate `crm node ready`, would it make sense to do the similar for this like `crm node standby true|false` and deprecate `crm node online`?

crmsh/ui_node.py
```diff
@@ -415,14 +414,22 @@ def do_online(self, context, *args):
             logger.info("online node %s", node)
 
     @command.wait
-    @command.completers(compl.nodes)
-    def do_maintenance(self, context, node=None):
-        'usage: maintenance [<node>]'
+    @command.completers(compl.nodes, compl.choice(['on', 'off']))
+    def do_maintenance(self, context, node=None, on_off='on'):
+        'usage: maintenance [<node>] [on|off]'
```
gao-yan, line 415: Boolean values have been kind of inconsistent between different places. Does it make sense to try to consistently advise `true|false` while internally still support `on|off`?

liangxin1300: Since we already have `crm maintenance on|off` and `crm resource maintenance <res> on|off`
to keep the command format the same and keep compatibility,
we can add option `crm node maintenance <node> [on|off]`, although we actually set the attribute value as "true" and "false"

We can see most of pacemaker attributes are using true|false as values

crmsh/ui_node.py
liangxin1300: Makes sense.

nicholasyang2022: Are you going to add this to this pull request?

liangxin1300: Better to create a new PR for standby.
Already removed standby-related code

See:
- #2193

doc/crm.8.adoc
```diff
@@ -2372,9 +2369,14 @@ property from them.
 
 The node parameter defaults to the node where the command is run.
 
+To disable maintenance mode on the node, use `crm node maintenance <node> off`.
+
+Note: maintenance on node is transient by default. Restarting Pacemaker
+will clear it and may trigger fencing.
```
nicholasyang2022: > by default

So is there a way to set it permanently?

liangxin1300: Removed, since this attribute is permanent

zzhou1:
Some UX discrepancies between "maintenance" in this pr#2163 and "standby" in https://github.com/ClusterLabs/crmsh/pull/2193:

UX-1: to support local node, eg. `crm node maintenance off`
```
161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance
INFO: Setting maintenance=true on node 161-1
161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance off
ERROR: Node 'off' not found in CIB
ERROR: Failed to update node attributes for off
161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance 161-1 off
INFO: Setting maintenance=false on node 161-1

161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby
INFO: standby node 161-2
161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby off
INFO: online node 161-2
161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby 161-2 off
INFO: online node 161-2
```

UX-2: to support multi-nodes, eg. `crm node maintenance 161-1 161-2 off`
```
161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance 161-1 161-2 off
ERROR: node.maintenance: Expected ([node] [on_off]), takes at most 2 arguments (3 given)

161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby 161-2 161-1 off
INFO: online node 161-2
INFO: online node 161-1
```

UX-3: small tweaks to make `--help` more similar a little
```
crm node maintenance --help
crm node standby --help
```

liangxin1300:
> Some UX discrepancies between "maintenance" in this pr#2163 and "standby" in #2193:
> 
> UX-1: to support local node, eg. `crm node maintenance off`
> 
> ```
> 161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance
> INFO: Setting maintenance=true on node 161-1
> 161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance off
> ERROR: Node 'off' not found in CIB
> ERROR: Failed to update node attributes for off
> 161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance 161-1 off
> INFO: Setting maintenance=false on node 161-1
> 
> 161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby
> INFO: standby node 161-2
> 161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby off
> INFO: online node 161-2
> 161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby 161-2 off
> INFO: online node 161-2
> ```
> 
Agree

> UX-2: to support multi-nodes, eg. `crm node maintenance 161-1 161-2 off`
> 
> ```
> 161-1:/var/hostfiles/crmsh-pr-2163 # crm node maintenance 161-1 161-2 off
> ERROR: node.maintenance: Expected ([node] [on_off]), takes at most 2 arguments (3 given)
> 
> 161-2:/var/hostfiles/crmsh-pr-2193 # crm node standby 161-2 161-1 off
> INFO: online node 161-2
> INFO: online node 161-1
> ```
> 
I remeber there is user story that we support standby for multi nodes(and --all), do you think here for maintenance on node level has the same user story?

> UX-3: small tweaks to make `--help` more similar a little
> 
> ```
> crm node maintenance --help
> crm node standby --help
> ```

zzhou1:
> I remeber there is user story that we support standby for multi nodes(and --all), do you think here for maintenance on node level has the same user story?

IIRC, `standby --all` is about the story to stop the whole cluster gracefully, and there is no "standby" at the cluster level.

What I have in mind here is set multiple nodes in the maintenance mode to manage a cluster in 5 or 7 nodes, for example.

liangxin1300:
> > I remeber there is user story that we support standby for multi nodes(and --all), do you think here for maintenance on node level has the same user story?
> 
> IIRC, `standby --all` is about the story to stop the whole cluster gracefully, and there is no "standby" at the cluster level.
> 
> What I have in mind here is set multiple nodes in the maintenance mode to manage a cluster in 5 or 7 nodes, for example.

So, no need to add --all option, right?

zzhou1:
> So, no need to add --all option, right?

Agree, no need `--all` for the node-level maintenance.

zzhou1 (approved):

PR #2172, opened 2026-08-13
Dev: crash_test: Prefer nft for split-brain blocking

liangxin1300:
The split-brain crash test should use nft when available because nft can
apply the full blocking ruleset atomically. Keep iptables as a fallback for
systems without nft.
    
- Apply nft blocking rules as one ruleset
- Prefer nft and fall back to iptables
- Destroy the nft table during cleanup
- Cover nft-first pre-check and ruleset generation

### DIFF
--- a/crmsh/crash_test/config.py
+++ b/crmsh/crash_test/config.py
@@ -2,6 +2,18 @@ FENCE_TIMEOUT = 60
 FENCE_NODE = "crm_attribute -t status -N '{}' -n terminate -v true"
 BLOCK_IP = '''iptables -{action} INPUT -s {peer_ip} -j DROP;
               iptables -{action} OUTPUT -d {peer_ip} -j DROP'''
-REMOVE_PORT = "firewall-cmd --zone=public --remove-port={port}/udp"
-ADD_PORT = "firewall-cmd --zone=public --add-port={port}/udp"
+NFT_TABLE = "crmsh_split_brain"
+NFT_RULESET = '''destroy table inet {table}
+table inet {table} {{
+  chain input {{
+    type filter hook input priority -300; policy accept;
+{input_rules}
+  }}
+  chain output {{
+    type filter hook output priority -300; policy accept;
+{output_rules}
+  }}
+}}'''
+NFT_APPLY_RULESET = "nft -f - <<'EOF'\n{ruleset}\nEOF"
+NFT_DELETE_TABLE = "nft destroy table inet {table}"
 SBD_CONF = "/etc/sysconfig/sbd"
--- a/crmsh/crash_test/task.py
+++ b/crmsh/crash_test/task.py
@@ -443,6 +443,8 @@ class TaskSplitBrain(Task):
         self.expected = "One of nodes get fenced"
         self.ports = []
         self.peer_nodelist = []
+        self.use_nft = False
+        self.shell = ShellUtils()
         super(self.__class__, self).__init__(self.description, flush=True)
         self.force = force
 
@@ -475,10 +477,14 @@ Fence timeout:     {}
         """
         self.task_pre_check()
 
-        for cmd in ["iptables"]:
-            rc, _, err = ShellUtils().get_stdout_stderr("which {}".format(cmd))
-            if rc != 0 and err:
-                raise TaskError(err)
+        rc, _, _ = self.shell.get_stdout_stderr("which nft")
+        if rc == 0:
+            self.use_nft = True
+        else:
+            rc, _, _ = self.shell.get_stdout_stderr("which iptables")
+            if rc != 0:
+                raise TaskError("Please provide the nft or iptables command")
+            self.use_nft = False
 
         if len(utils.online_nodes()) < 2:
             raise TaskError("At least two nodes online!")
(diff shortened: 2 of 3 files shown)


crmsh/crash_test/config.py
```diff
@@ -2,6 +2,11 @@
...
 BLOCK_IP = '''iptables -{action} INPUT -s {peer_ip} -j DROP;
               iptables -{action} OUTPUT -d {peer_ip} -j DROP'''
-REMOVE_PORT = "firewall-cmd --zone=public --remove-port={port}/udp"
-ADD_PORT = "firewall-cmd --zone=public --add-port={port}/udp"
+NFT_TABLE = "crmsh_split_brain"
+NFT_CREATE_TABLE = "nft add table inet {table}"
+NFT_CREATE_INPUT_CHAIN = "nft 'add chain inet {table} input {{ type filter hook input priority -300; policy accept; }}'"
+NFT_CREATE_OUTPUT_CHAIN = "nft 'add chain inet {table} output {{ type filter hook output priority -300; policy accept; }}'"
+NFT_BLOCK_IP = '''nft add rule inet {table} input ip saddr {peer_ip} drop;
+                nft add rule inet {table} output ip daddr {peer_ip} drop'''
```
nicholasyang2022: This is an anti-pattern in nft. nft supersedes iptables as it supports defining and applying the entire ruleset in an atomic and transactional way:

```
flush table inet crmsh_split_brain
table inet crmsh_split_brain {
  chain input {
    type filter hook input priority -300; policy accept;
    ip saddr {{peer_ip}} drop
  }
  chain output {
    type filter hook output priority -300; policy accept;
    ip daddr {{peer_ip}} drop
  }
}
```

And it would be better to use nft when available and to use iptables as a fallback.

crmsh/crash_test/task.py
```diff
@@ -494,15 +503,32 @@ def do_block(self):
...
+        if self.use_nft:
+            self.peer_ip_list = []
         for node in self.peer_nodelist:
             self.info("Trying to temporarily block {} communication ip".format(node))
             for ip in crmshutils.get_iplist_from_name(node):
-                ShellUtils().get_stdout_stderr(config.BLOCK_IP.format(action='I', peer_ip=ip))
+                if self.use_nft:
+                    self.peer_ip_list.append(ip)
+                    continue
+                self.shell.get_stdout_stderr(self.block_ip_cmd.format(action="I", peer_ip=ip, table=config.NFT_TABLE))
```
zzhou1:
```suggestion
                self.shell.get_stdout_stderr(self.block_ip_cmd.format(action="I", peer_ip=ip))
```

crmsh/crash_test/task.py
```diff
@@ -514,10 +540,13 @@ def un_block_iptables(self):
...
         Unblock corosync communication ip
         """
+        if self.use_nft:
+            self.shell.get_stdout_stderr(config.NFT_DELETE_TABLE.format(table=config.NFT_TABLE))
+            return
         for node in self.peer_nodelist:
             self.info("Trying to recover {} communication ip".format(node))
             for ip in crmshutils.get_iplist_from_name(node):
-                ShellUtils().get_stdout_stderr(config.BLOCK_IP.format(action='D', peer_ip=ip))
+                self.shell.get_stdout_stderr(self.block_ip_cmd.format(action='D', peer_ip=ip, table=config.NFT_TABLE))
```
zzhou1:
```suggestion
                self.shell.get_stdout_stderr(self.block_ip_cmd.format(action='D', peer_ip=ip))
```

crmsh/crash_test/task.py
```diff
@@ -475,10 +479,15 @@ def pre_check(self):
...
-                raise TaskError(err)
+        rc, _, _ = self.shell.get_stdout_stderr("which nft")
+        if rc == 0:
+            self.use_nft = True
+        else:
+            rc, _, _ = self.shell.get_stdout_stderr("which iptables")
+            if rc != 0:
+                raise TaskError("Please provide the nft or iptables command")
+            self.block_ip_cmd = config.BLOCK_IP
+            self.use_nft = False
```
zzhou1: These two lines are redundant with the default in __init__()
```suggestion
```

nicholasyang2022: AFAIC it is a good practice to set `self.use_nft` in this branch, as it does not depends on the previous value of the variable and make the code more readable.

However, storing `block_ip_cmd` in a member variable is already redundant. It depends entirely on `use_nft` and store the boolean is enough.

nicholasyang2022 (commented):
Hi @liangxin1300, 

Thank you for this PR! Prioritizing `nftables` when available is an excellent and modern improvement for the split-brain crash test. The implementation is highly robust, and the accompanying unit tests are comprehensive and well-structured.

I have a few architectural and code-style recommendations aimed at improving readability, separation of concerns, and keeping class state clean.

---

### **Architectural Recommendations**

#### **1. Decouple `nftables` and `iptables` Logic**
Currently, `_block_with_nft()` is called inside `do_block_iptables()`, and both code paths are interleaved within a single method using `if self.use_nft:` branches. 
* **Recommendation**: It is much cleaner to separate these into dedicated, highly cohesive methods and route them inside the orchestrators (`do_block` and `un_block`). This keeps both backend drivers fully isolated and leaves the original `iptables` code paths untouched.

#### **2. Minimize Object State (Avoid Leaky Member Variables)**
* **`self.peer_ip_list`**: This is only used as transient storage to build the `nft` ruleset during `do_block_iptables` and is never read again. We should make this a local variable instead of storing it as a member variable on `self`.
* **`self.block_ip_cmd`**: This is a static constant (`config.BLOCK_IP`). We can reference it directly from `config` rather than assigning it to `self`.
* **`self.use_nft`**: This should remain a member variable since it is set in `pre_check` and used in both setup/teardown.

#### **3. Redundant Formatting Arguments**
In `do_block_iptables` and `un_block_iptables`, when formatting `self.block_ip_cmd` (`config.BLOCK_IP`), `table=config.NFT_TABLE` is passed:
```python
self.shell.get_stdout_stderr(self.block_ip_cmd.format(action="I", peer_ip=ip, table=config.NFT_TABLE))
```
Since `config.BLOCK_IP` has no `{table}` placeholder, this argument is ignored by Python but can be confusing to future maintainers. Removing it keeps the `iptables` commands clean.

---

### **Suggested Refactoring**

Here is how we can clean up `crmsh/crash_test/task.py` to achieve clear separation:

```python
    @contextmanager
    def do_block(self):
        """
        Context manager to block and unblock ip/ports
        """
        if self.use_nft:
            self.do_block_nft()
        else:
            self.do_block_iptables()
        try:
            yield
        finally:
            self.un_block()

    def do_block_nft(self):
        """
        Apply blocking rules atomically using nftables
        """
        self.peer_nodelist = utils.peer_node_list()
        peer_ips = []
        for node in self.peer_nodelist:
            self.info("Trying to temporarily block {} communication ip".format(node))
            peer_ips.extend(crmshutils.get_iplist_from_name(node))

        input_rules = "\n".join("    ip saddr {} drop".format(ip) for ip in peer_ips)
        output_rules = "\n".join("    ip daddr {} drop".format(ip) for ip in peer_ips)

        ruleset = config.NFT_RULESET.format(
            table=config.NFT_TABLE,
            input_rules=input_rules,
            output_rules=output_rules,
        )
        self.shell.get_stdout_stderr(config.NFT_APPLY_RULESET.format(ruleset=ruleset))

    def do_block_iptables(self):
        """
        Block corosync communication ip using legacy iptables
        """
        self.peer_nodelist = utils.peer_node_list()
        for node in self.peer_nodelist:
            self.info("Trying to temporarily block {} communication ip".format(node))
            for ip in crmshutils.get_iplist_from_name(node):
                self.shell.get_stdout_stderr(config.BLOCK_IP.format(action="I", peer_ip=ip))

    def un_block(self):
        """
        Unblock corosync ip/ports
        """
        if self.use_nft:
            self.un_block_nft()
        else:
            self.un_block_iptables()

    def un_block_nft(self):
        """
        Remove the nftables blocking rules by destroying the table
        """
        self.shell.get_stdout_stderr(config.NFT_DELETE_TABLE.format(table=config.NFT_TABLE))

    def un_block_iptables(self):
        """
        Unblock corosync communication ip using legacy iptables
        """
        for node in self.peer_nodelist:
            self.info("Trying to recover {} communication ip".format(node))
            for ip in crmshutils.get_iplist_from_name(node):
                self.shell.get_stdout_stderr(config.BLOCK_IP.format(action='D', peer_ip=ip))
```

This makes each driver path simple, readable, and highly maintainable! Let me know what you think.

zzhou1 (approved):

nicholasyang2022 (approved):

PR #2173, opened 2026-08-17
Fix: storage_utils: Check multipath devices on all nodes

liangxin1300:
Multipath validation for SBD and cluster filesystems only checked the local node. This change runs the multipath inspection on each target node and reports where the invalid device was found.

- Add peer-aware MultipathInspector checks
- Pass node lists from SBD and cluster FS validation
- Log multipath errors with the failing node
- Add unit tests for multi-node checks

### DIFF
--- a/crmsh/storage_utils.py
+++ b/crmsh/storage_utils.py
@@ -209,10 +209,21 @@ class MultipathInspector:
         self._device_info = self._inspect(dev)
 
     def _get_parent_device(self, dev) -> str:
-        resolved = Path(dev).resolve()
-        cmd = f"lsblk -dn -o PKNAME {shlex.quote(str(resolved))}"
-        _, out, _ = self._shell.get_rc_stdout_stderr_without_input(self._peer, cmd)
-        return out or resolved.name
+        cmd = f"lsblk -dnP -o PKNAME,KNAME {shlex.quote(dev)}"
+        rc, out, _ = self._shell.get_rc_stdout_stderr_without_input(self._peer, cmd)
+        node = self._peer or utils.this_node()
+        if rc != 0 or not out:
+            raise ValueError(f"Cannot determine kernel device name for {dev} on {node}")
+
+        values = dict(
+            item.split("=", 1)
+            for item in shlex.split(out.splitlines()[0])
+            if "=" in item
+        )
+        device_name = values.get("PKNAME") or values.get("KNAME")
+        if not device_name:
+            raise ValueError(f"Cannot determine kernel device name for {dev} on {node}")
+        return device_name
 
     def _get_multipath_mapping(self) -> dict[str, str]:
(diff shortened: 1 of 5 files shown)


crmsh/storage_utils.py
```diff
@@ -242,8 +243,9 @@ def _is_under_multipath(self) -> bool:
...
-    def check_device_under_multipath(cls, dev):
-        inspector = cls(dev)
-        if inspector._is_under_multipath():
-            error_msg = f"Device {dev} is under multipath, please provide the multipath device instead"
-            raise ValueError(error_msg)
+    def check_device_under_multipath(cls, dev, node_list=None):
+        for node in node_list or [utils.this_node()]:
+            inspector = cls(dev, node)
+            if inspector._is_under_multipath():
+                error_msg = f"Device {dev} is under multipath on {node}, please provide the multipath device instead"
```
nicholasyang2022: > please provide the multipath device instead

I am not sure if this suggestion is accurate. If a device is only under multipath on some of the nodes, it is probably a different problem, for example, inconsistent device file names in different node. I incline to report only the error without providing suggestions blindly.

liangxin1300: > inconsistent device file names in different node

This is the prerequisite to configuring disk-based SBD, see https://documentation.suse.com/sle-ha/15-SP7/html/SLE-HA-all/cha-ha-storage-protect.html#sec-ha-storage-protect-req

nicholasyang2022: Even though consistent device naming is a prerequisite, validation logic should detect and report when this prerequisite is unsatisfied rather than assuming it is met. Suggesting users provide the multipath device when names are actually inconsistent is misleading and should be avoided.

crmsh/storage_utils.py
```diff
@@ -203,19 +203,20 @@ class DeviceInfo:
...
+    def __init__(self, dev, peer=None):
         self._shell = sh.cluster_shell()
+        self._peer = peer
         self._device_info = self._inspect(dev)
 
     def _get_parent_device(self, dev) -> str:
         resolved = Path(dev).resolve()
         cmd = f"lsblk -dn -o PKNAME {shlex.quote(str(resolved))}"
-        _, out, _ = self._shell.get_rc_stdout_stderr_without_input(None, cmd)
+        _, out, _ = self._shell.get_rc_stdout_stderr_without_input(self._peer, cmd)
```
nicholasyang2022: It looks problematic to resolve symlink on localhost and run lsblk on a remote host.

Moreover, `PKNAME` returns the kernel device name, which will possibly negate the benefits of using stable names like `/dev/disk/by-id/...`.  I don't know what kinds of symlink you have tested here. It looks quite complex when combining iscsi, device mapper, multipath and partitions together.

PR #2181, opened 2026-08-27
test(features): use local container image by default and build one if not exists

nicholasyang2022:


liangxin1300:
The commit log prefix can be changed as
```
test: features:
```
test(features) looks like a new feature

nicholasyang2022:
> test(features) looks like a new feature

No This is defined clearly in conventional commit message style: `type(optional scope): description`. It is unambiguous that "test" is the type and "features" is the scope.

> test: features:

This is invalid in format.

liangxin1300 (approved):

PR #2199, opened 2026-09-09
refactor(bootstrap): split Context class into Arguments and GlobalVariables

nicholasyang2022:
- Monolithic `Context` class in `crmsh/bootstrap.py` was split into two clearly bounded responsibilities:
    1. `Arguments`: A representation storing pure command-line parsed options and interactive user input/answers.
    2. `GlobalVariables`: A container for all the other global state and logic.
- The `diskless_sbd` flag is passed clearly as an arugment in the `sbd` module.
    * When initializing `SBDManager` from a `GlobalVariables` instance, set
      `SBDManager.diskless_sbd` from `GlobalVariables.args.diskless_sbd`.
    * Refactor `SBDTimeout` to receive `diskless_sbd` argument directly,
      without indirection through a `GlobalVariables`.
    * Refactor `SBDConfigCheck` to composite `SBDTimeout` rather than
      inherit it, as `SBDConfigCheck` is not a `SBDTimeout`.
- Semantic Cleanups & Code Style
    - Renamed in-memory setup helpers to use a `populate_*` prefix (e.g., `_populate_qdevice`, `populate_user`, `populate_sbd_manager`) to reflect that they populate memory state rather than performing live initialization.
    - Fixed SBD docstring styles to consistently use triple double quotes (""").

This decoupling prevents accidental modification of user-supplied CLI arguments. For example, active IPv6 detection state is now isolated in _global_variables.ipv6 instead of mutating the parsed Arguments.ipv6 configuration.


crmsh/bootstrap.py
```diff
@@ -2169,24 +2176,24 @@ def bootstrap_init(context):
...
 
     if not stage or stage in ('corosync', 'sbd'):
-        _context.load_profiles()
-    _context.init_sbd_manager()
+        _global_variables.load_profiles()
+    _global_variables.populate_sbd_manager()
 
     if stage in ('qnetd_remote', ):
-        args = _context.args
+        args = _global_variables.args.args
```
liangxin1300: `.args.args` chain is confusing — worth renaming the passthrough field (e.g. positional_args)

liangxin1300 (commented):
Commit messages reference a `GlobalVariablesMock` test helper that doesn't exist

nicholasyang2022:
> Commit messages reference a `GlobalVariablesMock` test helper that doesn't exist

Removed the commit message about unit tests. I had used `GlobalVariablesMock` previously and realized that it is not needed as it does not have any method, and had removed it.

crmsh/bootstrap.py
nicholasyang2022: Renamed `Arguments.args` to `Arguments.remaining_args`.

liangxin1300 (approved):
LGTM.

Thanks!

PR #2203, opened 2026-09-16
feat(resource): warn when tracing a non-shell resource agent

liangxin1300:
`crm resource trace` relies on the tracing support built into ocf-shellfuncs (bash 'set -x'-style tracing), so it only works for resource agents implemented as shell scripts. Agents written in other languages, such as most stonith:fence_* agents which are written in Python, silently ignore the trace_ra/trace_dir attributes, even though crmsh reports that tracing was set up successfully.

Add crmsh.ra.is_shell_agent(), which locates the agent's executable on disk and inspects its shebang line to determine whether it is a shell script. `crm resource trace` now emits a warning when the target RA is known not to be a shell script.

Also document this limitation in the `trace` command help so that `crm resource help trace` mentions it.


crmsh/ra.py
```diff
@@ -871,6 +876,68 @@ def pick_provider(providers):
...
+        rsc.node.get("class"),
+        rsc.node.get("type"),
+        rsc.node.get("provider")
+    )
+    if not path or not os.path.isfile(path):
+        return None
+
+    try:
+        with open(path, 'rb') as f:
+            first_line = f.readline()
```
nicholasyang2022: This would possibly read the entire file into memory if it is a binary.

```suggestion
            magic = f.read(2)
            if magic != b'#!':
                return False
            f.seek(0)
            first_line = f.readline()
```

crmsh/ra.py
```diff
@@ -871,6 +876,68 @@ def pick_provider(providers):
...
+    except OSError:
+        return None
+
+    if not first_line.startswith(b'#!'):
+        return None
```
nicholasyang2022:
```suggestion
        return False
```

crmsh/ra.py
```diff
@@ -871,6 +876,68 @@ def pick_provider(providers):
...
+    interpreter = first_line[2:].decode('utf-8', errors='replace').strip()
+    if not interpreter:
+        return None
+    parts = interpreter.split()
+    if parts and parts[0].endswith('/env'):
+        parts = parts[1:]
+    if not parts:
+        return None
```
nicholasyang2022:
```suggestion
        return False
```

crmsh/ra.py
```diff
@@ -871,6 +876,68 @@ def pick_provider(providers):
...
+
+    interpreter_name = os.path.basename(parts[0])
+    return interpreter_name in SHELL_INTERPRETERS
```
nicholasyang2022: 1. There are other shells not listed in `SHELL_INTERPRETERS`.
2. `#!/usr/bin/env -S bash` falls here and it is a shell.

```suggestion
    return True if (interpreter_name in SHELL_INTERPRETERS) else None
```

nicholasyang2022 (approved):

